PR #828, opened 2014-12-13
Make link for new github repo open in new tab

yufengg:
Unfortunately GitHub markdown does not support opening links in a new tab. This change replaces the link to create a new github repo with some html that does that.

Here's a github issue thread that shows that markdown is not likely to support this natively in the future: https://github.com/mojombo/github-flavored-markdown/issues/28


Gribbs:
Opening link in a new tab should be an option. It's useability 101. Users should not have to lose their current page if the page designer doesn't want it to work like that (like me right now).

PR #835, opened 2014-12-16
Corelist

marycampione:
New Using core-list guide.

Thanks.
mem


arthurevans:
@ebidel @robdodson What do you think about including the examples for this doc in docs/elements/core-list-examples? In the past, we've put them in samples/, but I don't feel super-strongly about that...

Also, is there an easier way to show the sample text? Maybe a page that uses the include_external macro so the samples don't need to be included twice (as .html files and .txt files)?

addyosmani:
> What do you think about including the examples for this doc in docs/elements/core-list-examples

Two-cents: given the rest of the samples exist in `samples/`, it would make them easier to locate if we were consistent with where they live.

robdodson:
I'd stick with using the samples dir for now, just to keep it consistent.
If it's becoming annoying we can discuss how to improve it and maybe make
that a future sprint goal.

Also, is there an easier way to show the sample text? Maybe a page that
> uses the include_external macro so the samples don't need to be included
> twice (as .html files and .txt files)?


Would it work to just link to the file in the samples dir on Github? Like
to show the text for the header-app sample I could just link to this file
<https://github.com/Polymer/docs/blob/master/samples/layout-elements/header-app.html>
.

ebidel:
This is how samples show their text:
https://github.com/Polymer/docs/blob/master/_includes/samples/basic-element.html#L15

Basically you need a file that includes it under _includes/

arthurevans:
@arthurevans @marycampione Are the only authors on these commits I believe...?

arthurevans:
LGTM. Only googlebot is confused.

Issue #804, opened 2014-12-05
bind repeat ref


gertcuykens:
(https://www.polymer-project.org/docs/polymer/binding-types.html)

Having trouble understanding the difference between?

``` html
<template ref="myTemplate">
<template bind ref="myTemplate">
```

Also this is confusing when to use bind?

``` html
<template ref="{{myTemplate}}">
<template bind ref="{{myTemplate}}">
```

Finally I wonder if the following has the same result?

``` html
<template bind>
<template is="auto-binding">
```

Note that it am thinking in the same logic behind the layout attribute.

PR #841, opened 2014-12-18
Clarify use of `ref` attribute and fix examples.

arthurevans:
Related to #804.


addyosmani:
Good addition. It took me a while to realize if you're using `ref`, you need explicit scope with `repeat` in order to make it work as expected.

PR #850, opened 2014-12-29
Updating markup as the card image

yourwebmaker:
I guess some users may get confused when the shown image is not the same as the followed markup code.


addyosmani:
On reviewing, I agree. It's much easier to transition in to the `post-card` example when the text matches the preview.

PR #857, opened 2015-01-02
Shouldn't the content be within the <span> ?

vitoc:


ebidel:
It doesn't need to be. I think this setup was intended. Thanks for the PR though!

PR #861, opened 2015-01-04
Remove extraneous commands

mckoss:
We create (and never use) a development directory.  And then cd .. to the parent of the repo - where the wct command will fail.


ebidel:
LGTM

PR #863, opened 2015-01-05
Restore the development directory

mckoss:
After my previous patch was merged, I noticed a problem I was not aware of.  This repo uses '..' in the .bowerrc file so that it installs dependencies in the PARENT of the repo directory.  This patch replaces the development directory, and also cd's to it so that the bower installed dependency files are not placed in a directory we did not create.


ebidel:
Good catch. I should have looked this over more accurately. I knew we did the .bowerrc trick in the seed-element guide, but didn't think we used it here as well.

PR #869, opened 2015-01-10
Fix URL in "using element" link

kappuccino:
/docs/start/using-elements.html ends with a 404 error
/docs/start/usingelements.html is the good page


kappuccino:
CLA ok

addyosmani:
Verified https://www.polymer-project.org/docs/start/using-elements.html is currently 404'ing. Change LGTM.

PR #870, opened 2015-01-13
Manage Ruby deps (e.g. Jekyll) with Bundler

ssorallen:
This is a more maintainable way of managing Ruby dependencies for
the site. When prepping to run the site locally, this adds a
`bundle install` step.

* Lock Jekyll to ~>2.4 as noted in the README because the Grunt task
  expects certain files from Jekyll that were moved in 2.5.

[1] http://bundler.io/


robdodson:
LGTM! @ebidel @arthurevans are you ok with this change? It should make it easier to manage our various ruby deps.

ssorallen:
By the way, I used this to run the site locally rather than copy/paste the `gem install` command that was listed in the README. The site does indeed need Jekyll 2.4.0 (and not 2.5.x).

arthurevans:
Yep, LGTM. Should make it easier for folks to play along at home..

addyosmani:
Changes LGTM too. cc @ebidel on the merge.

PR #871, opened 2015-01-13
Reset `<body>` scrolling when leaving mobile view

ssorallen:
The “noscroll” class is added when the drawer is opened, and so when the
view changes from mobile when the drawer is open the “noscroll" needs to
be removed to re-enable scrolling on the document.

Fixes https://github.com/Polymer/polymer/issues/1089


ebidel:
A corner case, but LGTM.

PR #872, opened 2015-01-13
Changed in learn-tabs of webcomponents.js for minified version

mateusortiz:
changed in learn-tabs of webcomponents.js for minified version webcomponents.min.js to show that using the minified version of polyfill, you gain performance. :smile:


robdodson:
@kevinpschaaf any objections?

kevinpschaaf:
SGTM

PR #873, opened 2015-01-14
update core-signals naming convention

innubili:
I've spent over 1/2 day debugging a core-signal named userLoggedIn before discovering that the same code with the core-signal named 'user-logged-in' (instead on 'userLoggedIn') worked just fine...

### DIFF
--- a/articles/communication.md
+++ b/articles/communication.md
@@ -422,7 +422,7 @@ elements**.
 
 Your element fires `core-signal` and names the signal in its payload:
 
-    this.fire('core-signal', {name: "foo", data: "Foo!"});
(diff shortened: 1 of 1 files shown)


innubili:
Done :-)

arthurevans:
This is a good catch, but the explanation isn't quite right. You can fire the event, but when you use the on-_eventname_ declarative format, _eventname_ is converted to lowercase (because attributes aren't case-sensitive). (In other words, if you use addEventListener('core-signal-userLoggedIn' on the core-signals element, you should get the event.)

So I'm going to leave in your note here, but also add this note to the section on declarative event mapping, because it's not mentioned there, either, and we want to avoid having other people get bit by this. 

Thanks for the PR!

PR #890, opened 2015-02-06
Don't AJAXify middle mouse button clicks.

### LINKED ISSUE #509: Support middle mouse button to open new tab
Clicking the middle mouse button / wheel on a link usually opens the target of the link in a new tab. It is a convenient way to manage a "to read list" of references without having to interrupt reading the current page.

On http://polymer-project.org, however, this does not work for all links. Most notable middle-clicking on an item in the the navigation column on the left hand side will navigate the current page.

arthurevans:
In the better-late-than-never department, fixes #509. Don't AJAXify step for middle-mouse-button clicks so the browser can open them in a new tab.

Tested on Chrome and Safari for Mac (both of which display the issue on pp.org currently), plus Firefox for Mac and IE 11 (which don't display the issue, and don't appear to be impacted by the change).

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -194,7 +194,7 @@ function ajaxifySite() {
 
   document.addEventListener('click', function(e) {
     // Allow users to open new tabs.
-    if (e.metaKey || e.ctrlKey) {
+    if (e.metaKey || e.ctrlKey || e.which == 2) {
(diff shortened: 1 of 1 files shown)


ebidel:
:mouse:

PR #900, opened 2015-02-17
Fix broken ssl links

ebidel:
R: all


addyosmani:
This one is fine too. Landing.

PR #901, opened 2015-02-18
Updated non working core-media-query example

DMTSource:
This documentation page:
https://www.polymer-project.org/docs/polymer/layout-attrs.html
makes mention of how to use the core-media-query to auto select between a vertical and horizontal alignment. However, when just dropping this code into a page for testing, it did not seem to work(nor does it appear to work on the documentation page itself on my chrome browser).

I searched and found that wrapping the core-media-query block in (template is="auto-binding") (/template) got it working just fine. Source:
http://stackoverflow.com/questions/27298131/polymer-core-media-query-element-for-responsive-layouts

### DIFF
--- a/docs/polymer/layout-attrs.md
+++ b/docs/polymer/layout-attrs.md
@@ -121,14 +121,16 @@ Otherwise, the layout becomes horizontal and the elements are laid out
 normally.
 
 {% raw %}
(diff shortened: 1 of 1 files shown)


DMTSource:
I have just signed the CLA as an individual.

ebidel:
:+1:

PR #905, opened 2015-02-22
Versioned URLs

ebidel:
R: all

Sorry this is so large, but it touches just about everything we have. Most of the update is correctly URLs on pages to be relative.

How this works...In _config.yaml, there's a new `default_version: 0.5` property that we can set to change which version the docs should serve by default. For now it's 0.5 (e.g. `polymer-project.org/0.5/`). If someone hits an unversioned URL, they get redirected to the default version. Everything under a version directory is self contained and will have its own bower.json. That way, demos and samples will run under the correct version.

Still left to do:
- [x] Update vulcanize build rules.
- [ ] Version switcher - https://github.com/Polymer/docs/issues/909
- [ ] Script to `bower update` each version directory. - https://github.com/Polymer/docs/issues/910
- [ ] if users hit an unknown version, they get a 404 page.
- [x] support for /latest

I'll probably add the others in a separate PR since we don't have other doc sets atm.


.gitignore
```diff
@@ -1,6 +1,6 @@
 node_modules
 bower_components
-/components
+components
```
arthurevans: I think as written this will ignore new files added to some existing directories, such as:

https://github.com/Polymer/docs/tree/master/samples/components

I don't know if there's a good way to write the .gitignore rule so it skips only the versioned component directories, like, `/[0-9.]*/components`, or whether we should rename the existing directories...

ebidel: Done.

0.5/docs/polymer/polymer.md
```diff
@@ -338,7 +330,7 @@ This second version of `app-globals` has a slightly different API than
 the first. The global variables are properties of the `values` object instead of
 direct properties of `app-globals`. Setting values using attributes imposes two
 limitations: the values must be strings, and the variable names are lowercase.
-(See [Attribute case sensitivity](#attrcase) for more information.)
+(See [Attribute case sensitivityy](#attrcase) for more information.)
```
arthurevans: typo (extra y)

ebidel: Done

addyosmani:
@ebidel Do we want to land support for the `/latest` set of redirects as part of this PR or should we keep that separate?

ebidel:
Done. PTAL

main.py
```diff
@@ -1,9 +1,73 @@
...
+  return default_version
+
+
+class VersionHandler(webapp2.RequestHandler):
+
+  def get(self, version=None):
+    version_dir = get_default_polymer_version()
+    if self.request.path.startswith('/latest'):
+      path = self.request.path.replace('/latest', '/%s' % version_dir)
+      return self.redirect('%s' % path)
```
arthurevans: For the URLs to be bookmarkable, we actually need to redirect the other way -- that is, the latest (or unversioned) link needs to show up in the URL bar. Looking at the app engine config, it looks like that's more complicated. And we don't really need it until we launch 0.8, at least. So unless someone sees something that would prevent us changing this behavior in the future, maybe we should merge this PR and iterate?

app.yaml
```diff
@@ -1,5 +1,5 @@
 application: polymer-project
-version: 0-5-5
+version: zerodoteight
```
robdodson: is this a temporary placeholder?

ebidel: Yes. So we don't upload it to the default version before it's ready.

main.py
ebidel: I see. So we can do that pretty easily with URL rewriting. I agree that we don't need to support this until 0.8 is actually a folder we have.

ebidel:
Here we go!

PR #911, opened 2015-02-25
Keeping output in sync with respective sample snippet

murtazahaveliwala:
The snippet in polymer/layout-attrs page, under 'full bleed' section, mentions text content as "Fitting a fullbleed body."
However, the linked iframe source was mentioning 'This <body> is fullbleed!'

So chose to update the sample with the text as provided in the documentation page.


addyosmani:
Confirmed the text is out of sync. Thanks!

PR #912, opened 2015-02-25
Fixed few sample snippets in polymer/polymer.html page

murtazahaveliwala:
I was reading through the documentation and found a few things amiss. I thought I could help in fixing it.

So here are few corrections -
1. Fixed 'job' sample snippet
2. Removed 'values' attribute from 'app-globals' example as it wasn't making any difference
3. Added missing trailing quote in computed properties snippet


arthurevans:
LGTM. Thanks for the fixes!

Issue #916, opened 2015-03-02
Add "valueattr" to step 1 of the tutorial


Nevraeka:
`valueattr` is missing in the paper-tabs example. This is based on this issue - https://github.com/Polymer/polymer/issues/1228
cc / @tjsavage

ebidel:
Cleared up in https://github.com/Polymer/docs/commit/78a0acb6d3a3ec173db880b175ea19e8b8674bd7

PR #917, opened 2015-03-02
[#916] Added valueattr to paper-tabs example

Nevraeka:
simple attribute add

### DIFF
--- a/0.5/docs/start/tutorial/step-1.md
+++ b/0.5/docs/start/tutorial/step-1.md
@@ -169,7 +169,7 @@ tabs.
 ...
 &lt;core-toolbar>
 
-  <strong class="highlight nocode">&lt;paper-tabs id="tabs" selected="all" self-end>
+  <strong class="highlight nocode">&lt;paper-tabs valueattr="name" id="tabs" selected="all" self-end>
     &lt;paper-tab name="all">All&lt;/paper-tab>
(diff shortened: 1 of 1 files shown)


arthurevans:
This is a little confusing, but `valueattr` actually defaults to 'name', so this is unnecessary. See `valueattr` docs here:

https://www.polymer-project.org/docs/elements/core-elements.html#core-selector

robdodson:
@arthurevans hm that is confusing, especially for a new user. Maybe we can modify the text a bit.

ex: "selected="all" chooses the first tab **with a matching `name` attribute** as the initially selected tab."

Nevraeka:
I see what you mean

Nevraeka:
:+1 @robdodson . I assume you are referring to the descriptive text, and not the code example. :)

arthurevans:
Sounds good. Ironically, we used to have the valueattr attribute here, but
a prior PR removed it as unnecessary.

Will file a text PR this afternoon (and see if I can clarify the somewhat
vague docs for x-selector.

Nevraeka:
Thanks @arthurevans . Should I cancel this PR then?

robdodson:
Yeah I'm referring to the descriptive text in the tutorial

ebidel:
Cleared up in https://github.com/Polymer/docs/commit/78a0acb6d3a3ec173db880b175ea19e8b8674bd7

PR #928, opened 2015-03-08
Image optimisation

addyosmani:
r: all


ebidel:
+1. We need this baked into grunt yo!

addyosmani:
I could prolly take care of that :)

PR #932, opened 2015-03-09
update deploy script

ebidel:
R: all

### DIFF
--- a/scripts/deploy_site.sh
+++ b/scripts/deploy_site.sh
@@ -42,7 +42,7 @@ compass compile
 
 # Run jekyll and build api docs.
 #grunt docs # grunt docs doesn't work on jekyll 2.0
-grunt vulcanize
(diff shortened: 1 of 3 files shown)


scripts/deploy_site.sh
```diff
@@ -42,7 +42,7 @@ compass compile
 
 # Run jekyll and build api docs.
 #grunt docs # grunt docs doesn't work on jekyll 2.0
-grunt vulcanize
+grunt vulcanize-elements
```
arthurevans: Why are we changing the target name?

robdodson:
why not replace both lines with `grunt docs`. That will do that plus run doc-merge

ebidel:
Done.

ebidel:
PTAL

robdodson:
LGTM :dancer:

PR #952, opened 2015-03-12
Link fix -- webview.md

arthurevans:
It looks like the sample repo was reorganized since the original doc was written. I think this is what you're looking for, though...?

@ebidel PTAL


ebidel:
Thanks!

PR #958, opened 2015-03-12
Remove build page link & misc cleanup

arthurevans:
image

Still much cruft left on this page, but at least we can remove the broken link.


addyosmani:
Lgtm

PR #965, opened 2015-03-17
Update unit-testing-elements.md

davidmaxwaterman:
Corrected some typos, but also commented on some confusing parts.

Don't merge as is....just a review of the doc.

### DIFF
--- a/0.5/articles/unit-testing-elements.md
+++ b/0.5/articles/unit-testing-elements.md
@@ -57,7 +57,7 @@ The only real piece missing from Mocha are helpers for performing assertions aga
 
 ### Chai
 
-Chai supports a lot of the assertion styles you might want to use (expect, assert, should). The real difference in these styles is their readability and you'll likely pick a style that fits in with your or your teams own preferences. The Chai [docs](http://chaijs.com/guide/styles/) explain their assertion styles in more detail.
+Chai supports a lot of the assertion styles you might want to use (expect, assert, should). The real difference in these styles is their readability and you'll likely pick a style that fits in with your or your team's own preferences. The Chai [docs](http://chaijs.com/guide/styles/) explain their assertion styles in more detail.
 
 Out of the box, web-component-tester also includes [Lodash](https://lodash.com/) to repeat fewer things and [Async](https://github.com/caolan/async) to keep your sanity.
 
@@ -79,7 +79,7 @@ WCT includes a number of helpful utilities meant to ease testing Web Components.
 
 `testImmediate()` allows you to run your test at declaration time before Mocha has begun tests. It's handy when you need to test document initialization. `testImmediate(name, testFn)` accepts a test name and test function. If an argument is accepted to `testFn`, the test will be treated as async, similar to Mocha tests.
 
-Note that with `testImmediate()`, should any errors be thrown asynchronously cannot be tied to your test. If you wish to catch them and pass them to the `done` event instead, a [safeStep()](https://github.com/Polymer/web-component-tester/blob/master/browser/environment/helpers.js#L19) utility is available to help.
+Note that with `testImmediate()`, should any errors be thrown asynchronously they cannot be tied to your test. If you wish to catch them and pass them to the `done` event instead, a [safeStep()](https://github.com/Polymer/web-component-tester/blob/master/browser/environment/helpers.js#L19) utility is available to help.
 
 ### Mocha helpers
 
@@ -101,7 +101,7 @@ Typically, we write many assertions inside of a single large test. This is on pu
 
 ### `.js` Suites
 
-web-component-tester test suites can either be `.js` sources, which run in the context of your text index. For example, `test/awesome-tests.js`:
+web-component-tester test suites can be `.js` sources, which run in the context of your text index. For example, `test/awesome-tests.js`:
 
 
     suite('AwesomeLib', function() {
@@ -113,7 +113,7 @@ web-component-tester test suites can either be `.js` sources, which run in the c
 
 ### `.html` Suites
 
-Or, you can write tests in separate `.html` documents. For example, `test/awesomest-tests.html`:
+Alternatively, you can write tests in separate `.html` documents. For example, `test/awesomest-tests.html`:
 
     <!doctype html>
     <html>
@@ -143,9 +143,9 @@ To create a new HTML test page:
 
 1. Create a new HTML file in the tests directory (e.g core-selector-basic-test.html). You can use our `<seed-element>` [unit test boilerplate](https://github.com/PolymerLabs/seed-element/blob/master/test/basic-test.html) as a starting point. It already references the relevant web-component-tester dependencies you'll need.
 
-2. Author your tests in the file you created (e.g in `core-selector-basic-test.html`). Some tips are available in the tutorial later on on how to test attributes and events.
+2. Author your tests in the file you created (e.g in `core-selector-basic-test.html`). Some tips are available in the tutorial below on how to test attributes and events.
 
-3. The WCT test runner creates an implicit suite for the entire test file. This means you can have any number of top level `test`s as you would like. That said, you can optionally define a new suite for a set of tests around a new element too. For the my-tabs element, this might look as follows:
+3. The WCT test runner creates an implicit suite for the entire test file. This means you can have any number of top level `test`s. That said, you can optionally define a new suite for a set of tests around a new element too. For the my-tabs element, this might look as follows:
 
 
     suite('<my-tabs>', function() {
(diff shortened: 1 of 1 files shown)


0.5/articles/unit-testing-elements.md
```diff
@@ -79,7 +79,7 @@ WCT includes a number of helpful utilities meant to ease testing Web Components.
 
 `testImmediate()` allows you to run your test at declaration time before Mocha has begun tests. It's handy when you need to test document initialization. `testImmediate(name, testFn)` accepts a test name and test function. If an argument is accepted to `testFn`, the test will be treated as async, similar to Mocha tests.
 
-Note that with `testImmediate()`, should any errors be thrown asynchronously cannot be tied to your test. If you wish to catch them and pass them to the `done` event instead, a [safeStep()](https://github.com/Polymer/web-component-tester/blob/master/browser/environment/helpers.js#L19) utility is available to help.
+Note that with `testImmediate()`, should any errors be thrown asynchronously they cannot be tied to your test. If you wish to catch them and pass them to the `done` event instead, a [safeStep()](https://github.com/Polymer/web-component-tester/blob/master/browser/environment/helpers.js#L19) utility is available to help.
```
davidmaxwaterman: I couldn't figure out what this was trying to say...

0.5/articles/unit-testing-elements.md
```diff
@@ -143,9 +143,9 @@ To create a new HTML test page:
 
 1. Create a new HTML file in the tests directory (e.g core-selector-basic-test.html). You can use our `<seed-element>` [unit test boilerplate](https://github.com/PolymerLabs/seed-element/blob/master/test/basic-test.html) as a starting point. It already references the relevant web-component-tester dependencies you'll need.
 
-2. Author your tests in the file you created (e.g in `core-selector-basic-test.html`). Some tips are available in the tutorial later on on how to test attributes and events.
```
davidmaxwaterman: "on on" is just confusing.

0.5/articles/unit-testing-elements.md
```diff
@@ -205,7 +205,7 @@ In this tutorial, we're going to add an existing Polymer element, [`<core-select
 
 First, get `<seed-element>` setup in a new `development` directory (if you haven't already in the Quick Start):
 
-    # Create a new directory to work in
+    # Create a new directory to work in (?)
```
davidmaxwaterman: Is this really necessary? Almost all devs will already have such a directory (probably named something else), and the 'clone' makes a self-contained directory anyway.

0.5/articles/unit-testing-elements.md
```diff
@@ -240,7 +240,7 @@ We begin writing our first test with a new HTML file in the "tests" directory of
 
 * An import for our element (core-selector)
 
-We can repurpose the existing "basic-test.html" file for this purpose. Let's rename it to "core-selector-tests.html" and customize it for `<core-selector>`:
+We can repurpose the existing "basic-test.html" file for this purpose. Let's rename it to "core-selector-tests.html" and customize it for `<core-selector>` (? has it been changed?):
```
davidmaxwaterman: Is it changed? ...or you're just quoting the file contents *before* it's changed?

0.5/articles/unit-testing-elements.md
```diff
@@ -269,7 +269,7 @@ Give it an ID so we can easily reference it amongst other instances on the page.
 Next, we'll query the DOM for the "selector1" element we just included included:
 
     <script>
-      var s = document.querySelector('#selector1');
+      var s = document.querySelector('#selector1'); // why not byid?
```
davidmaxwaterman: I often wonder why people use querySelector for IDs...but I suppose it's a side-issue.

0.5/articles/unit-testing-elements.md
```diff
@@ -308,7 +308,7 @@ We can now run the `wct` command to execute the tests written above. If all goes
 
 <img src="images/unit-testing-elements/wct-tests-passed.png" alt="Tests passes"/>
 
-How about testing if an attribute is the default value we expect it to be? `<core-selector>` supports a multi attribute in case you want to support multiple items being selectable. Let's add this before `done();` along with our other assertions:
```
davidmaxwaterman: "multi attribute" doesn't scan well...since the reader naturally joins those two words into "multi-attribute".

0.5/articles/unit-testing-elements.md
```diff
@@ -378,7 +378,7 @@ We'll go with the setup/teardown approach for this example, by adding a new suit
 
     });
 
-So the first thing you may notice is our use `flush()`.  As we covered earlier, `flush` allows us to asynchronously dirty check pending objects are observed and ensures notification callbacks are dispatched accordingly. A synchronous alternative is `[element].deliverChanges()`, but that only works for observers declared directly on the element.
+So the first thing you may notice is our use of `flush()`.  As we covered earlier, `flush` allows us to asynchronously dirty check pending objects are observed (what??) and ensures notification callbacks are dispatched accordingly. A synchronous alternative is `[element].deliverChanges()`, but that only works for observers declared directly on the element.
```
davidmaxwaterman: Re "(what??)" - I can't understand that sentence at all, so I can't even suggest a clearer way to say it.

0.5/articles/unit-testing-elements.md
```diff
@@ -421,11 +421,11 @@ As we can see, when we run `wct` once again we're still all green:
 
 **Go forth and write tests!** They can be simple and powerful.
 
-Remember, testing Web Components isn't vastly different to unit testing the JavaScript components you build everyday. Many of the same best practices apply. You're still working with events, objects and an API.
+Remember, testing Web Components isn't vastly different to unit testing the JavaScript components you build every day. Many of the same best practices apply. You're still working with events, objects and an API.
 
 The beauty of sticking with Mocha and Chai is tests can execute equally well in both the browser and continuous integration.
 
-Thanks for reading and do let us know if you have any questions about unit testing your elements!
+Thanks for reading and do let us know if you have any questions about unit testing your elements! (How/where?)
```
davidmaxwaterman: Some pointers to how questions should be asked would be useful here.

0.5/articles/unit-testing-elements.md
addyosmani: Thanks for catching this. The proposed text is an improvement.

0.5/articles/unit-testing-elements.md
addyosmani: We assumed originally almost all devs would, but after taking some developers through the first draft of the guide it became clear that they preferred it being explicitly called out rather than an assumption. Can we remove the `(?)`?

0.5/articles/unit-testing-elements.md
addyosmani: It hasn't been changed. We're just quoting the file contents prior to it being changed. Maybe we could rephrase this better.

0.5/articles/unit-testing-elements.md
addyosmani: We use qSA for IDs pretty regularly. Given that it'll just call out to the best fast path behind the scenes, I think it's a non-issue.

0.5/articles/unit-testing-elements.md
addyosmani: Good point. We should encourage folks to ask questions about testing on Stack Overflow and if they run into bugs with testing, file them against web-component-tester if the issue is with the tool or seed-element otherwise.

0.5/articles/unit-testing-elements.md
davidmaxwaterman: I've seen this benchmarked and iirc the performance difference is significant....though, I imagine those tests are for large(r) documents. Perhaps it doesn't make any difference in smaller (shadow) DOMs? It'd be nice to know, so we don't have to be concerned...still, a side-issue for this doc, imo.

0.5/articles/unit-testing-elements.md
davidmaxwaterman: Sure, you can remove the '(?)'...I wasn't intending this to be merged 'as is'. I added 'markers' as I read through so I can see where to make comments, since github isn't really set up for this sort of review.

0.5/articles/unit-testing-elements.md
davidmaxwaterman: Fair enough. Is there a page with that on it somewhere (I vaguely recall seeing something like that)? Hrm, perhaps this <https://www.polymer-project.org/0.5/resources/faq.html#contributing>

addyosmani:
Most of these changes have been manually addressed over the past few weeks. I'll do another walk through of the feedback here in case we missed anything. Thanks again for highlighting these issues @davidmaxwaterman!

ebidel:
ping?

arthurevans:
Can we close this out? This is on the 0.5 version, which we probably won't update any further.

davidmaxwaterman:
Yes, might as well close this.

PR #968, opened 2015-03-20
Clean-up of resources no longer used?

addyosmani:
I couldn't find references to the below files in source, so I've dropped them. Hoping dropping a few other redundancies will make cloning the repo a tiny tiny bit faster.

Please check the atomic commit log in case there's any doubts.


ebidel:
LGTM

PR #974, opened 2015-03-24
Set an id on "Providing heights" section.

pr4v33n:
Set an id on **Providing heights** section.


pr4v33n:
CLA signed.

arthurevans:
LGTM. Thanks!

PR #982, opened 2015-04-03
Update intro material

arthurevans:
In this PR:

- Site updates based on Eric's feedback.
- Removed the landing page warning about updating the bower.json file (no longer required with the new name).
- Added notes on the change watcher argument order.
- Added graph for benchmarks.


arthurevans:
@ebidel PTAL

0.8/docs/devguide/feature-overview.md
```diff
@@ -22,16 +22,47 @@ to make it easier and faster to make custom elements that work like standard DOM
...
+      Polymer({
+        is: "element-name",
+
+        // add properties and methods on the element's prototype
+
+        properties: {
+          // declare properties for the element's public API
+          greeting: {
+            type: String,
+            default: "Hello!"
```
ebidel: `value` right?

0.8/docs/devguide/properties.md
```diff
@@ -135,22 +136,27 @@ for more information.
...
-dashes.
+properties back to attributes, {{site.project_title}} maps attribute names to property
+names and the reverse. 
+
+When mapping attribute names to property names:
 
-When mapping attribute names to property names, the general rule is that
-attribute names are converted to lowercase (since the DOM is case-insensitive
-for attribute names).
+*   Attribute names are converted to lowercase property names. For example,
```
ebidel: So it's not a requirement that attribute names contain a dash if their name contains more than one word? I thought in my testing this was the vase. Not so?

arthurevans: Hm. I was trying to make this clearer. Here. the _attribute_ name is firstName, and it gets converted to the property name "firstname". (No camel case).

If you want camel case, you need to use a dash in the attribute name. Let me know if you can think of a clearer way to say this.

ebidel: > If you want camel case, you need to use a dash in the attribute name. Let me know if you can think of a clearer way to say this.

This is the key bit I glossed over

ebidel:
LGTM

PR #988, opened 2015-04-06
Add onMutation to watch for children updates.

rickymarcon:
Adding onMutation to watch for children updates in the feature-carousel element.

### DIFF
--- a/0.5/elements/feature-carousel.html
+++ b/0.5/elements/feature-carousel.html
@@ -28,11 +28,17 @@
       interval: null, // num milliseconds to cycle panels.
       selected: 0,
       domReady: function() {
-        // TODO: use onMutation to watch for children updates.
         this.panels = Array.prototype.slice.call(
             this.$.content.getDistributedNodes());
 
+        this.onMutation(this, this.childrenUpdated);
         this.startCycler();
+      },
+      childrenUpdated: function(observer, mutations) {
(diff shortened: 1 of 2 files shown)


0.5/elements/feature-carousel.html
```diff
@@ -28,12 +28,18 @@
       interval: null, // num milliseconds to cycle panels.
       selected: 0,
       domReady: function() {
-        // TODO: use onMutation to watch for children updates.
         this.panels = Array.prototype.slice.call(
```
ebidel: Why don't we just do:

    this.childrenUpdated();

That sets `this.panels` and the mutation observer

0.8/elements/feature-carousel.html
```diff
@@ -28,12 +28,18 @@
       interval: null, // num milliseconds to cycle panels.
       selected: 0,
       domReady: function() {
-        // TODO: use onMutation to watch for children updates.
         this.panels = Array.prototype.slice.call(
             this.$.content.getDistributedNodes());
 
+        this.onMutation(this, this.childrenUpdated);
```
ebidel: same

0.5/elements/feature-carousel.html
rickymarcon: Makes sense. Added.

ebidel:
:+1:

Issue #940, opened 2015-03-10
SEO for Polymer docs are pretty bad


davenotik:
The biggest issue is that it's hard to get to the specific element's page. This should be at the top of search results for specific queries. That it's not wastes seconds each time as I go to the core elements page and hunt through the menus, and it adds up and becomes frustrating. (Ironically, the issue seems to be SPAs such as with Polymer in general.)

Example:

Query: `core-header-panel polymer`
https://www.google.com/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#qscrl=1&q=core-header-panel+polymer

Expected: direct link to: https://www.polymer-project.org/0.5/docs/elements/core-header-panel.html

Related, the top link there is broken: https://www.polymer-project.org/components/core-header-panel/demo.html

ebidel:
We updated URLs yesterday so it's going to take a while for Google Search to re-index the new versioned URLs.

@robdodson also added static pages for element documentation as part of that push. This will hopefully make things a lot better when searching for elements!

davenotik:
Glad to hear! Re: providing static pages alongside the SPA, can you point to that code so I can learn from the practice?

ebidel:
Rob created a node script that generates static files from polymer element documentation. This means it can still be used dynamically at runtime, but you want static docs...boom!

https://github.com/robdodson/doc-merge

https://github.com/Polymer/docs/blob/master/_plugins/element_pages.rb is the jekyll bit.

salamanders:
+1 https://www.google.com/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=polymer%20button first hit is broken.

davenotik:
Yea, got to add a 0.5 now, as per yesterday's blog post: https://blog.polymer-project.org/updates/2015/03/10/updated-site/

So: https://www.polymer-project.org/0.5/components/paper-button/demo.html

ebidel:
Search is smart. Give it some time to recrawl :) We submitted links yesterday to speed up the process.

salamanders:
yay crawling (first hit fixed now.)  Yay me happening to run into this in the small window when it was broken - and it not being a long term issue.  :)

matteocrippa:
little OT, but it could be partially related: anyone tried to manage a SEO test case with a dynamic website developed in polymer against a static one? how google bot manage to render a page due to high use of js?

davenotik:
[...]

PR #991, opened 2015-04-07
Add element name to title for doc pages

robdodson:
Adds the element's name to the `<title>` on docs pages.

Related to https://github.com/Polymer/docs/issues/940#issuecomment-78173315


ebidel:
:+1: good change.

PR #992, opened 2015-04-07
Redirect old element pages

arthurevans:
Feels a little sketchy. But seems to work.


js/app.js
```diff
@@ -174,9 +174,31 @@ function injectPage(url, opt_addToHistory) {
...
+      window.location.hash = '';
+      window.location.pathname = newPath;
+    }
+  }
+}
+
 function initPage(opt_inDoc) {
   var doc = opt_inDoc || document;
 
+  // TODO: surely there's a better way to do this?
```
ebidel: I think we can rock:

    if (location.hash && location.pathname.indexOf('/docs/elements') != -1) {
      location.href = location.href.replace(/(\/docs\/elements\/)#(.*)$/, '$1$2.html')
    }

Havent tested it though

arthurevans: We also use hashes for deep-linking into the page, so we only want to redirect if the page is `elements/#page-name`, not `elements/page-name.html#deeplink`. But will change to the replace syntax.

(Also, I realized I need to deal with deep-link hashes, like #core-list.event-select -- only the first bit should be converted to a file name.)

arthurevans:
Superseded by #996.

PR #996, opened 2015-04-07
Redirect old element pages

arthurevans:
The PR formerly known as https://github.com/Polymer/docs/pull/992, but with clean history.

Now with proper handling of deep-link hashes and less . The following are examples of URLs that should work:

http://localhost:3000/0.5/docs/elements/#core-list
http://localhost:3000/0.5/docs/elements/#core-list.attributes.data

Note that the deep-link hash also needs to contain the element name for it to work 
correctly, so the redirect for the second link should look like this:

http://localhost:3000/0.5/docs/elements/core-list.html#core-list.attributes.data

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -174,8 +174,30 @@ function injectPage(url, opt_addToHistory) {
   xhr.send();
 }
 
+function redirectOldAPIDocs() {
+  // Old API reference URLs have the page name in the hash. After
+  // server-side redirects, they end up as "docs/elements/#page-name".
+  // Rewrite here to "docs/elements/page-name.html"
+  var oldAPILanding = 'docs/elements/'
+  var path = window.location.pathname;
+  var hash = window.location.hash;
+  var position = path.length - oldAPILanding.length;
+  var lastIndex =  path.indexOf(oldAPILanding, position);
+  if (lastIndex !== -1 && lastIndex == position) {
+    if (hash) {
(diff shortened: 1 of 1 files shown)


js/app.js
```diff
@@ -174,9 +174,31 @@ function injectPage(url, opt_addToHistory) {
...
+// in place to preserve the deep link.
+function redirectOldAPIDocs() {
+  var oldAPILanding = 'docs/elements/'
+  var path = window.location.pathname;
+  var hash = window.location.hash;
+  var position = path.length - oldAPILanding.length;
+  var lastIndex =  path.indexOf(oldAPILanding, position);
+  if (lastIndex !== -1 && lastIndex == position) {
+    if (hash) {
+      location.href = location.href.replace(/(\/docs\/elements\/)#([^.]*)(.*)$/, '$1$2.html#$2$3')
```
ebidel: won't this always add a "#'?

arthurevans: Yeah, it will. URLs like:

http://localhost:3000/0.5/docs/elements/#core-list

Become: 

http://localhost:3000/0.5/docs/elements/core-list.html#core-list

The extra hash doesn't do anything, so I didn't add a special case for it. But I could add:

if (hash.indexOf('.')) { ... } else { a different regex } ... ?

ebidel: Disregard my question. I think what you have is fine.

ebidel:
LGTM

PR #998, opened 2015-04-08
Replace itemlast with item.last in x-repeat documentation

fredj:


fredj:
Was already fixed in the code with https://github.com/Polymer/polymer/pull/1313

arthurevans:
LGTM. Thanks for the fix!

PR #999, opened 2015-04-08
Fix missing {%endraw%} tag

fredj:


arthurevans:
Thanks for the fix. Looks like this will expose a missing raw tag in example 3 below. (Leaving this as a note to myself so I don't merge this without fixing the other.)

arthurevans:
LGTM.

arthurevans:
Merged via #1000. Thanks!

PR #1000, opened 2015-04-08
Add missing {%raw%} escapes

arthurevans:
Includes #999 from @fredj.


arthurevans:
@fredj already submitted this commit in the above-mentioned PR (#999), so I think we're OK here and googlebot is just being over-cautious.

Issue #1006, opened 2015-04-09
Registering elements may have chopped off sentence


robdodson:
https://www.polymer-project.org/0.8/docs/devguide/registering-elements.html#custom-constructor

"While the Polymer method returns a basic constructor that can be used to instantiate the custom element, Polymer."

@arthurevans

arthurevans:
Fixed by prior PR.

PR #1008, opened 2015-04-09
Add automatic node finding section.

### LINKED ISSUE #1005: Ready method example could be improved
https://www.polymer-project.org/0.8/docs/devguide/registering-elements.html#ready-method

"Implement ready when it’s necessary to manipulate an element’s local DOM when the element is constructed."

```
ready: function() {
  this.$.ajax.go();
}
```

The above example doesn't really manipulate the DOM (it calls a method on something in the DOM) but the part that caught me was that automatic node finding has not been explained in this part of the 0.8 docs so if someone was reading this as a totally new user to Polymer they might not know what the $ means.

Suggestion: maybe have the example do something like

```
ready: function() {
  this.querySelector('.some-child').setAttribute('aria-hidden', 'true')
}
```

arthurevans:
Added a missing section.

Also fixes #1005, #1006.

### DIFF
--- a/0.8/docs/devguide/experimental.md
+++ b/0.8/docs/devguide/experimental.md
@@ -614,7 +614,7 @@ The Polymer standard layer adds declarative data binding, events, property notif
 
 | Feature | Usage
 |---------|-------
-| [Local node marshalling](local-dom.html#node-marshalling) | this.$.\<id>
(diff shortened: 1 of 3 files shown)


ebidel:
LGTM

PR #1009, opened 2015-04-09
Fixed broken link

eostrom:
The "Shadow DOM polyfill docs" link on https://www.polymer-project.org/0.5/docs/start/tutorial/step-2.html goes to https://www.polymer-project.org/0.5/docs/platform/shadow-dom.html, which is 404. https://www.polymer-project.org/0.5/platform/shadow-dom.html is a URL that exists, so adding `../` to the relative URL should work. But if this HTML is used in other contexts, that might not be the right fix.


eostrom:
Signed CLA.

arthurevans:
LGTM. Thanks for the fix!

PR #1012, opened 2015-04-10
Re-enable ZIP download.

arthurevans:


addyosmani:
LGTM. Side: You might want to separately bump to RC4 now that Dan's cut a new release.

PR #1021, opened 2015-04-17
Correct use of aria-disabled attribute

arthurevans:


0.8/docs/devguide/registering-elements.md
```diff
@@ -234,7 +234,12 @@ Example:
 
 Results in:
 
-    <x-custom role="button" aria-disabled tabindex="0"></x-custom>
+    <x-custom role="button" aria-disabled="true" tabindex="0"></x-custom>
+    
+**Note:** The `aria-disabled` attribute is set to the string "true" rather than 
+the boolean value `true` because ARIA attributes follow a non-standard convention of
```
robdodson: I wouldn't say it's non-standard, I would just say they follow the convention of using string values. I realize the WHATWG spec outlines how attrs should work in html but I don't know enough of the lore to know why aria went this way.

But this is now turning into a confusing example that has more info than a beginner needs. One thought is to replace the example with something that's simpler and demonstrates the proper use of a boolean attr. Then we can carve out a section in the guide (probably a separate page) where we specifically address accessibility. I can sign up to write that

arthurevans:
Addressed in a different PR, closing this one.

PR #1023, opened 2015-04-20
Remove dot from title class name

mleonardallen:
Example code for Cross-scope styling did not work with dot in class name


addyosmani:
:+1:

PR #1025, opened 2015-04-21
Better example for layout attributes

### LINKED ISSUE #1024: Layout classes example should be clearer
Existing example doesn't show the classes on local DOM elements, only on the host, which is confusing.

arthurevans:
Fixes #1024.


ebidel:
LGTM

PR #1027, opened 2015-04-23
Update unit-testing-elements.md

bradleytrager:
The quick start fails unless there is a .bowerrc file or all the paths are changed to point at the default bower_components directory.  This is one way to fix the issue.

### DIFF
--- a/0.5/articles/unit-testing-elements.md
+++ b/0.5/articles/unit-testing-elements.md
@@ -35,6 +35,7 @@ Our boilerplate for new Polymer elements, [`<seed-element>`](https://github.com/
         $ cd development
         $ git clone git://github.com/PolymerLabs/seed-element.git
         $ cd seed-element
+        $ echo '{ "directory": "../" }' > .bowerrc
         $ bower install
         $ npm install -g web-component-tester
         $ wct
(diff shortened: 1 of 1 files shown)


bradleytrager:
I signed it!

addyosmani:
Could you confirm which version of seed-element you were using? The current 0.8 version (in https://github.com/PolymerLabs/seed-element/) no longer includes the .bowerrc file, but the old version did. You can get it from https://github.com/PolymerLabs/seed-element/releases/tag/v0.1.5.

bradleytrager:
I cloned the master branch of seed-element as instructed by the quick-start guide.  At the time it was commit [fb7235f] (https://github.com/PolymerLabs/seed-element/tree/fb7235f94af52ad7abe9ca4e4b43f32690fbf640).  It did not include a .bowerrc file, and that caused my bower_components to be installed at the project root instead of a level above (which is the path that is referenced from all the source files).  So someone not familiar with polymer who just copies the quick start instructions will end up getting 404 errors like I did.

An alternate solution could be to have them checkout the 0.1.5 version that already has the .bowerrc file.  So in that case we could add `git checkout v0.1.5` after `git clone git://github.com/PolymerLabs/seed-element.git`.  Would you prefer doing that instead?

addyosmani:
I think for the 0.5 guide it does make sense to instruct users to directly use that version and we'll provide a better guide for 0.8. Happy to make those changes.

addyosmani:
Thanks!

PR #1037, opened 2015-04-30
fixed the forward button

mehrdadrafiee:
The forward button doesn't look like the other forward buttons on other pages. This is because, it has a ``raises`` class instead of ``raised`` one. See screenshots for details.
(Tested on Safari 8.0.5 and Google Chrome 42.0.2311.135)

screen shot 2015-04-30 at 1 58 31 pm

screen shot 2015-04-30 at 1 58 18 pm


mehrdadrafiee:
I signed it!

arthurevans:
LGTM. Thanks for the fix!

PR #1057, opened 2015-05-15
Fix Plunker examples

hadfieldn:
Bring them up to date with corresponding doc examples & fix broken links.


arthurevans:
Agh! Thanks for the catch.

I need to automate generating the extra versions of these samples.

PR #1070, opened 2015-05-21
strawman api docs on site. Check out 0.9/docs/api.html for the sample

garlicnation:
@arthurevans @addyosmani @robdodson


robdodson:
hey @garlicnation I think we're going to link directly to the gh-pages version of the docs so we don't have to try to squeeze it into the site in an iframe.

arthurevans:
For 1.0, we just linked out. Will see about another approach for the future.

PR #1072, opened 2015-05-22
observers is an array, not an object

jeffposnick:
R: @arthurevans 

It's correct elsewhere in the guide (e.g. https://www.polymer-project.org/0.9/docs/devguide/properties.html#multi-property-observers), but this portion incorrectly states that `observers` is an object.


arthurevans:
[LGTM](http://lgtm.in/i/NrW1)

PR #1090, opened 2015-05-29
Fix bad imports of iron-flex-layout into migration guide

LM450N:


LM450N:
Sorry, it was closed by mistake.

arthurevans:
Thanks for the fix!

PR #1096, opened 2015-05-29
Update setPathValue -> set

jakemac53:
Pretty sure this is incorrect, maybe an old version of the api?


arthurevans:
Right you are! Thanks.

PR #1100, opened 2015-05-30
missing closing quote

kasperpeulen:


kasperpeulen:
ah, haha, I'm not the first one I see :)

PR #1111, opened 2015-05-31
Update local-dom.md

DmitryAuine:
I think it's haven't any sense. 
have it?


arthurevans:
Hmm... Need to rephrase this, but given the discussion of different trees, I think it's important to note that you're always adding to someone's "light DOM" tree.

arthurevans:
This section still hasn't been improved, so merging. Thanks!

PR #1116, opened 2015-06-01
Unifying transform/translate3d description and agruments list with toggleClass/toggleAttribute

nazar-pc:


1.0/docs/devguide/utility-functions.md
```diff
@@ -54,13 +54,13 @@ provides a set of useful convenience functions for instances to use.
 
 *   `isDebouncerActive(jobName)`. Returns true if the named debounce task is waiting to run.
 
-*   `transform(transform, node)`. Applies a CSS transform to the specified node,
-    or this element if no node is specified.
+*   `transform(transform, [node])`. Applies a CSS transform to the specified node,
```
ebidel: It's Google's JS doc style to use `opt_Node`. `[node]` can be confused for an array argument.

@arthurevans we should probably rely on the API docs for these methods: http://polymer.github.io/polymer/

ebidel: nvm on the `[]`. I see we're already using it on this page and we should be consistent with it.

arthurevans: Yep, we need deep-linking for the API docs before we can really point people to them. See:

https://github.com/PolymerElements/iron-component-page/issues/28

Once we get that I'd like to link to the API docs more aggressively.

arthurevans:
LGTM.

PR #1167, opened 2015-06-08
Template: add auto-binding template dom-change event

ebidel:
R: @arthurevans


arthurevans:
I believe it works for dom-if and dom-repeat as well, FWIW.

PR #1168, opened 2015-06-08
Assorted simple a11y fixes.

arthurevans:
Mostly text contrast + alt text.

Changes banner colors. Doing something slightly funky with the articles now (1.0 articles get a different banner color than 0.5 articles). Not sure if we want to do this -- it just looked a little funky to me with the articles being a top-level heading in 1.0 but getting the same color...

The whole banner color thing is a little magic. And the fact that 0.5 articles show up in 1.0 is also a little peculiar.


1.0/index.md
```diff
@@ -5,7 +5,7 @@ title: Welcome
 
 <section id="future" class="main-bg">
   <div class="panel left">
-    <img src="./images/polymer1.0-01.svg">
+    <img src="./images/polymer1.0-01.svg" alt="">
```
robdodson: is this missing alt text?

1.0/index.md
```diff
@@ -216,7 +216,7 @@ title: Welcome
     </firebase-collection>
     <template is="dom-repeat" items="{{data}}">
       <contact-card starred="{{item.starred}}">
-        <img src="{{item.img}}" alt="{{item.name}}">
+        <img src="{{item.img}}">
```
robdodson: why was the alt text removed here?

sass/elements/site-banner.scss
```diff
@@ -84,7 +84,7 @@ polyfill-next-selector { content: ':host header.shrink'; }
 polyfill-next-selector { content: 'header summary'; }
 :host ::content header summary {
   margin-top: 20px;
-  opacity: 0.6;
+  font-size: 20px;
```
ebidel: can we keep the contrast between header and subheader? I think Zach
wanted it.

1.0/index.md
arthurevans: This is intentionally blank alt text, because the image is decorative and doesn't add anything.

If we add the alt text, the screen reader just says, "Image: Polymer 1.0 logo".

1.0/index.md
arthurevans: Per the original bug, this alt text duplicates the name that immediately follows the image, meaning the screen reader reads it as "Image <person name> <person name>".

sass/elements/site-banner.scss
arthurevans: The a11y concern here is that with the opacity: 0.6, the contrast is really, really terrible. I had actually bumped this up to 24px after feedback from @lpalmaro. I can add screenshots if you like.

arthurevans:
Here are the headers using Indigo500 and Pink600 with 24px subheads:

new_indigo500_24pxsub
new_pink600_24pxsub

@zgibson can you take a look and let me know if these work for you?

arthurevans:
Out-of-band LGTMs acknowledged.

PR #1173, opened 2015-06-09
fix mixins passed to @apply in layout attributes

batista:
Since 0.9.1 that passing multiple mixins to @apply doesn't work. As mentioned in Polymer/polymer#1601 and PolymerElements/iron-flex-layout#15


ebidel:
LGTM

PR #1189, opened 2015-06-14
Specify recommended install path in overview

niedzielski:
Highlight the recommended install path in the overview where it's most useful.


ebidel:
LGTM

PR #1206, opened 2015-06-18
changed shadow root to local dom

kasperpeulen:
not totally sure about this, but I've read the docs chronically from top to bottom and I don't think that the term "shadow root" is used or explained before. Therefore I think that the term "local dom" is more appropiate


arthurevans:
You are correct. FYI, we try to use the capitalized DOM in text, since it's an acronym. But it looks like there's at least one other occurence of "dom" in the file, so we'll merge this and update later.

PR #1213, opened 2015-06-19
Custom search prototype -- Do Not Merge

arthurevans:
Hey folks,

This is a prototype of using custom search in the current site. It is VERY rough, but it's a starting point if anyone wants to run with it. DON'T MERGE THIS VERSION, it's fugly.

Major issues with this version:

- styling is just kind of sad.

- search field should stay open when displaying search results.

- search is only implemented on /1.0/. 0.5 would require its own version of the search results page (using `defaultToRefinement="0.5"`), and its version of app bar would need to be updated as well.

Note that an alternate style (using links instead of tabs to select versions) can be selected using `refinementStyle="link"`.

More info on the search elements is here: https://developers.google.com/custom-search/docs/element#overview


addyosmani:
> styling is just kind of sad.

:lipstick: Is this something that would benefit from some input from @zgibson? 

> search is only implemented on /1.0/. 0.5 would require its own version of the search results page (using defaultToRefinement="0.5"), and its version of app bar would need to be updated as well.

Do we have a clear idea of how long we want to keep adding functionality (e.g better search results) for 0.5 vs. focusing on 1.0?

arthurevans:
In this case, search is going to look really broken if we don't support 0.5. 

Adding this to 0.5 is just a matter of copying these changes to the 0.5 versions of the files, and making the one-line change described above. (s/defaultToRefinement="1.0"/defaultToRefinement="0.5"/)

arthurevans:
That level of change would almost undoubtedly require input from @zgibson, though.

PR #1221, opened 2015-06-22
feat(welcome): Use file-download icon

Awk34:
Change welcome section's 'Get Polymer' button icon from `archive` to `file-download`


ebidel:
While the icon description is more appropriate, I'm not really sure this adds much.

Awk34:
Sure it doesn't add much, but it bothered me that it showed the archive icon when I went to the Polymer homepage.

ebidel:
LGTM

PR #1227, opened 2015-06-25
Increase contrast of code samples.

arthurevans:
Sample:

code-sample-contrasty

Affects most code samples, but not the ones on the home page, which are in a separate element.


arthurevans:
@robdodson @ebidel @addyosmani PTAL.

addyosmani:
lgtm. Doubt it does, but do we need a review from Zach here too?

arthurevans:
Easier to get forgiveness than permission. Plus, getting ready for a site push.

PR #1235, opened 2015-06-29
Fixed all the google-map tags

abdonrd:
The google-map element has latitude and longitude properties, not lat and long.


ebidel:
Lgtm

PR #1241, opened 2015-07-02
Update setup and build instructions. 

kaycebasques:
* Updated and expanded the setup and build instructions in `README.md`.
* Changed `"grunt-vulcanize": "*"` to `"grunt-vulcanize": "^0.6.4"`. I believe the latest release, 1.0, released a couple days ago, breaks the build process. I'm not sure if there are repercussions to this for anyone else. 
* In the instructions I instruct the user to install all of the Ruby stuff into the user's home directory. The original instructions just say `gem install bundler` and `bundle install`, which defaults to the System directory of Ruby gems (and requires `sudo`). Installing to system can cause all sorts of problems and I think it's a bad practice that we do not want to encourage. 
* I also instruct them to install version 0.7 of `vulcanize` via `npm -g install vulcanize@0.7`. The original instructions just say `npm -g install vulcanize` which will install the latest (version 1.10), and I'm pretty sure this version also breaks the build process.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ _site
 *.vulcanized.js
 *.sublime-*
 _data/versions
+.bundle
--- a/README.md
+++ b/README.md
@@ -1,26 +1,149 @@
-Polymer docs are mostly in Markdown with some HTML. [Jekyll][jekyll] is used to generate the static HTML for the site. The output is generated into a folder called `_site` and served from Google App Engine.
+Polymer docs are mostly in Markdown with some HTML. [Jekyll][jekyll] is used 
+to generate the static HTML for the site. The output is generated into a 
+folder called `_site` and served from Google App Engine.
 
-## Prereqs and installation requirements
-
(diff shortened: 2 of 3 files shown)


kaycebasques:
A couple other things to note:
* In the instructions I instruct the user to install all of the Ruby stuff into the user's home directory. The original instructions just say `gem install bundler` and `bundle install`, which defaults to the System directory of Ruby gems (and requires `sudo`). Installing to system can cause all sorts of problems and I think it's a bad practice that we do not want to encourage. 
* I also instruct them to install version 0.7 of `vulcanize` via `npm -g install vulcanize@0.7`. The original instructions just say `npm -g install vulcanize` which will install the latest (version 1.10), and I'm pretty sure this version also breaks the build process.

robdodson:
LGTM.

kaycebasques:
R: @ebidel

ebidel:
LGTM

PR #1247, opened 2015-07-06
Fixes #1246

### LINKED ISSUE #1246: Mention layout classes in migration guide
_From @eaflores4 on July 2, 2015 5:46_

Hello I pretty new to Polymer and I was noticing that I could directly apply a iron-flex-layout class to an element, such as class="flex". The migration guide does not show this behavior it only shows that I can include these classes as a mixins. I also noticed that these iron-flex-layout classes that can be directly applied to an element come from a file in the iron-flex-layout folder, iron-flex-layout/classes/iron-flex-layout.html. How is this file being included in my project since I am not including it? Also, what should I use, mixins, direct styling, or both? Thanks. 

_Copied from original issue: PolymerElements/iron-flex-layout#20_

ebidel:
R: @kaycebasques @addyosmani

### DIFF
--- a/1.0/docs/migration.md
+++ b/1.0/docs/migration.md
@@ -623,7 +623,7 @@ in the Developer guide, or use `setAttribute` directly.
 Also, note that the `class` attribute is ignored if it is specified in the `hostAttributes` 
 object.
 
-## Layout attributes replaced by custom properties {#layout-attributes}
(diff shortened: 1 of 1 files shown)


kaycebasques:
LGTM
* Reviewed new content.
* Built website and verified that new content displayed properly.

PR #1255, opened 2015-07-10
Improve get the code page

robdodson:
R: @ebidel @addyosmani @kaycebasques

This currently doesn't say anything about installing additional elements. Let me know if you think it would be worthwhile to add that. Also, if you LGTM these changes I can add them to the 0.5 docs as well.

### DIFF
--- a/1.0/docs/start/getting-the-code.md
+++ b/1.0/docs/start/getting-the-code.md
@@ -31,39 +31,6 @@ title: Get the code
 
 {% include toc.html %}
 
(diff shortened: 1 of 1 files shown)


ebidel:
Much less cruft :) LGTM but I'll wait to merge if others have comments.

danrschlosser:
Much better, it's far less confusing to provide a default option and alternatives rather than lay out a bunch of options.  LGTM

PR #1272, opened 2015-07-23
Add note about dom-if adding spans

esprehn:


arthurevans:
Is this an expected effect of the `<template>` tag? Or something introduced by dom-bind? (And if the latter, is it something that's likely to change at any point?)

addyosmani:
@arthurevans afaik this is dom-bind/Polymer's autobinding system and not expected of raw template. Changes LGTM minus one minor nit @esprehn

addyosmani:
Friendly ping @esprehn

arthurevans:
Will merge and munge.

arthurevans:
Well, crap. I merged this, then I tested and... it appears to no longer be the case. Sigh.

Issue #1275, opened 2015-07-27
Error in example from /devguide/templates.html#dom-bind


tomalec:
After copy'n'pasing example from https://www.polymer-project.org/1.0/docs/devguide/templates.html#dom-bind I get:
```
index.html:29 Uncaught TypeError: Cannot read property 'addEventListener' of null
```

I suspect it's because `<template id="t" is="dom-repeat" items="{{data}}">` is not yet stamped from `<template is="dom-bind">`.

tomalec:
I posted a PR for it https://github.com/Polymer/docs/pull/1276
As `dom-change` event works fine for root `dom-bind` element.

I tried to make it work for nested templates but with no luck

```html
<body>

  <!-- Wrap elements with auto-binding template to -->
  <!-- allow use of Polymer bindings in main document -->
  <template is="dom-bind" id="r">
    List:
    <iron-ajax url="http://..." last-response="{{data}}"></iron-ajax>

    <template id="t" is="dom-repeat" items="{{data}}">
        <div>first:<span>{{item.first}}</span> last:<span>{{item.last}}</span></div>
    </template>

  </template>

</body>
<script>
  var r = document.querySelector('#r');

  // The dom-change event signifies when the template has stamped its DOM.
  r.addEventListener('dom-change', function() {
    // auto-binding template is ready.
    var t = document.querySelector('#t');
    console.log("Root dom-bind works!");

    // The dom-change event signifies when the template has stamped its DOM.
    t.addEventListener('dom-change', function() {
    // dom-repeat template is ready.
      console.log("Does nested dom-repeat work?");
    });
  });
</script>
```

Is there a way to do something like this?

tomalec:
Ok, above will work, once I will add `auto` attribute to `<iron-ajax>`:
```html
<iron-ajax url="/path/to/json/api" last-response="{{data}}" auto></iron-ajax>
```

tomalec:
Fixed by https://github.com/Polymer/docs/pull/1276

PR #1276, opened 2015-07-27
Fix `dom-bind` example at templates devguide

tomalec:
Proposed solution for Polymer/docs#1275


arthurevans:
LGTM. You're correct, this is the way the code was supposed to read.

PR #1282, opened 2015-08-02
Limit search to polymer 1.0 docs

bendavis78:
When searching from within 1.0 docs, I should only see results for Polymer 1.0. This can be done using `site:polymer-project.org/1.0` in the google search query.

### DIFF
--- a/1.0/elements/app-bar.html
+++ b/1.0/elements/app-bar.html
@@ -71,7 +71,7 @@
         if (e.keyCode == 13) { // Enter
           if (sender.value) {
             recordSearch(sender.value);
-            var q = 'site:polymer-project.org+' + sender.value;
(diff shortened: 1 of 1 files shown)


ebidel:
Thanks for the change. I'm indifferent about restricting search to just 1.0 content. A lot of good content is still left in the 0.5 docs that we need to port. @arthurevans has a PR that returns a 0.5 and 1.0 tab in the custom search results. I think something like that might be more friendly.

ebidel:
@arthurevans should we close this now that the 0.5 docs are no longer being crawled?

arthurevans:
There are still some stray "generic" URLs sneaking through at the moment. (Like /docs/polymer/data-binding.html). Since we're no longer trying to show results from both versions, we can safely do this, at least as a short-term patch. Results are strictly better.

PR #1285, opened 2015-08-04
devguide/properties: Fix 'until once' wording (random typo)

t-nelis:
Should be either:

1. [...] are not invoked until all [...]
2. [...] are only invoked once all [...]

This change does (1).

### DIFF
--- a/1.0/docs/devguide/properties.md
+++ b/1.0/docs/devguide/properties.md
@@ -317,7 +317,7 @@ array.
 
 These observers differ from single-property observers in a few ways:
 
-*   Observers are not invoked until once all dependent properties are defined (`!== undefined`).  
+*   Observers are not invoked until all dependent properties are defined (`!== undefined`).  
     So each dependent properties should have a default `value` defined in `properties` (or otherwise 
     be initialized to a non-`undefined` value) to ensure the observer is called.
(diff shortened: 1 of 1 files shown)


ebidel:
LGTM. Just need you to sign the CLA.

t-nelis:
@ebidel I doubt you actually need a signature to merge such a trivial "contrib". Otherwise please do take "credit" if that's really necessary. Or just drop the PR, it's not exactly critical (I was just passing by).

(I was technically at work when I committed that, so IANAL but I believe it would mean you'd need a corporate agreement, and I probably won't bother anyone to get a signature for this.)

ebidel:
I realize it's a pain, but not my rules :\ I believe an individual sign would suffice here and would give you the ability to contribute to other Google-related OS projects. If instead, you want to close the PR, I'm happy to make the changes directly. Anyways, thanks for the PR.

t-nelis:
@ebidel I understand. Some projects have rules about what a "non-trivial" or "legally significant" contribution is ([example](https://www.gnu.org/prep/maintain/maintain.html#Legally-Significant)), you might want to nudge the right people if this kind of thing happens often.

Anyway, I can't sign the individual agreement, sorry. (I really would, but clause 4 voids it AFAICT.)

Do what you must, no worries.

Issue #1301, opened 2015-08-17
😩 Doc searching UX is frustrating.


samccone:
As a newish user to polymer I have found looking over docs to be frustrating, others on my team have also expressed this frustration. Here is the current flow that I have been seeing / experiencing. 

### User Flow

* A user newish to polymer wants to double check 2 way data binding
* ![screen shot 2015-08-17 at 9 31 55 am](https://cloud.githubusercontent.com/assets/883126/9305898/e88f6a7a-44c2-11e5-8ace-7d3500e35607.png)

##### They use the search bar

* ![screen shot 2015-08-17 at 9 32 07 am](https://cloud.githubusercontent.com/assets/883126/9305902/f1c425fe-44c2-11e5-9755-c3e5df8e72ca.png)

##### Oh a new tab... odd not in site (they think)

* ![screen shot 2015-08-17 at 9 32 14 am](https://cloud.githubusercontent.com/assets/883126/9305907/fc83aa78-44c2-11e5-8176-89acfee62284.png)

##### Lets switch to 1.0 via the dropdown

* ![screen shot 2015-08-17 at 9 33 35 am](https://cloud.githubusercontent.com/assets/883126/9305931/19dcfb88-44c3-11e5-947c-e4bf3a339efa.png)
* ![screen shot 2015-08-17 at 9 33 43 am](https://cloud.githubusercontent.com/assets/883126/9305937/22059a90-44c3-11e5-9077-2c6fbb7c0637.png)

##### Nooooo :skull: :hear_no_evil: 

----------

### Potential Fixes

1) scope the search bar per version i.e. (0.5/1.0)
2) fix 0.5 -> 1.0 redirects to go to corresponding doc pages
3) Implement an in site doc search.

Thoughts? Terrible Idea? :goat: :sparkles:

alexpaluzzi:
I am all for anything that keeps 0.5 from popping up for me.

For a lot of people new to Polymer, they might not even notice they are reading old and completely different documentation. The way it's indexed with Google makes it that much worse.

If it were up to me, I would completely move and archive the 0.5 links and make sure it's not crawled anymore. But I'm sure someone has a reason why we can't do that. :)

In short...
+1

samccone:
I think we can do more on this one:

Our goal should be to make finding and consuming docs sooo easy...

ideally making things like this not needed :)
https://chrome.google.com/webstore/detail/polysearch/gchibjlnlbpgcfjpbebnlecbbjndiidj

PR #1302, opened 2015-08-17
Mothball the 0.5 docs

robdodson:
PTAL @ebidel @addyosmani @arthurevans @jeffposnick

Aims to fix https://github.com/Polymer/docs/issues/1301

### DIFF
--- a/robots.txt
+++ b/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Disallow: /0.5/
--- a/app.yaml
+++ b/app.yaml
@@ -19,6 +19,10 @@ handlers:
   script: main.app
   secure: always
 
+- url: /robots.txt
+  static_files: robots.txt
+  upload: robots.txt
+
 # These needs to be in this order and come for the short URL versions.
 # Map unversioned wc.js allowed because it will churn less.
 - url: /webcomponents(\.min)?.js$


robots.txt
```diff
@@ -0,0 +1,2 @@
+User-agent: *
+Disallow: /0.5/*
```
ebidel: just `Disallow: /0.5/` to get the entire directory. http://www.robotstxt.org/robotstxt.html

ebidel:
Thinking we should probably do this now that after seeing https://github.com/Polymer/docs/issues/1301. 

In the first link though, it should [automagically](https://github.com/Polymer/docs/blob/master/app.yaml#L105) get redirected to the latest (1.0) version of that unversioned doc. It's too bad we also changed the urls between 1.0 and 0.5. Users could have changed the URL and have things work :\

arthurevans:
Is the goal here to stop giving people 0.5 results?

If we do this via robots.txt, I don't know how long it takes for stuff to get dropped from the indexes. It could be a long time, and it's not going to be reliable. (i.e., it may cause the 0.5 results to just rot slowly, rather than get removed cleanly in one go).

Through webmaster tools, we can _explicitly_ request that the 0.5 pages be removed from the index... but in either case, if we remove the results from Google's index, the on-site search will stop showing 0.5 results, too.

ebidel:
Good point on the CSE not getting results either. I guess the question we need to answer is what @arthurevans said. Do we care about 0.5 results?

robdodson:
Personally I think it's ok but maybe we can ask the community somehow? Maybe a poll on mailing list or g+?

arthurvr:
> Do we care about 0.5 results?

FWIW I personally don't think we should anymore.

samccone:
> Do we care about 0.5 results?

I would argue it is pretty critical to keep old docs around... pretty much forever

based on analytics I have after doing this for Marionette

http://marionettejs.com/docs/current/
http://marionettejs.com/docs/v1.8.5/

http://marionettejs.com/annotated-src/v2.0.1/

We still get tons and tons of hits and people have reached out to say how much they appreciate them..

robdodson:
@samccone to be clear, we're not talking about deleting the docs but removing them from Google Search. Users will still be able to access them via the dropdown menu. Looking at analytics we can see that even though our site already has a search feature, most folks don't use it, so I don't feel like having it scoped to different doc sets will help that much.

I would wager that the majority of our doc searching comes from Google Search and if folks keep ending up in 0.5 docs by default that's a bummer. We can also hand map the url redirects when you click the dropdown but that doesn't fix the Google Search problem, which is the primary one I think we should target.

ebidel:
0.5 docs will definitely still be available to browse on the site. This is about making 1.0 (and future content) more findable. We'll continue to port articles and other great 0.5 content.

LGTM

samccone:
:+1: kk

addyosmani:
LGTM to mothball. I would do this sooner rather than later so we have time to clean up any remaining weeds we've missed before the summit.

PR #1308, opened 2015-08-19
removed feature-carousel, not used in 1.0

yufengg:
this one is in the 1.0 folder (which I assume means it's getting built into the 1.0 part of the site). The 0.5 folder has its own copy of feature-carousel.


ebidel:
LGTM

PR #1311, opened 2015-08-21
Pin 0.5 dir to ^0.5.0

robdodson:
This is to keep the 0.5 dir from asking if you want to install Polymer 1.0

@arthurevans @ebidel PTAL


robdodson:
Realized the 1.0 dir needs it too because it's using 0.5

ebidel:
LGTM

arthurevans:
LGTM.

PR #1313, opened 2015-08-21
Update "Getting the Code" to instruct user to install latest and greatest

kaycebasques:
@arthurevans

For some reason it was instructing to install 1.0.0 specifically.

### DIFF
--- a/1.0/docs/start/getting-the-code.md
+++ b/1.0/docs/start/getting-the-code.md
@@ -53,7 +53,7 @@ This generates a basic `bower.json` file. Some of the questions, like
 
 The next step is to install {{site.project_title}}:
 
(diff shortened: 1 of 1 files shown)


arthurevans:
So, ^1.0.0 is correct -- this is a  "caret range" which doesn't allow the left-most non-zero digit to change. So ^1.0.0 matches (=>1.0.0 <2.0.0), which is what we generally use. Specific users may want more restricted ranges, though.

addyosmani:
I'm with @arthurevans on the current behaviour being desirable. If a user desires more specific ranges they're likely more familiar with semver and can manually change this per their needs :star:

kaycebasques:
Makes sense. Closing this.

Issue #1310, opened 2015-08-19
Add examples for event retargeting


arthurevans:
At the very least, we should add one for localTarget (which is probably the most frequently used) and maybe path. Inspiration:

local target (gives you the logical target, which is what you "probably" want).

https://github.com/PolymerElements/paper-drawer-panel/blob/master/paper-drawer-panel.html#L507-L516
https://github.com/PolymerElements/iron-form/blob/f0b56d38f46af2095073c3525c5bc274673a9ee0/demo/index.html

root target gives you the "deep" target, so you could use it, for example, if you wanted to intercept any anchor clicks, no matter what shadow tree they lived in (breaks encapsulation, is probably bad, but there you go). Something like:

```
document.addEventListener('click', function(ev) {
  var polymerEv = Polymer.dom(ev);
  var t = polymerEv.rootTarget;
  if (t.nodeType === Node.ELEMENT_NODE && t.localName === 'a') {
    // push state
  }
});
```

path:

https://github.com/Polymer/polymer/blob/master/src/standard/gestures.html#L51-L61

laco0416:
:+1:

sbley:
:+1:

arthurevans:
Fixed by https://github.com/Polymer/docs/pull/1546.

PR #1320, opened 2015-08-27
Add an example of `localTarget`

laco0416:
Add an example of `localTarget` to the Events guide.
Examples about `rootTarget` and `path` are not added because I couldn't imagine easy use cases of them in library-user's situation.

refs. #1310


arthurevans:
Thanks for the PR! Ended up going with a more complete example in https://github.com/Polymer/docs/pull/1546.

PR #1324, opened 2015-08-27
0.8 -> 1.0, also fix implicit change observers

rictic:
They were spelled like propertyNameChanged not propertyName-changed

### DIFF
--- a/1.0/docs/devguide/properties.md
+++ b/1.0/docs/devguide/properties.md
@@ -123,7 +123,7 @@ for more information.
 
 The value is interpreted as a method name to be invoked when the property value 
 changes. Note that unlike in 0.5, <strong>property change handlers must be registered 
-explicitly.</strong> The <code><var>propertyName</var>-changed</code> method will not be 
+explicitly.</strong> The <code><var>propertyName</var>Changed</code> method will not be 
 invoked automatically. See <a href="#change-callbacks">Property change callbacks (observers)</a> 
 for more information.
 </td>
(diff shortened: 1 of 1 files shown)


1.0/docs/devguide/properties.md
```diff
@@ -123,7 +123,7 @@ for more information.
 
 The value is interpreted as a method name to be invoked when the property value 
 changes. Note that unlike in 0.5, <strong>property change handlers must be registered 
-explicitly.</strong> The <code><var>propertyName</var>-changed</code> method will not be 
+explicitly.</strong> The <code><var>propertyName</var>Changed</code> method will not be
```
ebidel: technically, it's whatever you name it

1.0/docs/devguide/properties.md
```diff
@@ -150,7 +150,7 @@ names (for example, if a property is defined using `reflectToAttribute: true`.)
 
 **Compatibility note:** In 0.5, Polymer attempted to map attribute names to corresponding properties.
 For example, the attribute `foobar` would map to the property `fooBar` if it was
-defined on the element. This **does not happen in 0.8** &mdash; attribute to property
+defined on the element. This **does not happen in v1** &mdash; attribute to property
```
ebidel: I think we're using full 1.0

rictic: Aren't we on 1.1 already? I guess 1.0 has slightly better connotations than v1, but it may make the docs feel a bit out of date too. WDYT?

1.0/docs/devguide/properties.md
rictic: I read this as referring to the 0.5 style implicit on change functions rather than any kind of explicitly named function

1.0/docs/devguide/properties.md
ebidel: True. In the docs, I think we're just saying "1.0" is the new version of polymer. @arthurevans wdyt?

arthurevans: For now, that's correct -- it all says 1.0.

Guess we should think about whether those should be 1.x, but for now please go with 1.0 -- it'll make it easier to track down references.

rictic: SGTM. Done!

1.0/docs/devguide/properties.md
arthurevans: That was the original intention, so +1 to the correction. (If I come up with a clearer way to say this, I might revamp this para.)

arthurevans:
LGTM.

Issue #1332, opened 2015-09-02
Getting started with Web Animations Polyfill


reccanti:
An updated article on how to use Web Animations would be really helpful. I tried to get started using [this article](https://www.polymer-project.org/0.5/platform/web-animations.html) from the v0.5 docs, but it makes reference to things like Animation objects which don't seem to be supported.

ebidel:
@samthor any interest in helping us updating the guide?

arthurevans:
Are you including the web animations polyfill directly? It's no longer included with the Polymer library.

It is included with the neon elements, so one option is to use that API: 

https://elements.polymer-project.org/guides/using-neon-animations

If you want to use the polyfill directly, there is some information in the polyfill repo:

https://github.com/web-animations/web-animations-next

recanti:
Hi @arthurevans. I was trying to include the web animations polyfill directly The page only says to include web-animations-next.js. My main issue with the documentation is that most of its examples involve using an Animation() object. However, when I try to do it this way, I get the error "Cannot read property 'cancel' of null'. I posted an [issue](https://github.com/web-animations/web-animations-js/issues/39#issuecomment-136940439) in the web-animations repository and was told that the correct way to do it was to just use Element.animate().

samthor:
@ebidel I'll take this, and update the guide included with 0.5. It's mostly naming changes since that was written.

@recanti for now, check out this codelab I wrote for Google I/O 2015-
http://io2015codelabs.appspot.com/codelabs/web-animations-transitions-playbackcontrol

There's also the [web-animations-codelabs](https://github.com/web-animations/web-animations-codelabs) repo, but it just contains the single same codelab for now.

recanti:
Thanks @samthor. The codelab helped clear up a lot of things.

PR #1336, opened 2015-09-03
updates Web Animations to match modern spec

samthor:
@ebidel - this should solve #1332

### DIFF
--- a/0.5/platform/web-animations.md
+++ b/0.5/platform/web-animations.md
@@ -67,12 +67,14 @@ important ones here: `KeyframeEffect`, `AnimationEffectReadOnly`, `AnimationEffe
 
 A `KeyframeEffect` object defines a single keyframe effect that applies to a single element target. For example:
 
-    var effect = new KeyframeEffect(targetElement,
-        [{left: '0px'}, {left: '100px'}], 2000);
+    var effect = new KeyframeEffect(targetElement, [
+        {transform: 'translate(0)'},
+        {transform: 'translate(100px)'}
+    ], 2000);
 
-Here, the target element's "left" CSS property is modified smoothly from `0px` to `100px` over 2 seconds. It's a subclass of `AnimationEffectReadOnly`, which provides timing configuration.
+Here, the target element's "transform" CSS property is modified smoothly from `0px` to `100px` over 2 seconds. The `KeyframeEffect` is a subclass of `AnimationEffectReadOnly`, which provides timing configuration.
 
-Note! Animating the "left" property is good for simple sites, but it will cause the browser to run layout on every frame. Regardless of your approach to animation, try to stick to animating "tranform" and "opacity" for the greatest performance.
+See the Web Fundamentals [guide to Animations and Performance](https://developers.google.com/web/fundamentals/look-and-feel/animations/animations-and-performance) for more information on the best CSS properties to animate, and other animation best practices.
 
 ### Animating between keyframes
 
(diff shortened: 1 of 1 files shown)


samthor:
@adanilo if you're curious!

0.5/platform/web-animations.md
```diff
@@ -55,36 +53,32 @@ a `<div>` over 0.5 seconds. The animation alternates producing a pulsing effect.
...
 
-The Web Animations model is a description of an engine for animation content on the web. The engine is sufficiently powerful to support CSS Transitions, CSS Animations and SVG Animations.
+The Web Animations model is a description of an engine for animation content on the web. In browsers like Chrome, the engine is sufficiently powerful to support CSS animations.
 
 Web Animations also exposes a JS API to the model. This API defines a number of
 new interfaces that are exposed to JavaScript. We'll go through some of the more
-important ones here: Animations, AnimationEffects, TimingDictionaries, TimingGroups, and AnimationPlayers.
-
-An `Animation` object defines a single animation effect that applies to a single element target. For example:
+important ones here: `KeyframeEffect`, `AnimationEffectReadOnly`, `AnimationEffectTiming`, and `Animation`.
```
ebidel: no more demo?

0.5/platform/web-animations.md
```diff
@@ -55,36 +53,32 @@ a `<div>` over 0.5 seconds. The animation alternates producing a pulsing effect.
...
+        [{left: '0px'}, {left: '100px'}], 2000);
 
-An `AnimationEffect` object controls which CSS properties and SVG attributes are
-modified by an animation, and the values that those properties and attributes
-vary between. AnimationEffect objects also control whether the effect replaces
-or adds to the underlying value.
+Here, the target element's "left" CSS property is modified smoothly from `0px` to `100px` over 2 seconds. It's a subclass of `AnimationEffectReadOnly`, which provides timing configuration.
 
-There are three major kinds of effects: `KeyframeEffect`, `MotionPathEffect`, and `EffectCallback`.
+Note! Animating the "left" property is good for simple sites, but it will cause the browser to run layout on every frame. Regardless of your approach to animation, try to stick to animating "tranform" and "opacity" for the greatest performance.
```
ebidel: we should just show this as the example :) or update the example to use opacity or something else.

ebidel:
@arthurevans we should probably move this into the 1.0 docs now that it's up to date. Where would be a good places?

arthurevans:
My understanding is that the web animations polyfill is no longer shipping with Polymer, so it seems like the connection is mostly historical.

I think it's used in neon-animation, but shouldn't this kind of doc be on the web animations repo?

ebidel:
We haven't shipped the WA polyfill for a long time, even in 0.5 days. It would still be nice to have a place for ancillary new technologies like this on the site. WA is used by the neon-elements: https://elements.polymer-project.org/guides/using-neon-animations and we don't have other info on the API anywhere else.

0.5/platform/web-animations.md
samthor: I've updated the example, and simplified this paragraph - linking to Web Fundamentals (which we do a few places over the Polymer docs).

samthor:
@ebidel @arthurevans I'd like to get these changes in for now, then we can decide where to move it :)

ebidel:
:+1:

arthurevans:
Cool. Thanks @samthor! @ebidel, I'll make a note to bring this up in docs meeting. Another alternative would be for this info to be on, say, MDN, which right now is kind of lacking:

https://developer.mozilla.org/en-US/docs/Web/API/Web_Animations_API

samthor:
I'd love to contribute to MDN! I'm not quite sure how to do that, though.

PR #1338, opened 2015-09-03
Add codelabs - DO MERGE

ebidel:
R: @robdodson @crhym3

This shouldn't be merged until the domain is ready AND we want the codelabs nav item to show up.
The reason we can't iframe is that the code-labs.io doesn't support SSL, so we'd get mixed content warnings.

### DIFF
--- a/summit/app/codelabs.html
+++ b/summit/app/codelabs.html
@@ -1,1 +1,1 @@
-<meta http-equiv="refresh" content="0;URL='http://code-labs.io/polymer-summit'">
+<meta http-equiv="refresh" content="0;URL='http://www.code-labs.io/polymer-summit'">


crhym3:
I suppose you could also do it with [RedirectRoute in main.py](https://github.com/Polymer/docs/blob/c0645f12b8023c89b34bfcf34cd0c99a54387397/main.py#L76-L87) like you do for other paths?

ebidel:
Thought about that. But ideally, I want to iframe this in when it's SSL is ready and having the shell codelabs.html page means we can do whatever we want with it.

crhym3:
Ah I see. Makes sense :+1:

summit/app/index.html
```diff
@@ -67,6 +68,7 @@
               <iron-selector class="wide-only" selected="[[page]]" activate-event="click" attr-for-selected="name">
                 <a name="schedule" class="tab" href="/summit/schedule">SCHEDULE</a>
                 <a name="speakers" class="tab" href="/summit/speakers">SPEAKERS</a>
+                <a name="codelabs" class="tab" href="/summit/codelabs" target="_blank">CODELABS</a>
```
robdodson: Is this going to pop open a new tab but also cause the router to change routes in the previous tab?

ebidel: Done

robdodson:
There's also a button at the bottom of the page that I think will need to link out as well.

screen shot 2015-09-03 at 6 59 13 pm

ebidel:
PTAL

ebidel:
PTAL

robdodson:
LGTM

arthurevans:
Updating title and merging per @ebidel.

PR #1342, opened 2015-09-04
Add note about template before script in core docs

### LINKED ISSUE #1340: Mention the importance of template before script in main docs
This quote from the migration guide (https://www.polymer-project.org/1.0/docs/migration.html#local-dom-template) should be somewhere in the main docs as well.

"The `<script>` tag can be inside or outside of the `<dom-module>` element, but the element’s template must be parsed before the call to Polymer."

kaycebasques:
@robdodson how about this?

Fixes #1340.


kaycebasques:
I split up the paragraph into two sentences because the two thoughts seem separate and I don't want people to miss the part about template before script.

arthurevans:
:+1: LGTM.

PR #1345, opened 2015-09-07
Make svg less tied to font rendering specifics

### LINKED ISSUE #1344: /1.0/images/webcomponents_stack.svg renders poorly in firefox
[/1.0/images/webcomponents_stack.svg](/Polymer/docs/tree/master/1.0/images/webcomponents_stack.svg)

screen shot 2015-09-06 at 3 59 37 pm

Looks like the issue is the multiple tspans have us tied to particular font layout implementation details.

rictic:
Should fix #1344

I'm having trouble reproducing the original bug though. It was first reported privately. Can anyone else reproduce?


rictic:
Hm, perhaps the reason that I'm unable to reproduce the bug locally is that, as it turns out I have Roboto installed in my system fonts. That could do it.

ebidel:
Lgtm

PR #1346, opened 2015-09-08
Change local hosthost:3000 to localhost:5000

chuckh:
gulp serve now uses `http://localhost:5000` for psk.


kaycebasques:
@chuckh why the change? Just curious.

chuckh:
@kaycebasques this is in the master. It will be included in PSK 1.1.0 release tentatively for this Friday. it was part of the PR `<platinum-sw>` retooling #292.

>It changes the default dev environment port to 5000 and dist environment port to 5001.
This ensures that the dev and dist environments will no longer share a service worker.
It makes it less likely that other local unrelated development projects running in the same browser will accidentally pick up a PSK service worker, since 5000/5001 aren't common default HTTP ports.

PR #1350, opened 2015-09-11
Update behaviors.md

matthieubosquet:
If you omit the variable statement, it will trigger an "Uncaught ReferenceError: MyBehaviors is not defined".

BTW, I just signed the CLA (Sep 11, 2015 04:06 PDT).

Cheers,
Matthieu

### DIFF
--- a/1.0/docs/devguide/behaviors.md
+++ b/1.0/docs/devguide/behaviors.md
@@ -90,7 +90,7 @@ particular method for referencing your behaviors. Behaviors created by the {{sit
 team are added to the {{site.project_title}} object. When creating your own behaviors, you should 
 use some other namespace to avoid collisions with future {{site.project_title}} behaviors. For example:
 
-    MyBehaviors = MyBehaviors || {};
+    var MyBehaviors = MyBehaviors || {};
     MyBehaviors.HighlightBehavior = { ... }
 
 ## Extending behaviors {#extending}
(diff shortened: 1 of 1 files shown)


arthurevans:
What environment are you using this in? 

This is intended to be a global reference here, so `var` will not do, but we should be more explicit and specify `window.MyBehaviors`.

matthieubosquet:
Hi Arthur,

I use the 'use strict'; statement.
My environment is Chromium/Ubuntu 14.04

I changed my commit to use a global reference.
I am not sure why you would do that though. It works just fine with the html imports and var. Also, if you want to make it a global reference in that context why is it not so anywhere else (with standard Polymer behaviors).

Therefore when you declare a behavior as an element's property, you would have to do :
behaviors: [window.MyBehaviors.HighlightBehavior]
(...and could skip the html import?)

It makes little sense to me.

Could you please explain the goal of namespacing a set of custom behaviors under window?

Cheers,
Matthieu

arthurevans:
Hi Matthieu,

If I understand correctly, to satisfy strict mode, you only need to use the `window` when creating the object (in other words, you're letting the interpreter know that the you're intentionally creating a global reference).

In the case of the Polymer behaviors, they're namespaced on the `Polymer` object. You don't see this because the `Polymer` object is explicitly attached to `window`:

https://github.com/Polymer/polymer/blob/master/src/lib/polymer-bootstrap.html#L22

This lets you reference `Polymer` (and its behaviors) from anywhere. 

You don't _have_ to use global references, it's just the simplest way to do this. We should add a discussion of other methods (like using [IMD](https://github.com/PolymerLabs/IMD)).

arthurevans:
As modified, LGTM.

PR #1357, opened 2015-09-16
Add playlist.

arthurevans:
Note: schedule.json now includes URLs for individual videos, but they're not added to the schedule page yet.


ebidel:
Lgtm

PR #1359, opened 2015-09-18
Add some details on property config. Fixes #1328.

### LINKED ISSUE #1318: Property docs should mention deserializing JSON values
Currently documented here:

https://www.polymer-project.org/1.0/docs/migration.html#attr

But not, as one might expect, here:

https://www.polymer-project.org/1.0/docs/devguide/properties.html#attribute-deserialization

### LINKED ISSUE #1328: Boolean attributes don't behave as described
If a Boolean attribute is defined as:
``` javascript
properties: {
    active: Boolean
}
```
And a tag is implemented as:
``` html
<example-tag></example-tag>
```
The result is that `active` is `undefined`, not `false` as [described in the documentation](https://www.polymer-project.org/1.0/docs/devguide/properties.html#attribute-serialization).
<hr>
Likewise, if a Boolean value is defined as:
``` javascript
properties: {
    active: {
        type: Boolean,
        value: true
    }
}
```
And used in a tag as:
``` html
<example-tag></example-tag>
```
The value is `true` (not `false` because the attribute isn't assigned), which would run contrary to the same documentation.

It's not clear to me if this is a bug or misleading documentation.

arthurevans:
Also fixes #1318.

### DIFF
--- a/1.0/docs/devguide/properties.md
+++ b/1.0/docs/devguide/properties.md
@@ -167,11 +167,14 @@ directly as the value of the property in the `properties` object; otherwise it
 should be provided as the value to the `type` key in the `properties`
 configuration object.
 
-The type system includes support for Number values, Object and Array values 
+The type system includes support for Boolean and Number values, Object and Array values 
 expressed as JSON, or Date objects expressed as any Date-parsable string 
-representation. Boolean properties set based on the existence of the attribute: 
-if the attribute exists at all, its value is true, regardless of its string-value 
-(and the value is only false if the attribute does not exist).
+representation. 
+
+Boolean properties are set based on the _presence_ of the attribute: 
+if the attribute exists at all, the property is set to `true`, regardless 
+of the attribute _value_. If the attribute is absent, the property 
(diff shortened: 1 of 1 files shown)


1.0/docs/devguide/properties.md
```diff
@@ -233,6 +236,19 @@ encouraged that attributes only be used for configuring properties in static
 markup, and instead that properties are set directly for changes at runtime. 
 {: .alert .alert-info }
 
+### Configuring boolean properties
+
+For a Boolean property to be configurable from markup, it must default to `false`. If it defaults to `true`, you cannot set it to `false` from markup, since the presence of the attribute, with or without a value, equates to `true`. This is the standard behavior for attributes in the web platform.
```
ebidel: standard behavior for attributes -> standard behavior for Boolean attributes

ebidel:
LGTM

addyosmani:
LGTM2

arthurevans:
OK. Merging.

PR #1361, opened 2015-09-18
Typo fixed for array-selector helper description

doertedev:
Small typo here:
Keeping structured data in sync requires that {{site.project_title}} understand the path associations of data being bound.
Keeping structured data in sync requires that {{site.project_title}} understands the path associations of data being bound.

### DIFF
--- a/1.0/docs/devguide/templates.md
+++ b/1.0/docs/devguide/templates.md
@@ -198,7 +198,7 @@ different name for the index property.
 
 ## Array selector (array-selector) {#array-selector}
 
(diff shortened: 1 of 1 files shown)


doertedev:
I signed it!

addyosmani:
Lgtm. Ty!

arthurevans:
Thanks for the PR! The sentence is actually correct as it is -- the clause "Polymer understand the path associations..." is subjunctive, and we drop the 's' or 'es' in the subjunctive 3rd person singular. At least in the US.

That said, if I need that many words to explain it, we should probably just re-write the sentence.

PR #1365, opened 2015-09-23
remove broken link

danielbaak:
prollyfills.org does not exist.


ebidel:
We'll stick with prollyfill for the 0.5 docs. Thanks for the PR though!

PR #1373, opened 2015-10-02
Add details on property config (take 2)

### LINKED ISSUE #1328: Boolean attributes don't behave as described

arthurevans:
Fixes #1328.
These commits got mysteriously reverted after they were merged. Take 2.

### DIFF
--- a/1.0/docs/devguide/templates.md
+++ b/1.0/docs/devguide/templates.md
@@ -198,7 +198,7 @@ different name for the index property.
 
 ## Array selector (array-selector) {#array-selector}
 
-Keeping structured data in sync requires that {{site.project_title}} understands the path
+Keeping structured data in sync requires that {{site.project_title}} understand the path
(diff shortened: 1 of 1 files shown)


ebidel:
LGTM

PR #1383, opened 2015-10-15
Fix for automatic node finding example

johnbradley:
The example was using this.name (which was undefined for me) so it didn't have any effect.
I changed this to this.tagName so it will display 'x-custom'.  I think this was the intent.


johnbradley:
I signed it!

ebidel:
LGTM. Thanks!

PR #1385, opened 2015-10-19
Update settings.md

MartinsThiago:
Inline polymer settings will cause unexpected behaviors when using ES6 Recipe

### DIFF
--- a/1.0/docs/devguide/settings.md
+++ b/1.0/docs/devguide/settings.md
@@ -10,20 +10,27 @@ Document-level global {{site.project_title}} settings can be set
 by creating a `Polymer` object on window before importing the {{site.project_title}}
 library:
 
+First create `scripts/polymer-settings` with this content:
+```javascript
+window.Polymer = window.Polymer || {};
+window.Polymer.dom = 'shadow';
+```
+
+Then reference `scripts/polymer-settings` right before importing `Polymer`
+
+```html
 	<html>
 	<head>
 	  <meta charset="utf-8">
 	  <script src="components/webcomponentsjs/webcomponents-lite.js"></script>
-	  <script>
-        window.Polymer = window.Polymer || {};
-        window.Polymer.dom = 'shadow';
-      </script>
-      <!-- import a component that relies on Polymer -->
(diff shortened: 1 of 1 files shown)


1.0/docs/devguide/settings.md
```diff
@@ -10,20 +10,27 @@ Document-level global {{site.project_title}} settings can be set
 by creating a `Polymer` object on window before importing the {{site.project_title}}
 library:
 
+First create `scripts/polymer-settings` with this content:
+```javascript
+window.Polymer = window.Polymer || {};
```
ebidel: combine: `window.Polymer = window.Polymer || {dom: 'shadow'`};`

1.0/docs/devguide/settings.md
```diff
@@ -41,4 +48,4 @@ Available settings:
 
 * `dom` - options:
     * `shady`. All local DOM is rendered using shady DOM, even where shadow DOM is supported (current default).
-    * `shadow`. Local DOM is rendered using shadow DOM where supported (this will be the default in the future).
\ No newline at end of file
+    * `shadow`. Local DOM is rendered using shadow DOM where supported (this will be the default in the future).
```
ebidel: will -> may

ebidel:
Why does this cause errors? I'd like us not to recommend an external file b/c t saves a network request and is a bit overkill for this.

MartinsThiago:
Because ES6 recipe instruct us to use crisper (which is good)

So when running gulp serve this except of js is extracted and then referenced in the end, then its overwrite `Polymer.dom` function

'I think it's not so bad adding referencing scripts once build will concatenate all js scripts'

ebidel:
So, I don't think we should change this just to support the PSK recipe. Instead, users can move this bit of script around to fit their setup or we can figure out how better to bake this into PSK/ES6 recipe.

MartinsThiago:
Yup, it's an option, can I close it? I'am going to make a PR on ES6 recipe

ebidel:
Let's start with the PR over. We don't want to overcomplicate this settings for users that are not using es6/babel.

PR #1386, opened 2015-10-19
Update homepage catalog image

ebidel:
R: all


samccone:
why the switch from jpg -> png?

arthurevans:
LGTM. 

(JPG is about 10K bigger than the previous image, but a naive conversion to PNG ends up way larger... We should probably take a pass through our images and see what needs optimizing.)

PR #1391, opened 2015-10-23
Update Polymer version to 1.2.0

abdonrd:


ebidel:
The article should stay at 1.1.0 since that's the version it was written
for.

abdonrd:
Updated! :)

ebidel:
Thanks.

Issue #1382, opened 2015-10-14
Doc version selector click area is broken


samccone:
:question: **What**

It is possible to select a dropdown item without changing the page depending on where you click.

:eyes: **Proof:**

![screen shot 2015-10-14 at 3 52 24 pm](https://cloud.githubusercontent.com/assets/883126/10500228/9afd8ce6-728b-11e5-9547-725e96883977.png)

:bug: **Why:**

![screen shot 2015-10-14 at 3 52 30 pm](https://cloud.githubusercontent.com/assets/883126/10500230/9c787374-728b-11e5-8551-5f5599551a0e.png)

Because we have anchor tags inside of the paper-items , it is possible top click on the paper-item and select it without clicking on the anchor tag :boom: 

:wrench: **Fix:**

option 1) Instead of using anchor tags inside of paper-items simply hook into the dropdown api to react when the selected item has been chosen 
option 2) hack the anchor tag to have a negative margin to take up the entire area, so that you have to click on it

-------

cc @ebidel @tjsavage

arthurevans:
I think #1 is an antipattern: the crawler expects navigation links to be anchors (and even if we don't want the 0.5 docs to be indexed, we may at some point have more than one version that we want to index).

But I defer to @robdodson and @ebidel on this.

ebidel:
Think we just get rid of the padding and make the anchors take up the full space using flexbox.

arthurevans:
I tried to fix this and failed. It sounds simple, but I just can't figure out how to get rid of the padding without  totally messing up the styling of the element. I'm kicking this back to the more advanced CSS brains. 

@ebidel @robdodson if one of you can fix this without burning too many cycles, please do. If not, we should defer it for site redesign.

PR #1410, opened 2015-11-03
Close Gh-1382. Remove site switcher margin

robdodson:
Nix padding on site switcher. Should look like this now

screen shot 2015-11-03 at 9 06 22 am


ebidel:
LGTM

PR #1411, opened 2015-11-03
Fixes #580. Fixes #1360

### LINKED ISSUE #580: Explicitly mention that Android browser is not supported
Could I suggest explicitly mentioning that Android browser is not supported on the [browser compatibility page](http://www.polymer-project.org/resources/compatibility.html)? The combination of the following points left me a bit unsure and I had to stumble around for a while to find a definitive answer.

- Being a Google product I had assumed Android browser would be relatively similar to Chrome in terms of features. I assumed, despite being deprecated that the polyfills at least would work for the most part.
- The fact that Android browser has been completely replaced in new phones (like mine) with Chrome means its very hard to do a quick test to see if Polymer will work without finding someone with an older Android device.
- The fact that Android browser still seems to be the [dominant mobile browser](http://gs.statcounter.com/#mobile+tablet+console-browser-ww-monthly-201408-201408-bar) means that this is going to be a relatively large problem for adopters for at least the next couple of years.

### LINKED ISSUE #1360: Compatibility page should include polyfill section first
People are interpreting the chart on native support to mean that Polymer only supports Chrome:

https://www.polymer-project.org/1.0/resources/compatibility.html

The polyfill section should be first, and the compatibility matrix should be inline, not a link, to make it clear which browsers Polymer actually supports today.

ebidel:
R: @arthurevans @kaycebasques

### DIFF
--- a/1.0/resources/compatibility.md
+++ b/1.0/resources/compatibility.md
@@ -5,8 +5,6 @@ shortname: Resources
 title: Browser Compatibility
 ---
 
-TL;DR:
-
 > **{{site.project_title}} 1.0 works in all _latest version_ "evergreen" browsers (Chrome, Firefox, IE, Opera, and Safari)**.
 
 ---
@@ -63,7 +61,7 @@ With the polyfills, {{site.project_title}} works in these browsers:
 
 Notes:
 
-- Android Browser is not supported. Chrome for Android is.
+- **Android Browser is not supported**. Chrome for Android is supported.
 
 
 #### Should I use webcomponents-lite.js or webcomponents.js?
--- a/1.0/resources/compatibility.md
+++ b/1.0/resources/compatibility.md
@@ -5,26 +5,29 @@ shortname: Resources
 title: Browser Compatibility
 ---
 
-Unlike other frameworks, **{{site.project_title}} is intended to work in the _latest
-version_ of "evergreen" browsers**. That is, Chrome, Firefox, IE, Opera and Safari. 
-This means that your favorite browser might not be supported. 
+TL;DR:
 
-The features required by {{site.project_title}} are not supported natively in 
-all of the evergreen browsers, but a set of web components polyfills are available
-that **do** run in the evergreen browsers and provide the support {{site.project_title}}
-requires.
+> **{{site.project_title}} 1.0 works in all _latest version_ "evergreen" browsers (Chrome, Firefox, IE, Opera, and Safari)**.
 
-## Native browser support
+---
+
+## Browser support
+
+The features used by {{site.project_title}} are not (yet) supported natively in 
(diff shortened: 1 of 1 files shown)


ebidel:
ping

arthurevans:
Will get to this this week. Which means, in practice, tomorrow :grin:

1.0/resources/compatibility.md
```diff
@@ -32,53 +35,56 @@ requires.
...
+  <td><paper-checkbox checked></paper-checkbox></td>
 </tr>
 </table>
 
-<p>
-<small>* Includes Edge</small>
-</p>
+Notes:
+
+- Android Browser is not supported. Chrome for Android is.
```
arthurevans: Bold!

1.0/resources/compatibility.md
```diff
@@ -5,26 +5,29 @@ shortname: Resources
 title: Browser Compatibility
 ---
 
-Unlike other frameworks, **{{site.project_title}} is intended to work in the _latest
-version_ of "evergreen" browsers**. That is, Chrome, Firefox, IE, Opera and Safari. 
-This means that your favorite browser might not be supported. 
+TL;DR:
```
arthurevans: Cut

ebidel: Done

1.0/resources/compatibility.md
ebidel: Done

ebidel:
PTAL

1.0/resources/compatibility.md
```diff
@@ -5,26 +5,27 @@ shortname: Resources
...
-## Native browser support
+With the polyfills, {{site.project_title}} works in these browsers:
 
 <table>
 <thead>
-  <tr><th></th><th>Chrome</th><th>Firefox</th><th>IE&nbsp;10+*</th><th>Safari 8+</th><th>Chrome (Android)</th><th>Safari (iOS&nbsp;8.1)</th></tr>
+  <tr><th></th><th>Chrome</th><th>Firefox</th><th>IE&nbsp;10+/Edge</th><th>Safari 8+</th><th>Chrome (Android)</th><th>Safari (iOS&nbsp;8.1)</th></tr>
 </thead>
 <tr>
   <td class="feature-title"><a href="http://www.html5rocks.com/en/tutorials/webcomponents/template/">Template</a></td>
```
arthurevans: Wonder if we even need to list the different features here since it's all the same?

But open to just merging as is.

arthurevans:
LGTM. Shall I merge it?

I'll keep a ticket open to add more polyfill-specific information

ebidel:
SG.

PR #1415, opened 2015-11-06
Correct usage of this.$$

arthurevans:
Fixes Polymer/polymer#2693.

### DIFF
--- a/1.0/docs/devguide/local-dom.md
+++ b/1.0/docs/devguide/local-dom.md
@@ -226,9 +226,9 @@ You can also locate nodes in the local DOM using `querySelector`, `querySelector
 
     var cancelButton = Polymer.dom(this.root).querySelector('#cancelButton');
 
-    // Shorthand for finding local DOM children by ID
+    // Shorthand for finding local DOM children by selector
     // (equivalent to the above):
-    this.$$('cancelButton');
+    this.$$('#cancelButton');
 
 ### Work with light DOM children {#light-dom-children}
 
--- a/1.0/docs/devguide/local-dom.md
+++ b/1.0/docs/devguide/local-dom.md
@@ -226,7 +226,7 @@ You can also locate nodes in the local DOM using `querySelector`, `querySelector
 
     var cancelButton = Polymer.dom(this.root).querySelector('#cancelButton');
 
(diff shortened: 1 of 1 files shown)


1.0/docs/devguide/local-dom.md
```diff
@@ -226,9 +226,9 @@ You can also locate nodes in the local DOM using `querySelector`, `querySelector
 
     var cancelButton = Polymer.dom(this.root).querySelector('#cancelButton');
 
-    // Shorthand for finding local DOM children by ID
+    // Shorthand for finding local DOM children by selector
     // (equivalent to the above):
-    this.$$('cancelButton');
+    this.$$('#cancelButton');
```
ebidel: i wonder if a class would be a better example. It's usually assumed only a single node on the page has an ID.

1.0/docs/devguide/local-dom.md
```diff
@@ -226,9 +226,9 @@ You can also locate nodes in the local DOM using `querySelector`, `querySelector
 
     var cancelButton = Polymer.dom(this.root).querySelector('#cancelButton');
```
ebidel: this should be querySelectorAll right? $$ returns more than one node.

arthurevans: I believe it only returns one:

https://github.com/Polymer/polymer/blob/master/src/standard/utils.html#L28

1.0/docs/devguide/local-dom.md
```diff
@@ -226,9 +226,9 @@ You can also locate nodes in the local DOM using `querySelector`, `querySelector
...
 
-    // Shorthand for finding local DOM children by ID
+    // Shorthand for finding local DOM children by selector
```
arthurevans: Maybe I should change this to 'a local DOM child'?

1.0/docs/devguide/local-dom.md
ebidel: Ah you're. Man, that's pretty confusing. Literally everyone uses `$$` for qSA

ebidel:
LGTM then

arthurevans:
Corrected comment, merging.

PR #1423, opened 2015-11-11
No spaces around em dashes. Because editors.

arthurevans:
Also standardized on using the actual character instead of the HTML entity, because typing `**something important**&mdash;something less important` is fugly and probably one reason we ended up with spaces in the first place.

Also, no hyphens masquerading as em dashes. Are you kidding me? Hyphens?

A sop to the style guide, 73 em dashes and one Oxford comma.

Confession time: I did not attempt to deal with the rogue hyphens in the 0.5 docs, because there are a bunch of them and the hyphens take more manual intervention. This is a future-facing library and I ain't fixing hyphens for folks living in the past—so there!


ebidel:
LGTM

PR #1424, opened 2015-11-11
Fix code samples for sub-property observer docs

### LINKED ISSUE #1384: Declared properties: observing path changes and deep path observation code examples not working
Hi,

I'm using `polymer` (v1.1.5) with `webcomponentsjs` (v0.7.14) in Chrome (v45).

Unless I'm missing something, it looks like the Polymer code examples provided in the "Declared properties" section, under  "[Observing path changes](https://www.polymer-project.org/1.0/docs/devguide/properties.html#observing-path-changes)"  and  "[Deep path observation](https://www.polymer-project.org/1.0/docs/devguide/properties.html#deep-observation)" are not working.

### Observing path changes
```js
Polymer({
  is: 'x-custom',
  properties: {
    user: Object
  },
  observers: [
    'userManagerChanged(user.manager)'
  ],
  userManagerChanged: function(user) {
    console.log('new manager name is ' + user.name);
  }
});
```

This is the result:
```js
document.querySelector('x-custom').user // -> 'undefined'
```

I tried to initialize it differently:
```js
Polymer({
  is: 'x-custom',
  properties: {
    user: {
      type: Object,
      value: function () {
        return {};
      }
    }
  },
  observers: [
    'userManagerChanged(user.manager)'
  ],
  userManagerChanged: function(user) {
    console.log('new manager name is ' + user.name);
  }
});
```

Now: 
```
// -> Uncaught TypeError: Cannot read property 'name' of undefined
```

And even like this:
```js
Polymer({
  is: 'x-custom',
  properties: {
    user: {
      type: Object,
      value: function () {
        return {manager: 'John'};
      }
    }
  },
  observers: [
    'userManagerChanged(user.manager)'
  ],
  userManagerChanged: function(user) {
    console.log('new manager name is ' + user.name);
  }
});
```

Result:
```js
// -> new manager name is undefined
document.querySelector('x-custom').user.manager // -> "John"
document.querySelector('x-custom').user.manager = 'Paul' // -> "Paul"
// document.querySelector('x-custom').user.manager // -> "Paul"
// The property value has been updated, but the change has not been observed and no callback has been called
```

The same holds true for the code examples in **"Deep path observation"**.

@ebidel, @arthurevans, @robdodson, @addyosmani could someone please take a look when you have a sec and let me know if I'm getting something wrong or the code examples need to be fixed? Thanks!

kaycebasques:
Fixes #1384.

hallo @samccone can you code review the new code? Just thought I'd mix it up and interact with someone new :)

hallo @arthurevans I changed the content to be step-by-step instructions, let me know if this works.


1.0/docs/devguide/properties.md
```diff
@@ -370,40 +370,59 @@ Example:
...
 
-You can also observe changes to object sub-properties using the 
-`observers` array, by specifying a full path (`user.manager.name`)
-as a function argument.
+To observe changes in object sub-properties:
 
-Example:
-
-    Polymer({
+1.  Initialize your object to a default value.
```
arthurevans: These aren't really steps (or have a necessary order), so they'd be better as bullet points than numbered items.

#1 here isn't actually necessary. (For example, imagine if your `user` object is only initialized when the user is logged in. You can still set up an observer for that path—it just won't fire until you set the object for the first time.)

kaycebasques: Yes, the numbered list felt a little funky. Thank you.

1.0/docs/devguide/properties.md
```diff
@@ -370,40 +370,59 @@ Example:
...
+2.  Define an `observers` array.
+3.  For each object sub-property that you want to observe, add an element to
```
arthurevans: Generally: might want to use "item" rather than the (more technically correct) "element", just because we use the term "element" so frequently in other contexts in the Polymer doc.

Also might be a little simpler if this list were in terms of a single observer rather than "For each ..."

1.0/docs/devguide/properties.md
kaycebasques: :heavy_check_mark: changed list to unordered
:heavy_check_mark: removed stuff about initializing obj to default val

1.0/docs/devguide/properties.md
kaycebasques: YES! "element" is this context was irking me. Thank you. 

:heavy_check_mark: changed element (in context of arrays) to item
:heavy_check_mark: changed instructions to talk about a single item

1.0/docs/devguide/properties.md
```diff
@@ -370,40 +370,59 @@ Example:
...
+    the `observers` array. Each element must be a method call that 
+    accepts a single argument. The argument is the path to the object
```
arthurevans: method call -> this is difficult one. What you're really specifying is a method _name_, followed by the path you want to observe in parenthesis. I wouldn't emphasize the "single argument" angle -- the observers array can be used to specify path observers, deep path observers, and multiple-property (or path) observers, like;

    observers: [
      'observeSomeStuff(cats,dogs,primates.*,dolphins.bottlenose)'
    ]

In the existing docs, we've used the term "dependent property" (or path would be more accurate) to refer to the stuff inside an observer or computed property definition. The distinction is between the thing being observed (the dependent property/path) and the value sent to the observer method (the argument).

1.0/docs/devguide/properties.md
```diff
@@ -370,40 +370,59 @@ Example:
...
+    sub-property. The method is the action that you wish to take when the 
+    object sub-property changes. The value of the sub-property is passed to
+    the method.
+4.  Define each method in your element registration.
```
arthurevans: registration -> prototype

1.0/docs/devguide/properties.md
```diff
@@ -370,40 +370,59 @@ Example:
...
+                return {};
+              }
+            }
+          },
+          /* Each item of observers array is a method name followed by 
+             a comma-separated list of one or more paths. */
+          observers: [
+            'userNameChanged(user.name)'
+          ],
+          /* Each method referenced in observers must be defined in
```
robdodson: I think elsewhere in the docs we use `//` for comment style

robdodson:
LGTM

1.0/docs/devguide/properties.md

arthurevans:
LGTM.

PR #1429, opened 2015-11-12
Add section on how to handle third-party libs that modify local DOM

### LINKED ISSUE #1372: Styling: document scopedSubtree
A lot of people are running into this issue. I think we should add a section in the styling guide.

When using a third party lib that manipulates the dom (and doesnt call Polymer.dom methods), styling is sometimes not picked up. Calling scopeSubtree works.

https://github.com/Polymer/polymer/issues/2299#issuecomment-133178981

kaycebasques:
Fixes #1372.

### DIFF
--- a/1.0/docs/devguide/styling.md
+++ b/1.0/docs/devguide/styling.md
@@ -699,3 +699,23 @@ that includes a `custom-style` element.
       } 
     </style>      
 
+## Third-party libraries that modify local DOM
+
+If you are using a third-party library that adds local DOM nodes to your
+Polymer element, you may notice that styles on the element do not update
+properly. 
+
+The correct way to add DOM nodes to a Polymer element's local DOM is via
+`Polymer.dom`. This method adds the proper scoping classes
+to the element, which ensures that styles are updated properly. 
+
+Many libaries do not know about this method, so they do not add DOM nodes 
+correctly. Define the `scopeSubtree` method in your element's `ready` method to 
+enable a mutation observer on your element. This mutation observer applies
+proper CSS scoping to the element's local DOM whenever a third-party library
+manipulates the local DOM.
+
+    ready: function() {
+      this.scopeSubtree(this.$.content, true);
+    }
+
--- a/1.0/docs/devguide/styling.md
+++ b/1.0/docs/devguide/styling.md
@@ -710,12 +710,10 @@ The correct way to add DOM nodes to a Polymer element's local DOM is via
 to the element, which ensures that styles are updated properly. 
 
 Many libaries do not know about this method, so they do not add DOM nodes 
-correctly. Define the `scopeSubtree` method in your element's `ready` method to 
-enable a mutation observer on your element. This mutation observer applies
-proper CSS scoping to the element's local DOM whenever a third-party library
-manipulates the local DOM.
+correctly. Call `scopeSubtree` in `ready` to enable a mutation observer on 
+your element's local DOM. This mutation observer detects changes to the 
+local DOM, and applies proper CSS scoping to the element and its descendants.
 
     ready: function() {
       this.scopeSubtree(this.$.content, true);
     }
-
--- a/1.0/docs/devguide/styling.md
+++ b/1.0/docs/devguide/styling.md
@@ -699,21 +699,30 @@ that includes a `custom-style` element.
       } 
     </style>      
 
-## Third-party libraries that modify local DOM
+## Third-party libraries that modify local DOM {#scope-subtree}
 
 If you are using a third-party library that adds local DOM nodes to your
 Polymer element, you may notice that styles on the element do not update
 properly. 
 
 The correct way to add DOM nodes to a Polymer element's local DOM is via
-`Polymer.dom`. This method adds the proper scoping classes
-to the element, which ensures that styles are updated properly. 
+the Polymer DOM API. This API lets you manipulate nodes in a way that respects
+the local DOM and ensures that styles are updated properly.
 
-Many libaries do not know about this method, so they do not add DOM nodes 
-correctly. Call `scopeSubtree` in `ready` to enable a mutation observer on 
-your element's local DOM. This mutation observer detects changes to the 
-local DOM, and applies proper CSS scoping to the element and its descendants.
+When using third-party libraries that do not know about the Polymer DOM
+API, use `scopeSubtree` to apply proper CSS scoping to a node and all of its
+descendants. 
+
+    scopeSubtree(containerNode, false);
+
+`containerNode` is the root node of the tree you wish to scope. Setting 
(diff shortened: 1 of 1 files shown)


1.0/docs/devguide/styling.md
```diff
@@ -699,3 +699,21 @@ that includes a `custom-style` element.
       } 
     </style>      
 
+## Third-party libraries that modify local DOM
+
+If you are using a third-party library that adds local DOM nodes to your
+Polymer element, you may notice that styles on the element do not update
+properly. 
+
+The correct way to add DOM nodes to a Polymer element's local DOM is via
```
arthurevans: Description is close but not quite right—we usually talk about the Polymer DOM API (rather than the `Polymer.dom` method) because the API structure is kind of weird. Briefly, the `Polymer.dom(node)` returns an unnamed object that lets you manipulate `node` in a way that respects local DOM. I think I've called that object a wrapper elsewhere in the docs, although it's not clear to me whether it is or whether it's more like a proxy. Anyway, point being, in _normal_ use, it doesn't actually modify the element itself. 

Whew, that was a long way of suggesting that you shorten:

via Polymer.dom => using the Polymer DOM API, which ensures ...

1.0/docs/devguide/styling.md
```diff
@@ -699,3 +699,21 @@ that includes a `custom-style` element.
...
+`Polymer.dom`. This method adds the proper scoping classes
+to the element, which ensures that styles are updated properly. 
+
+Many libaries do not know about this method, so they do not add DOM nodes 
+correctly. Call `scopeSubtree` in `ready` to enable a mutation observer on
```
arthurevans: Do you want to talk about the 2nd arg?

If you call:

    scopeSubtree(containerNode, false);

It scopes `containerNode`'s current tree, one-time, with no MO.. If you specify `true`, then you get the MO and continuous updates. False is lower overhead, but only useful if you are inserting a bunch of static DOM.

1.0/docs/devguide/styling.md
```diff
@@ -699,3 +699,21 @@ that includes a `custom-style` element.
...
+`Polymer.dom`. This method adds the proper scoping classes
+to the element, which ensures that styles are updated properly. 
+
+Many libaries do not know about this method, so they do not add DOM nodes 
+correctly. Call `scopeSubtree` in `ready` to enable a mutation observer on 
+your element's local DOM. This mutation observer detects changes to the 
+local DOM, and applies proper CSS scoping to the element and its descendants.
+
+    ready: function() {
+      this.scopeSubtree(this.$.content, true);
```
arthurevans: This is a true nit, but I recommend using an ID other than `content` here (like `container`, say). We use this in many places for `<content>` tags in the local DOM, and the suggestion here could set developers off down a very odd path.

1.0/docs/devguide/styling.md
```diff
@@ -699,3 +699,21 @@ that includes a `custom-style` element.
...
+properly. 
+
+The correct way to add DOM nodes to a Polymer element's local DOM is via
+`Polymer.dom`. This method adds the proper scoping classes
+to the element, which ensures that styles are updated properly. 
+
+Many libaries do not know about this method, so they do not add DOM nodes 
+correctly. Call `scopeSubtree` in `ready` to enable a mutation observer on 
+your element's local DOM. This mutation observer detects changes to the 
+local DOM, and applies proper CSS scoping to the element and its descendants.
```
arthurevans: "to the element'' -- which element?

1.0/docs/devguide/styling.md
```diff
@@ -699,3 +699,21 @@ that includes a `custom-style` element.
...
+
+If you are using a third-party library that adds local DOM nodes to your
+Polymer element, you may notice that styles on the element do not update
+properly. 
+
+The correct way to add DOM nodes to a Polymer element's local DOM is via
+`Polymer.dom`. This method adds the proper scoping classes
+to the element, which ensures that styles are updated properly. 
+
+Many libaries do not know about this method, so they do not add DOM nodes
```
arthurevans: Instead of "correctly"—since there's a strong argument that the way third-party libraries are doing this is perfectly correct—might say, "Third-party libraries that are not designed to work with Polymer's local DOM implementation." Or "with the Polymer DOM API"

1.0/docs/devguide/styling.md
```diff
@@ -699,3 +699,21 @@ that includes a `custom-style` element.
       } 
     </style>      
 
+## Third-party libraries that modify local DOM
```
arthurevans: Please add a short ID for this header, like `{#scope-subree}`. I expect people will be linking to it, and we don't want links to break if the header text changes a little later on.

(Background info: at some point, planning to add an containing section on working with third-party libraries & frameworks, so we can cover other cases, like patch-dom. It's possible that might necessitate tweaking this header text.)

arthurevans:
Apologies for putting so many comments on such a small PR! But a lot of it is background info. Hope it makes sense, hit me up if anything doesn't.

1.0/docs/devguide/styling.md
ebidel: +1 on mentioning. I didn't know about the 2nd arg and perf nuts will love it!

1.0/docs/devguide/styling.md
kaycebasques: :heavy_check_mark:

1.0/docs/devguide/styling.md
kaycebasques: :heavy_check_mark:

1.0/docs/devguide/styling.md
kaycebasques: :heavy_check_mark:

1.0/docs/devguide/styling.md
kaycebasques: :heavy_check_mark: I actually thought that this was referring to the content tag and agree that it could have created a rabbit hole

1.0/docs/devguide/styling.md
kaycebasques: :heavy_check_mark: second arg, such useful, much perf

1.0/docs/devguide/styling.md
kaycebasques: :heavy_check_mark:

kaycebasques:
thank you @arthurevans the doc is much cleaner now (and ready for review)

kaycebasques:
hallo @mbaroi can you read this new section and let me know if it makes sense?

arthurevans:
LGTM. :+1:

mbaroi:
@kaycebasques just saw this. I'll take a look at it!

Issue #1414, opened 2015-11-05
Document using dynamic filtering and sorting using `dom-repeat`


arthurevans:
A computed binding that returns a function is the recommended technique. Show how this works.

See also https://github.com/Polymer/docs/issues/1381

arthurevans:
Closed in #1431.

Issue #1191, opened 2015-06-15
"Filtering and sorting lists" needs more advice around `observe`


sjmiles:
https://www.polymer-project.org/1.0/docs/devguide/templates.html#filtering-and-sorting-lists

Suggest we need to add more about the scope of `observe` here, and specifically, about the need to call `render` when non-item-scoped filter/sort criterion changes, as is described in the `dom-repeat` docs themselves.

arthurevans:
Closed in #1431.

PR #1431, opened 2015-11-16
Fixes #1381, #1414, #1191. Update docs for filter API.

### LINKED ISSUE #1381: Correct description of dom-repeat filter API
See: https://www.polymer-project.org/1.0/docs/devguide/templates.html#filtering-and-sorting-lists

arthurevans:
@kevinpschaaf PTAL -- Anything to add or remove on the "Filtering on an index" section? On the dynamic filter and sort section -- should I mention using `render` as a fallback? Or give that its own little section? 

@kaycebasques PTAL, everything else :D


1.0/docs/devguide/templates.md
```diff
@@ -166,6 +168,79 @@ Changing a `manager.type` field should now cause the list to be re-sorted:
...
+
+Because of the way {{site.project_title}} tracks arrays internally, the array 
+index isn't passed to the filter function. Looking up the array index for an
+item is an O(n) operation. Doing so  in a filter function could have 
+**significant performance impact**.
+
+If you want to look up the array index, you can use code like this:
+
+    filter: function(item) {
+      var index = this.items.indexOf(item);
```
kevinpschaaf: To be clear, this is precisely the costly `O(n)` operation referenced above;  as such, `filter` won't do this as a matter of course, but the user could decide to pay the price and write the code you have here.

As is, it sort-of reads that the example code is a way to _avoid_ the **"significant performance impact"**.  L231 could instead read something like "Nevertheless, if array index is required, it can be manually looked up using code like this:".  Slight nuance.

1.0/docs/devguide/templates.md
```diff
@@ -134,8 +134,10 @@ model data that generated a given element. (There are also corresponding
 To filter or sort the _displayed_ items in your list, specify a `filter` or 
 `sort` property on the `dom-repeat` (or both):
 
-*   `filter`. Specifies a filter callback function following the standard 
-    `Array` [`filter`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) API.
+*   `filter`. Specifies a filter callback function, that takes a single argument
+    (the item) and returns true to display the item, false to omit it.
+    (Note that this is **similar** to the standard 
+    `Array` [`filter`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) API, but the callback only takes a single argument.)
```
kevinpschaaf: Could make a little more clear about what's different ala "...but the callback only takes a single `element` argument, and no `index` argument, for performance reasons."

kevinpschaaf:
Aside from inline feedback, LGTM.

1.0/docs/devguide/templates.md
```diff
@@ -166,6 +168,79 @@ Changing a `manager.type` field should now cause the list to be re-sorted:
...
+      <script>
+        Polymer({
+          is: "employee-search",
+          matches: function(string) {
+            if (!string) {
+              // if no search string, show all records
+              return function() { return true; }
+            } else {
+              // return a filter function for the current search string
+              string = string.toLowerCase();
```
kaycebasques: maybe add a comment as a reminder: `// employee var is accessible because this function is called from the context of the dom-repeat`

1.0/docs/devguide/templates.md
```diff
@@ -166,6 +168,79 @@ Changing a `manager.type` field should now cause the list to be re-sorted:
...
+      </template>
+
+      <script>
+        Polymer({
+          is: "employee-search",
+          matches: function(string) {
+            if (!string) {
+              // if no search string, show all records
+              return function() { return true; }
+            } else {
```
kaycebasques: took me a while to figure out what was going on. could summarize with comment: "iterate through each item in `employees` array, and return true if the search string is a substring of `firstname` or `lastname`." that's just me though and I don't know if I represent our audience

1.0/docs/devguide/templates.md
kaycebasques: Actually, it's not a reminder, because this isn't mentioned until after the code sample. I would recommend placing that info before the code sample. That is, if I'm understanding how this data flow even works :smile:

1.0/docs/devguide/templates.md
arthurevans: Yep, I was wondering about that. Thanks, will rephrase a bit.

1.0/docs/devguide/templates.md
arthurevans: Not exactly. But your comment points out to me that `matches` is a terrible function name here, because the function doesn't _match_, anything, it _returns_ a filter function. So with a slight rephrasing: 

          filterFactory: function(searchString) {
            if (!searchString) {
              // if no search string, return a function that matches all records
              return function() { return true; }
            } else {
              // return a filter function for the current search string
              searchString = searchString.toLowerCase();
              return function(employee) {
                var first = employee.firstname.toLowerCase();
                var last = employee.lastname.toLowerCase();
                return first.includes(searchString) ||
                    last.includes(searchString);
              }
            }
          },

So what's happening here? Whenever `searchString` changes, the computed binding `{{filterFactory(searchString}}` is re-evaluated. It returns a Function object, which is assigned as the value of the `dom-repeat`'s `filter` property.

Changing the `filter` property causes the `dom-repeat` to re-filter its items (by testing each one with the new filter function).

arthurevans:
@kevinpschaaf Given that a function like I've written above is somewhat convoluted, and there's some overhead to creating new functions all the time, would it be easier for the user to just have an observer on `searchString` and call `render`?

kevinpschaaf:
I'm guessing that re-sorting/filtering any reasonably sized list (let alone any dom updates thereof) will dwarf the cost of creating the function/closure, but that's just my gut talking.  I like the "computed filter function" construction because it conveys intent more clearly than an observer/render-call with a static filter function that references the observed property off of this.  But that's definitely an alternative.

`computeFilter` might also be a good choice of renaming, since people might be more familiar with the `computeBlah` convention for normal computed properties, and better connect what's going on.

Also note that you can just return `null` from `computeFilter` to disable filtering; more efficient than returning an always-true function.

arthurevans:
Thanks @kevinpschaaf @kaycebasques. Updated (& also added a brief ref to the `render` method, which was missing).

1.0/docs/devguide/templates.md
```diff
@@ -166,6 +171,83 @@ Changing a `manager.type` field should now cause the list to be re-sorted:
...
+            if (!string) {
+              // set filter to null to disable filtering
+              return null;
+            } else {
+              // return a filter function for the current search string
+              string = string.toLowerCase();
+              return function(employee) {
+                var first = employee.firstname.toLowerCase();
+                var last = employee.lastname.toLowerCase();
+                return first.includes(string) ||
```
kevinpschaaf: FYI, `String.prototype.includes` is ES6, and won't work in IE afaik.  Might be better to use ES5 here (`indexOf` or `match` or something like that)...

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

arthurevans:
Updated code sample (& tested on IE this time). Going to go ahead and merge now.

PR #1450, opened 2015-11-28
properties: Use HTML <code> tag in HTML block-level tags

TimothyGu:
Markdown syntax is not supported in block-level tags.

See https://www.polymer-project.org/1.0/docs/devguide/properties.html, and look for <code>`</code>, and you'll see that the current syntax doesn't work.


arthurevans:
LGTM. Thanks for the fix!

PR #1453, opened 2015-12-02
Fix grammar in first sentence of local-dom.md

BenoitPerrot:
Say "an element that is" instead of "that an element is"

### DIFF
--- a/1.0/docs/devguide/local-dom.md
+++ b/1.0/docs/devguide/local-dom.md
@@ -8,7 +8,7 @@ subtitle: Developer guide
 
 {% include toc.html %}
 
-We call the DOM that an element is in charge of creating and managing its _local
+The DOM that an element creates and manages is called its _local
 DOM_. This is distinct from the element's children which are sometimes called
 its _light DOM_ for clarity.
 


1.0/docs/devguide/local-dom.md
```diff
@@ -8,7 +8,7 @@ subtitle: Developer guide
 
 {% include toc.html %}
 
-We call the DOM that an element is in charge of creating and managing its _local
+We call the DOM an element that is in charge of creating and managing its _local
```
ebidel: IMO, it reads better before the change. A `,` after managing might help.

The sentence would read better as a statement:
The DOM that an element creates and/or manages is called its _local DOM_.

robdodson: +1 for Eric's suggestion of `The DOM that an element creates and/or manages is called its local DOM.`

BenoitPerrot:
Ha, indeed! I read the initial sentence too quickly. Let me update patch with your suggestion.

arthurevans:
I'd go with "and" rather than "and/or", otherwise that would be good.

BenoitPerrot:
(done)

PR #1457, opened 2015-12-07
Rewrite "Array observers" section. Also fixes #1427.

### LINKED ISSUE #1427: Documentation for array indexes and `Polymer.Collection`
_From @David-Mulder on November 12, 2015 20:49_

So... after having fought with this A LOT I am just going to put a request for documentation out here:

 - What's the difference between `array.n` and `array.#m`? (I get this part, but as far as I know this isn't documented anywhere)
 - If I have `n`, how do I get `#m`? (Provided that my understanding is correct that `#m` is constant regardless of array manipulation)
 - Maybe some generic advice on when to use what and how.

_Copied from original issue: Polymer/polymer#2725_

kaycebasques:
@arthurevans please review


1.0/docs/devguide/properties.md
```diff
@@ -524,16 +543,29 @@ Example:
...
-registered interest.
-{: .alert .alert-info }
 
-When you specify a wildcard path on an array, the observer is for both splices as
-well as array element sub-property changes.  So the  observer in the
-following example will be called for all additions, removals, and deep changes
-that occur in the array:
+#### Wildcard observers
+
+Wildcard observers watch for both [splices](#splice-observers) and for
```
arthurevans: This section duplicates the #deep-observation section above. We need to re-think this structure.

Perhaps it would be clearer if we reverse the ordering of the deep path and array observation sections, so we have an ordering like this:

-   Observe sub-property changes
    No changes here.
  
-   Observe array mutations
    Describe the special ".splices" path for observing array mutations. Mention that if they're interested in both mutations _and_ property changes to array items, they need  a deep observer.
    Describe the change record format for ".splices" changes.
-   Observe deep sub-property changes
    Here, you can use the `changeRecord.path` property to figure out whether you're seeing a property change, sub-property change, or an array mutation.

kaycebasques:
hallo @arthurevans I re-org'd per your instructions

1.0/docs/devguide/properties.md
```diff
@@ -422,70 +422,42 @@ Example:
...
     </dom-module>
 
-### Deep sub-property observation {#deep-observation}
-
-To call an observer when any (deep) sub-property of an
-object changes, specify a path with a wildcard (`*`).
-
-When you specify a path with a wildcard, the argument passed to your
-observer is a change record object with the following properties:
+### Observe array mutations
```
arthurevans: Be sure to keep the original fragment identifier after the title, so existing links don't break:

    ### Observe array mutations {#array-observation}

It's good practice to always specify a fragment identifier for headings. Otherwise one gets auto-generated from the heading text, and will change if you ever reword the heading.)

1.0/docs/devguide/properties.md
```diff
@@ -422,70 +422,42 @@ Example:
...
 
-*   `path`. Path to the property that changed. 
-*   `value`. New value of the path that changed.
-*   `base`. The object matching the non-wildcard portion of the path. 
+Use a splice observer to call a method whenever an array item is added
```
arthurevans: I would go with "array mutation observer" as the generic term, as used in the heading.

1.0/docs/devguide/properties.md
```diff
@@ -422,70 +422,42 @@ Example:
...
+or deleted via `push`, `pop`, `shift`, `unshift`, or `splice`. Note that 
+these methods are collectively referred to as **splices**.
 
-Example:
+If you want to observe splices and deep sub-property changes on an array,
```
arthurevans: maybe add emphasis on the and. This is the most common case, really.

1.0/docs/devguide/properties.md
```diff
@@ -422,70 +422,42 @@ Example:
...
-When you specify a path with a wildcard, the argument passed to your
-observer is a change record object with the following properties:
+### Observe array mutations
 
-*   `path`. Path to the property that changed. 
-*   `value`. New value of the path that changed.
-*   `base`. The object matching the non-wildcard portion of the path. 
+Use a splice observer to call a method whenever an array item is added
+or deleted via `push`, `pop`, `shift`, `unshift`, or `splice`. Note that 
+these methods are collectively referred to as **splices**.
```
arthurevans: I think this last sentence isn't quite right. The original text that I wrote was also not quite right. Here's the background:

A splice operation (as performed by the `splice` method) removes zero or more elements from a given point in the array, then inserting zero or more new elements in the same location. So (and here's the crucial bit) all of these array mutation methods can be _represented_ in terms of splices.

So, `arr.push(item)` is equivalent to `arr.splice(arr.length,0,item)`. `arr.pop` is equivalent to `arr.splice(arr.length-1,1)`, `arr.unshift(item)` is equivalent to `arr.splice(0,0,item)`, and so on. 

Maybe, "Whenever the array is mutated, the observer receives a change record representing the mutation as a set of array splices." Whew. That's a mouthful, but it's probably more correct.

1.0/docs/devguide/properties.md
```diff
@@ -524,44 +496,68 @@ Example:
...
 
-    Polymer({
+When you specify a path with a wildcard, the argument passed to your
+observer is a change record object with the following properties:
 
-      is: 'x-custom',
+*   `path`. Path to the property that changed. Use this to determine whether
+    a property changed, a sub-property changed, or an array was mutated.
+*   `value`. New value of the path that changed.
+*   `base`. The object matching the non-wildcard portion of the path.
```
arthurevans: Should also mention: 

For array mutations, `path` is the path to the array that changed, followed by `.splices`. And the change record includes the `indexSplices` and `keySplices` properties described in [Observe array mutations](link)

arthurevans:
Finished reviewing. PTAL. Sorry for the delay.

1.0/docs/devguide/properties.md

1.0/docs/devguide/properties.md

1.0/docs/devguide/properties.md

1.0/docs/devguide/properties.md

1.0/docs/devguide/properties.md

kaycebasques:
damn you @googlebot

1.0/docs/devguide/properties.md
```diff
@@ -422,70 +422,44 @@ Example:
...
-object changes, specify a path with a wildcard (`*`).
-
-When you specify a path with a wildcard, the argument passed to your
-observer is a change record object with the following properties:
+### Observe array mutations {#array-observation}
 
-*   `path`. Path to the property that changed. 
-*   `value`. New value of the path that changed.
-*   `base`. The object matching the non-wildcard portion of the path. 
+Use a splice observer to call an array mutation observer whenever an array
```
arthurevans: "Use a splice observer to call an array mutation observer ..." is confusing--they're the same thing. I prefer the term "array mutation observer".

Suggest:
"Use an array mutation observer to call an observer function whenever ..."

arthurevans:
One minor nit & a rebase needed.

1.0/docs/devguide/properties.md

PR #1459, opened 2015-12-08
change bower resolution

garlicnation:


arthurevans:
Fixes API doc issue. LGTM, with the understanding that we need to update to using latest stable + prebuilt docs on the new site.

PR #1466, opened 2015-12-11
Remove duplicated 0.5 components, fix paths.

### LINKED ISSUE #1331: Remove duplicated components
_From @ebidel on September 2, 2015 5:2_

Googling "paper elements demo" produces https://www.polymer-project.org/1.0/components/paper-elements/demo.html#paper-checkbox, but the correct demo is https://elements.polymer-project.org/browse?package=paper-elements

This is no good and very confusing. We need to kill the components under 1.0, unless we're going to update them to 1.0.

_Copied from original issue: Polymer/polymer#2392_

arthurevans:
Fixes #1331

Recent API doc addition broke a few samples. Rather than patch up the existing paths, I bit the bullet and removed the duplicated 0.5 components and moved the (real) 1.0 components to 1.0/components.

The following should now work (but should probably be checked on another machine in case I missed anything):

- home page samples
- 0.5 docs, including element docs
- 1.0 docs (note: plunker button samples have to be tested _after_ deploy, because they link back to the main site to load components).
- Inline samples in Quick Tour of Polymer
- api docs on 1.0/api/

Before testing:

- sync this branch
- remove 1.0/components 1.0/v1components/components
- run scripts/update_bower.sh
- enjoy.


robdodson:
Reviewed. LGTM! Can you rebase?

PR #1471, opened 2015-12-17
data-binding: Fix sample indentation

TimothyGu:


arthurevans:
LGTM. Thanks for the fix!

PR #1474, opened 2015-12-19
add unit test guide

kaycebasques:
@robdodson please do a technical review

@arthurevans can do a writer's review afterwards


robdodson:
Just started reviewing :) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

1.0/docs/devguide/tests.md
```diff
@@ -0,0 +1,367 @@
...
+subtitle: Developer guide
+---
+
+<link rel="import" href="../../../0.5/components/google-youtube/google-youtube.html">
+
+{% include toc.html %}
+
+Use Web Component Tester (`wct`) to unit test your Polymer elements. 
+
+`wct` is an end-to-end testing environment built by the Polymer core team. It
```
ebidel: by the Polymer team/

1.0/docs/devguide/tests.md
```diff
@@ -0,0 +1,367 @@
...
+If you prefer step-by-step instructions, the instructions below 
+teach you how to set up `wct`, create a unit test for
+[`<seed-element>`](https://github.com/PolymerElements/seed-element), and then
+run the test. Note that `<seed-element>` is already set up to support `wct`,
+this is just an example of how to set up your own elements to support `wct`.
+
+1.  Install Web Component Tester globally so that you can run it from 
+    the command line.
+
+        npm -g install web-component-tester
```
ebidel: `npm install -g web-component-tester`

1.0/docs/devguide/tests.md
```diff
@@ -0,0 +1,367 @@
...
+    a demonstration.
+
+1.  `cd` to the base directory of the element.
+
+1.  (Optional) Install and save `wct` locally as a bower component so that your 
+    tests can always import the `wct` runtime.
+
+        bower install --save Polymer/web-component-tester
+
+    The global `wct` that you installed earlier actually serves its own
```
ebidel: If you install `wct` globally, ....

1.0/docs/devguide/tests.md
```diff
@@ -0,0 +1,367 @@
...
+    the path to the directory when you run `wct` (`wct path/to/tests`).
+
+1.  Create a test.
+
+        <!doctype html>
+        <html>
+          <head>
+            <meta charset="utf-8">
+            <script src="../bower_components/webcomponentsjs/webcomponents-lite.js"></script>
+            <script src="../bower_components//web-component-tester/browser.js"></script>
```
ebidel: Double `//`

1.0/docs/devguide/tests.md
```diff
@@ -0,0 +1,367 @@
...
+          </body>
+        </html>
+
+1.  Go to the base directory of your element and run your tests.
+
+        wct
+
+    ![output from successful wct unit test run](/1.0/images/wct-output.png)
+
+    `wct` automatically finds all of the browsers on your system and runs
```
ebidel: mention that `wct -l chrome` runs only Chrome.

1.0/docs/devguide/tests.md
```diff
@@ -0,0 +1,367 @@
...
+    </script>
+  </body>
+</html>
+{% endhighlight %}
+
+You can also use test indexes to configure your tests via query strings
+when `wct` loads them. See [Test Shadow DOM](#shadow-dom) for an example.
+
+## Prevent shared state with test fixtures {#test-fixtures}
+
```
ebidel: Should this be moved up? You use test fixtures in previous exmples.

1.0/docs/devguide/tests.md
```diff
@@ -0,0 +1,367 @@
...
+      myEl = fixture('seed-element-fixture');
+    });
+    test('defines the "author" property', function() {
+      assert.equal(myEl.author.name, 'Dimitri Glazkov');
+    });
+  });
+</script>
+{% endhighlight %}
+
+## Make tests asynchronous
```
ebidel: Asynchronous tests

ebidel:
Should we throw in an example of an event listener?

arthurevans:
Very cool! One meta comment—let's put this under tools/ rather than devguide/. Easier to start out with the new URL rather than move later and break links.

1.0/docs/devguide/tests.md
```diff
@@ -0,0 +1,367 @@
...
+{% include toc.html %}
+
+Use Web Component Tester (`wct`) to unit test your Polymer elements. 
+
+`wct` is an end-to-end testing environment built by the Polymer core team. It
+enables you to test your elements locally, against all of your installed
+browsers, or remotely, via Sauce Labs. `wct` is composed of the following
+tools:
+
+*   [Mocha](https://mochajs.org/) for test framework, complete with support
```
robdodson: Maybe "a test framework" instead?

1.0/docs/devguide/tests.md
```diff
@@ -0,0 +1,367 @@
...
+<div class="yt-embed">
+  <google-youtube
+    videoid="YBNBr9ECXLo"
+    autoplay="0"
+    rel="0"
+    fluid>
+  </google-youtube>
+</div>
+
+If you prefer step-by-step instructions, the instructions below
```
robdodson: "prefer a step-by-step guide, the instructions below..." just to replace the redundant use of 'instructions'

1.0/docs/devguide/tests.md
```diff
@@ -0,0 +1,367 @@
...
+        click: function() {
+          console.log('paper-button.click called');
+        }
+      });
+    });
+
+You don't have to use stubs directly on individual elements. You can override
+the implementation for all elements of a given type. 
+
+## Create stub elements
```
robdodson: You may want to explain a little bit more about why someone would want to create a fake element. Reading the example, it almost seems like you're testing `fake-paper-button`. I think the idea with stubs is to help isolate the code under test. Might be good to ask @cdata if he has any thoughts on these sections

1.0/docs/devguide/tests.md
```diff
@@ -0,0 +1,367 @@
...
+<html>
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" 
+        content="width=device-width, minimum-scale=1.0, initial-scale=1.0, user-scalable=yes">
+  <script src="../../webcomponentsjs/webcomponents-lite.js"></script>
+  <script src="../../web-component-tester/browser.js"></script>
+  <!-- import the element -->
+  <link rel="import" href="../bower_components/iron-ajax/iron-ajax.html">
+</head>
```
robdodson: Do you need all of this markup or could you shorten this down to just the test fixture and the script block?

1.0/docs/devguide/tests.md
```diff
@@ -0,0 +1,367 @@
...
+      });
+      suite('when making simple GET requests for JSON', function() {
+        setup(function() {
+          // get fresh instance of iron-ajax before every test
+          ajax = fixture('simple-get');
+        });
+        test('has sane defaults that love you', function() {
+          request = ajax.generateRequest();
+          server.respond();
+          expect(request.response).to.be.ok;
```
robdodson: test switched to expect here, where previously the article was using assert. Might need to call out Chai somewhere?

1.0/docs/devguide/tests.md
```diff
@@ -0,0 +1,367 @@
...
+      });
+    });
+  </script>
+</body>
+</html>
+{% endhighlight %}
+
+## Control which tests are run
+
+Use a test index to run a subset of suites.
```
robdodson: I'm not sure I see where the index is being used?

robdodson: Oh I didn't realize that was the term for `loadSuites`. I think maybe that should be explained a bit

1.0/docs/devguide/tests.md
```diff
@@ -0,0 +1,367 @@
...
+  'basic-test.html',
+  'basic-test.html?dom=shadow'
+]);
+{% endhighlight %}
+
+Within your tests, you can access Shadow DOM elements using Polymer's DOM 
+API.
+
+    test('click sets isWaiting to true', function() {
+      Polymer.dom(myEl.root).querySelector('button').click();
```
robdodson: Alternatively, it looks like you can also do this:

`myEl.$$('button')` which is much easier. Wish I'd know that when I did the video :P
https://www.polymer-project.org/1.0/docs/devguide/utility-functions.html

1.0/docs/devguide/tests.md
```diff
@@ -0,0 +1,367 @@
...
+## Make tests asynchronous
+
+To create an asynchronous test, pass `done` as an argument to the test function
+and then call `done()` when the test is complete. The `done` argument is a
+signal to Mocha that the test is asynchronous. When Mocha runs the tests it 
+waits until it encounters the `done()` call, and eventually times out if it 
+does not encounter it. 
+
+    test('defines the "author" property', function(done) {
+      setTimeout(function() {
```
robdodson: This would be a good place for an async test. You can listen for the event coming from the fireLasers method or whatever it's called :D /cc @ebidel who mentioned doing an event listener somewhere

1.0/docs/devguide/tests.md
```diff
@@ -0,0 +1,367 @@
...
+<link rel="import" href="../../../0.5/components/google-youtube/google-youtube.html">
+
+{% include toc.html %}
+
+Use Web Component Tester (`wct`) to unit test your Polymer elements. 
+
+`wct` is an end-to-end testing environment built by the Polymer core team. It
+enables you to test your elements locally, against all of your installed
+browsers, or remotely, via Sauce Labs. `wct` is composed of the following
+tools:
```
arthurevans: "is composed of" suggests that these tools comprise the entirety of WCT (and by extension, that all those tools are built by the core team).

I'm having a little trouble summing up, because what exactly WCT is is kind of vague. But I think it's something like this:

Web Component Tester provides a command line interface for running tests (`wct`) and a set of utility functions to make it easier to write tests for web components using a set of popular third-party testing tools:

(Might want to distinguish between Web Component Tester, the package (WCT) and `wct` the command-line utility.)

1.0/docs/devguide/tests.md
```diff
@@ -0,0 +1,367 @@
...
+
+*   [Mocha](https://mochajs.org/) for test framework, complete with support 
+    for BDD and TDD.
+*   [Chai](http://chaijs.com/) for assertions.
```
arthurevans: Is it worth adding for Chai and Sinon that they're adding optional extras. That is, something like:

Chai: adds more assertion types that can be used with Mocha tests.
Sinon: adds test spies, stubs and mocks.

The Sinon one feels extra awkward because I don't expect everyone to know what test spies, stubs and mocks are in the first place. Open to suggestions here.

1.0/docs/devguide/tests.md
arthurevans: From my reading of the README, I think it's shorthand for "an index.html" (that is, to create a set of suites, create an HTML file that calls them all). So the HTML file shown here is the "index"--but that doesn't really make any sense. We should just say that you can create an HTML file to run a subset of suites.

The argument to loadSuites is called `files` in the code (although 'relativeURL` would probably be more accurate):

https://github.com/Polymer/web-component-tester/blob/044c8bd5fe463543b54b5a73c2670625c54b135e/browser/suites.js#L21

1.0/docs/devguide/tests.md
```diff
@@ -0,0 +1,367 @@
...
+      WCT.loadSuites([
+        'basic.html',
+        'async.html'
+      ]);
+    </script>
+  </body>
+</html>
+{% endhighlight %}
+
+You can also use test indexes to configure your tests via query strings
```
arthurevans: Maybe alternately, "The arguments to `loadSuites` are relative URLs. You can configure your tests using query strings in the URLs.See ..."

1.0/docs/devguide/tests.md
arthurevans: I think so. From the tests I've looked at, use of test-fixture is much more common than mocks or stubs, which are kind of advanced topics.

1.0/docs/devguide/tests.md
arthurevans: I don't think "Go watch this Polycast" makes sense as the first sentence of a quick start section. I think the order's backwards. 

Suggest a flow like:

This section shows you how to:

- Set up WCT
- Create a unit test
- Run the unit test

Follow the steps below to get set up, or watch the Polycast:

[polycast embed]

1.0/docs/devguide/tests.md
```diff
@@ -0,0 +1,367 @@
...
+    copy of the `wct` runtime (`browser.js`) whenever it encounters any URL
+    that ends with `web-component-tester/browser.js`. Installing it locally
+    is just a precaution. 
+
+1.  Create a directory for the tests. 
+
+        mkdir test
+
+    When you run `wct` with no arguments, it automatically searches for
+    a directory named `test` and runs any tests it finds in there. You
```
arthurevans: "You can use" sounds more like permission... "If you use ..., you'll need to specify..." sounds a little more like a warning. In this case, I'd prefer the second because standard naming == good.

1.0/docs/devguide/tests.md
```diff
@@ -0,0 +1,367 @@
...
+    can use another name for your directory, but you'll need to specify
+    the path to the directory when you run `wct` (`wct path/to/tests`).
+
```
arthurevans: New para here, I'd add:

   `<seed-element>` and Polymer Starter Kit come complete with basic test suites, so if you're starting from one of
  those projects, you can skip to Step N to run the tests.

1.0/docs/devguide/tests.md
arthurevans: This entry confuses me because the test is asynchronous, but it's testing something that isn't asynchronous. The question being, is this:

Asynchronous tests?
Test asynchronous actions?

I'm not really sure from the example how I'm supposed to use this.

arthurevans: Also, related to async tests, I wonder whether it'd be worthwhile to note something about async dom creation. 

In particular, we've now learned that if you have a dom-repeat or dom-if, you can use Polymer.dom.flush() (or dom-repeat.render()) to force pending async DOM actions so you can test the contents of the dom-repeat/dom-if.

1.0/docs/devguide/tests.md
robdodson: pinging @azakus for the definitive definition :D

1.0/docs/devguide/tests.md
```diff
@@ -0,0 +1,367 @@
...
+{% highlight javascript %}
+WCT.loadSuites([
+  'basic-test.html',
+  'basic-test.html?dom=shadow'
+]);
+{% endhighlight %}
+
+Within your tests, you can access Shadow DOM elements using Polymer's DOM 
+API.
```
arthurevans: You use the Polymer.dom API to access local DOM children regardless of whether you're using shadow DOM or shady DOM, so this phrasing is a little confusing.

Since it's not the focus of the section, but a detail, I'd explain it after the test code sample ("`myEl.$$('button')` returns an element from `myEl`'s local DOM that matches the selector `'button'`" or WTTE).

After the `loadSuites` sample, you could explain what this does:

This sample runs the `basic-test.html` suite twice, once using shady DOM and once using native shadow DOM (if the browser supports it).

1.0/docs/devguide/tests.md
robdodson: Ehhhh it's tricky. If you've done any unit testing before the concept of spies and stubs should be familiar and stubs at least are explained later on..

TimvdLippe:
A great unit testing guide, really liked to read it. One thing though I would like to see added is testing using `polyserve`. It allows the developer to use features of the dev tools (break points, local css changes, etc...) and you can inspect the actual HTML elements on the page. Of course, for a full test suite run `wct` is superior, but when I am developing my own unit test, I prefer to run `polyserve`. My 2 cents.

robdodson:
I think (but could be mistaken) that both wct and polyserve should work
using the standard seed-element structure. We should test and if so maybe
we could include a paragraph about polyserve

1.0/docs/devguide/tests.md
kaycebasques: :heavy_check_mark:

1.0/docs/devguide/tests.md
kaycebasques: :heavy_check_mark:

1.0/docs/devguide/tests.md
kaycebasques: :heavy_check_mark:

1.0/docs/devguide/tests.md
kaycebasques: :heavy_check_mark:

1.0/docs/devguide/tests.md
kaycebasques: :heavy_check_mark:

1.0/docs/devguide/tests.md
kaycebasques: @ebidel or @robdodson can you help me with a code sample for this section? I'm not sure what makes for a good example here.

1.0/docs/devguide/tests.md
kaycebasques: :heavy_check_mark:

1.0/docs/devguide/tests.md
kaycebasques: :heavy_check_mark:

1.0/docs/devguide/tests.md
kaycebasques: :heavy_check_mark:

1.0/docs/devguide/tests.md
kaycebasques: :heavy_check_mark:

1.0/docs/devguide/tests.md
kaycebasques: :heavy_check_mark:

kaycebasques:
So @TimvdLippe if I understand correctly, you're recommending adding something about using `polyserve` to locally run an element, and then using Chrome DevTools to inspect the element. Is that correct?

TimvdLippe:
Yes, so the current workflow I use when working on a unit test is:

1. Run `polyserve` on root-element folder (or on a specific user test)
* Open Chrome and see the unit tests run
* Add/modify unit tests
* Refresh the page and immediately get response

Whenever a test fails, I can click on stacktraces and put breakpoints using the dev tools to debug the behavior. It is also possible to view the rendered CSS and modify rules accordingly.

For me these two advantages favor `polyserve` above `wct` for developing unit tests. When all unit tests succeed I run the whole suite with `wct`.

1.0/docs/devguide/tests.md
kaycebasques: :heavy_check_mark:

1.0/docs/devguide/tests.md
kaycebasques: :heavy_check_mark:

1.0/docs/devguide/tests.md
kaycebasques: :heavy_check_mark: updated the Chai description

:x: leaving the Sinon description as is

1.0/docs/devguide/tests.md
kaycebasques: :heavy_check_mark: "composed of" --> "built on top of popular third-party tools"
:heavy_check_mark: added clarification that the guide uses `wct` CLI to refer to WCT
:heavy_check_mark: pinged @azakus via Hangouts to make sure that the definition of WCT is OK

1.0/docs/devguide/tests.md
kaycebasques: :heavy_check_mark:

1.0/docs/devguide/tests.md
kaycebasques: :heavy_check_mark:

1.0/docs/devguide/tests.md
kaycebasques: :heavy_check_mark: added a paragraph of when / why to use stubs... the sample could be much improved to demonstrate why you would use this, but the new paragraph explains it pretty well

1.0/docs/devguide/tests.md
kaycebasques: :heavy_check_mark: nixed any references of confusing "index" term

1.0/docs/devguide/tests.md
kaycebasques: :heavy_check_mark: added a callout below the code sample

1.0/docs/devguide/tests.md
kaycebasques: :heavy_check_mark: replaced code sample with the `fireLasers()` example from `<seed-element>`... pew pew

1.0/docs/devguide/tests.md
kaycebasques: :heavy_check_mark: replaced code sample with `fireLasers()` test

kaycebasques:
damn you, @googlebot

kaycebasques:
thanks @TimvdLippe I'm going to open a separate issue to improve the doc. May take me some time. You're welcome to draft out a section, too.

robdodson:
Just started reviewing :) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

1.0/tools/tests.md
```diff
@@ -0,0 +1,385 @@
...
+    });
+  });
+</script>
+{% endhighlight %}
+
+The example above uses Chai's [`expect`](http://chaijs.com/api/bdd/) assertion
+style. 
+{: .alert .alert-info }
+
+## Run a subset of tests {#test-subsets}
```
robdodson: I tend to use this feature to run all of my tests, not just a subset. @arthurevans wdyt?

Technically to run a subset you should be able to use Mocha's `only` option:

`describe.only(function() {...}`

but I think this may be broken in wct.

1.0/tools/tests.md
```diff
@@ -0,0 +1,385 @@
...
+  </body>
+</html>
+{% endhighlight %}
+
+The argument to `loadSuites()` should be an array of strings. Each string
+should be a relative URL to a test suite. You can configure your tests 
+using query strings in the URLs. See [Test Shadow DOM](#shadow-dom)
+for an example.
+
+## Asynchronous tests
```
robdodson: I think this could be moved above "Preventing shared state with test fixture". Async tests are super common (and it uses assert, like the other tests in that section)

1.0/tools/tests.md
arthurevans: Good point. Should this section be titled something like "Run a set of test suites" instead?

In our elements, it seems like the most common use for this is to create an `index.html` file that lists all of the suites, so you can just run wct on test/index.html.

1.0/tools/tests.md
arthurevans: Makes sense to me.

Out of curiosity, Rob, do you ever use `dom-repeat.render()` or `Polymer.dom.flush()` in unit tests? 

We use them in the core library, in unit tests for `dom-repeat` and friends, but I'm not sure whether they're common (or helpful) elsewhere.

1.0/tools/tests.md
```diff
@@ -0,0 +1,385 @@
...
+
+To create an asynchronous test, pass `done` as an argument to the test function
+and then call `done()` when the test is complete. The `done` argument is a
+signal to Mocha that the test is asynchronous. When Mocha runs the tests it
```
arthurevans: tests => test?

1.0/tools/tests.md
robdodson: I always forget those exist :) We should probably mention them... (sorry @kaycebasques!)

1.0/tools/tests.md
```diff
@@ -0,0 +1,385 @@
...
+should be a relative URL to a test suite. You can configure your tests 
+using query strings in the URLs. See [Test Shadow DOM](#shadow-dom)
+for an example.
+
+## Asynchronous tests
+
+To create an asynchronous test, pass `done` as an argument to the test function
+and then call `done()` when the test is complete. The `done` argument is a
+signal to Mocha that the test is asynchronous. When Mocha runs the tests it 
+waits until it encounters the `done()` call, and eventually times out if it
```
arthurevans: "it does not encounter it" is a little funky. Also, we should identify what happens when the test times out.

Maybe: "it waits until the test code invokes the `done()` callback. If the `done()` callback isn't invoked, the test eventually times out and Mocha reports it as a failure."

We normally don't use parens after method names, but in this case I think they're worth including because the API is somewhat funky, and it might not be clear that the parameter you're passing (`done`) is a callback that's pre-defined by Mocha.

1.0/tools/tests.md
arthurevans: It appears that WCT supplies a `flush` method that takes a callback. It's basically calling `Polymer.dom.flush()` behind the scenes in most cases.  So I think we'd want to say something like:

Some tests rely on asynchronous DOM creation, especially if your element contains `dom-repeat` or `dom-if` templates. You can use the `flush` method ensure that asynchronous changes have taken place. 
For example, if you set the `items` property of a `dom-repeat` template, then want to inspect the generated DOM nodes, use `flush` to ensure the items have been rendered:

     test('creates one list item per item', function() {
       myEl.items  = [ 'One', 'Two', 'Three' ];
       // wait for items to render
       flush(function() {
         assert.equal(Polymer.dom(myEl.root).querySelectorAll('li').length, 3);
      });
     });

(Side note: I believe that nowadays using `Polymer.dom.flush()` to synchronously flush changes would work just as well as `flush()`, and be easier to read. Like:

        myEl.items  = [ 'One', 'Two', 'Three' ];
       // force items to render
       Polymer.dom.flush();
       assert.equal(Polymer.dom(myEl.root).querySelectorAll('li').length, 3);

However, this approach only works for Polymer... Specifically, I believe Polymer 1.2+, while the WCT `flush` method should work in a wider range of circumstances. Would like to refer that to @kevinpschaaf and @azakus for best practices.

This info could be grouped with either the async testing section or with the local DOM testing info. In either case it probably deserves a cross-reference. You could do:

- Test local DOM
    - Test asynchronous DOM creation (dom-repeat and dom-if).
    - Test with native shadow DOM.

1.0/tools/tests.md
```diff
@@ -0,0 +1,385 @@
...
+
+    test('fires lasers', function(done) {
+      myEl.addEventListener('seed-element-lasers', function(event) {
+        assert.equal(event.detail.sound, 'Pew pew!');
+        done();
+      });
+      myEl.fireLasers();
+    });
+
+## Test Shadow DOM {#shadow-dom}
```
arthurevans: "Test with native shadow DOM", perhaps?

1.0/tools/tests.md
```diff
@@ -0,0 +1,385 @@
...
+
+To test out how a test suite behaves when the browser runs native
+Shadow DOM, create a [test subset](#test-subsets) and pass `dom=shadow` as
```
arthurevans: Here (& below), lc "shadow" unless it's the first word of a title or sentence, or you're giving the name of the specification. This is the style we've adopted for the standards (so: shadow DOM, HTML imports, custom elements, not Shadow DOM, HTML Imports, Custom Elements).

1.0/tools/tests.md
kaycebasques: :heavy_check_mark: subset --> set

I don't know why the tests define `index.html` and then list all suites, as you could achieve the same effect by just calling `wct` with no args (since `wct` will execute all tests it finds in `test/`)...

1.0/tools/tests.md
kaycebasques: :heavy_check_mark:

1.0/tools/tests.md
kaycebasques: :heavy_check_mark:

1.0/tools/tests.md
kaycebasques: :heavy_check_mark:

1.0/tools/tests.md
kaycebasques: :heavy_check_mark:

1.0/tools/tests.md
kaycebasques: :heavy_check_mark:

1.0/tools/tests.md
```diff
@@ -0,0 +1,410 @@
...
+
+Use Polymer's [DOM API](/1.0/docs/devguide/local-dom.html#dom-api) to access 
+and modify local DOM children.
+
+    test('click sets isWaiting to true', function() {
+      myEl.$$('button').click();
+      assert(myEl.isWaiting, true);
+    });
+   
+`myEl.$$('button')` returns the first `button` element encountered
```
arthurevans: By convention, we start all notes with a bold run-in head. They don't have to be "Note:" -- lately I've taken to making them more descriptive.

1.0/tools/tests.md
```diff
@@ -0,0 +1,410 @@
...
+in the local DOM of `myEl`.
+{: .alert .alert-info }
+ 
+### Test asynchronous mutations
```
arthurevans: I think this should be "DOM mutations" just to avoid confusion.

1.0/tools/tests.md
```diff
@@ -0,0 +1,410 @@
...
+
+Always wrap your test in `flush` if your element template contains a [template
```
arthurevans: This is slightly overkill (not so much "always" as "if you change the DOM during the test")

1.0/tools/tests.md
kaycebasques: :heavy_check_mark:

1.0/tools/tests.md
kaycebasques: :heavy_check_mark:

1.0/tools/tests.md
kaycebasques: :heavy_check_mark: leaving as is

1.0/tools/tests.md
```diff
@@ -0,0 +1,418 @@
...
+Always wrap your test in `flush` if your element template contains a [template 
+repeater (`dom-repeat`)](/1.0/docs/devguide/templates.html#dom-repeat) or
+[conditional template (`dom-if`)](/1.0/docs/devguide/templates.html#dom-if),
+or if your test involves a local DOM mutation. Polymer lazily performs these 
+operations in some cases for performance. `flush` ensures that asynchronous 
+changes have taken place. The test function should take one argument, `done`,
+to indicate that it is [asynchronous](#async), and it should call
+`done()` at the end of `flush`.
+
+    test('Item lengths should be equalled', function(done) {
```
arthurevans: equalled => equal

1.0/elements/docs-menu.html
```diff
@@ -76,9 +76,16 @@
       <a href="https://polymer.github.io/polymer" target="_blank"></a>
     </core-item>
 
-    <core-item label="Tools" icon="editor:format-paint">
-      <a href="/{{versionPrefix}}/tools/overview.html"></a>
-    </core-item>
+<core-submenu label="Tools" icon="editor:format-paint">
```
robdodson: just noticed the indentation on this is off

1.0/elements/docs-menu.html
```diff
@@ -76,9 +76,16 @@
...
+    <core-submenu label="Tools" icon="editor:format-paint">
+      <core-item label="Overview">
+        <a href="/{{versionPrefix}}/tools/overview.html"></a>
+      </core-item>
+      <core-item label="Unit tests">
+        <a href="/{{versionPrefix}}/tools/tests.html"></a>
+      </core-item>
+    </core-submenu>
+
+     </core-item>
```
arthurevans: Looks like this closing tag is a remnant of the original "Tools" `core-item`

1.0/tools/tests.md
```diff
@@ -0,0 +1,421 @@
+---
+layout: default
+type: guide
+shortname: Docs
+title: Unit tests
+subtitle: Developer guide
```
arthurevans: Developer guide => Test your code with Web Component Tester.

1.0/tools/tests.md
```diff
@@ -0,0 +1,421 @@
...
+    test('click sets isWaiting to true', function() {
+      myEl.$$('button').click();
+      assert(myEl.isWaiting, true);
+    });
+   
+**Note:** `myEl.$$('button')` returns the first `button` element encountered
+in the local DOM of `myEl`.
+{: .alert .alert-info }
+ 
+### DOM mutations
```
arthurevans: Sorry, I wasn't very clear on my previous comment. I just meant to insert the word "DOM":
"Test asynchronous DOM mutations". You could leave out the "asynchronous" if you want, but for consistency, should keep the "Test." Sorry for the confusion.

1.0/bower.json
```diff
@@ -27,7 +27,8 @@
     "iron-icons": "PolymerElements/iron-icons#^1.0.0",
     "firebase-element": "GoogleWebComponents/firebase-element#^1.0.0",
     "google-map": "GoogleWebComponents/google-map#^1.0.5",
-    "iron-component-page": "PolymerElements/iron-component-page#^1.1.2"
+    "iron-component-page": "PolymerElements/iron-component-page#^1.1.2",
+    "google-youtube": "GoogleWebComponents/google-youtube#~1.1.2"
```
arthurevans: We don't need this line here. The component we use on the site is the 0.5 version.

1.0/elements/docs-menu.html
kaycebasques: :heavy_check_mark:

1.0/tools/tests.md
kaycebasques: :heavy_check_mark:

1.0/tools/tests.md
kaycebasques: :heavy_check_mark: I left out "asynchronous" so that people don't accidentally glaze over the section

1.0/bower.json
kaycebasques: :heavy_check_mark:

1.0/bower.json
```diff
@@ -27,7 +27,7 @@
     "iron-icons": "PolymerElements/iron-icons#^1.0.0",
     "firebase-element": "GoogleWebComponents/firebase-element#^1.0.0",
     "google-map": "GoogleWebComponents/google-map#^1.0.5",
-    "iron-component-page": "PolymerElements/iron-component-page#^1.1.2"
+    "iron-component-page": "PolymerElements/iron-component-page#^1.1.2",
```
arthurevans: Dangling comma.

kaycebasques: :heavy_check_mark: old habit from Python

kaycebasques:
rob showed me some rebase magic and we [merged](https://github.com/Polymer/docs/commit/e52f23849a1c0d59acbd1a301230403bd6feb5bd) from my local master

Issue #1476, opened 2015-12-21
error reading file .../docs/1.0/docs/elements/element-template.md?


kaycebasques:
`grunt doc` throws:

<img width="625" alt="screen shot 2015-12-21 at 2 43 38 pm" src="https://cloud.githubusercontent.com/assets/4713486/11943326/c7b7dc82-a7f1-11e5-955e-ff820b5aeabd.png">

I think the culprit may be `_plugins/element_pages.rb`

<img width="604" alt="screen shot 2015-12-21 at 2 47 44 pm" src="https://cloud.githubusercontent.com/assets/4713486/11943346/fdabc772-a7f1-11e5-80af-f7c6aaaef02a.png">

arthurevans:
Yeah, I'm getting this too, now, but I'm not sure why. It used to be that this only ran on the 0.5 directory.

robdodson:
Hm I just pulled latest and am not able to reproduce it :\

mercmobily:
I am getting this as well. I am using Ruby 2.1.5 -- could that be it?

samumist:
I'm getting this as well. Just pulled the latest. I'm using ruby 2.2.4.

robdodson:
I tried on 2.2.0 and wasn't able to reproduce but I can try again on 2.2.2

robdodson:
PR for potential fix is here, check out the instructions as you may need to delete the `_data/versions` directory

mercmobily:
#1480 worked for me...

robdodson:
yay !

mercmobily:
Hi,

Just to make sure... The errors are gone, and the site looks ok. But,
should I check something else in terms of what's been generated? Anything
to look out for, in case we chopped out an important piece of the site? :)

Merc.

robdodson:
The process is generating element docs for the old 0.5 elements. To verify it's still working you can click the dropdown in the top right to switch to the old 0.5 site. Then click on 'Elements' in the top nav. Then expand the elements drawer over on the side nav (great UI, i know...) then click on an element, like core-ajax.

If you see element docs that looks like this, it should all still be working:

<img width="1428" alt="screen shot 2015-12-28 at 5 43 51 pm" src="https://cloud.githubusercontent.com/assets/1066253/12028347/d40cf64a-ad8a-11e5-9f7c-16a4ba226bf7.png">

mercmobily:
Alright, confirming that it all works 100% then!
ᐧ

PR #1480, opened 2015-12-28
fix doc-merge task. Close Gh-1476

robdodson:
The doc merge task was previously being run for all polymer versions. It should now only run for 0.5. I still wasn't able to reproduce the error you were seeing but I *think* this may be the fix. You'll need to delete your `_data/versions` directory so it no longer has a `1_0` dir in it.

@arthurevans @kaycebasques PTAL

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -152,29 +152,8 @@ module.exports = function(grunt) {
     grunt.task.run('vulcanize');
   });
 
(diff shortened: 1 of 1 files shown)


ebidel:
LGTM

mercmobily:
Just confirming that deleting _data/versions and apply this PR fixed the problem for me.

robdodson:
Going to self merge as it seems like it's working for everyone

PR #1481, opened 2016-01-01
Add note about escaping in shells

eshansingh:
When I ran the install command in `zsh`, it reported `zsh: no matches found: Polymer/polymer#^1.2.0`. So I added a small note.

### DIFF
--- a/1.0/docs/start/getting-the-code.md
+++ b/1.0/docs/start/getting-the-code.md
@@ -55,6 +55,10 @@ The next step is to install {{site.project_title}}:
 
     bower install --save Polymer/polymer#^1.2.0
 
+Depending on your shell, `Polymer/polymer#^1.2.0` may need to be escaped appropriately.
+If you get an error somewhere around the lines of 'match not found', try surrounding 
+it with single or double quotes.
+
 Bower adds a `bower_components/` folder in the root of your project and
 fills it with {{site.project_title}} and its dependencies.
 


eshansingh:
Signed!

robdodson:
Actually I'm wondering why we're encoding a specific version number here? Wouldn't `bower install --save polymer/polymer` be a better approach?

eshansingh:
It wouldn't, because that would grab 1.2.3, which I believe is a dev version (not sure though!)

robdodson:
1.2.3 is just the latest release (we don't do dev versions). When you run `bower install --save` it writes whatever the latest tag is to your `bower.json`. The advantage is you'll always get the latest and greatest (which is important if we do a major version revision but forget to go back and update the article). Ultimately it's probably not a huge deal, but just something I noticed. If at some point we're on Polymer v1.5.7 it's going to feel weird to tell folks to peg their install to 1.2.3

arthurevans:
It used to just say #^1.0, to pin it to "latest version >= 1.0, < 2.0, which seemed appropriate.
But people keep updating it while I'm not looking :D

robdodson:
If you do `bower install --save polymer/polymer` it'll give you `~.1.2.3` which will protect from a major version rev. I think we should shorten it to that command

arthurevans:
That makes sense. I guess the `#^version` is a holdover from the 0.5 days. I'll merge this and massage the text a bit.

arthurevans:
Just realized that if we don't include the `#^1.2.0`, we don't need the note. Sorry. A little slow today.

arthurevans:
New PR here: https://github.com/Polymer/docs/pull/1488

PR #1488, opened 2016-01-07
Simplify install command, add note on versioning.

arthurevans:


arthurevans:
@robdodson @ebidel PTAL.

1.0/docs/start/getting-the-code.md
```diff
@@ -53,26 +53,30 @@ This generates a basic `bower.json` file. Some of the questions, like
...
-```
+The `--save` adds the item as a dependency in *your* app's bower.json:
+
+<pre>
 {
   "name": "my-project",
   "version": "0.0.0",
   "dependencies": {
-    "polymer": "Polymer/polymer#^1.2.0"
+    "polymer": "Polymer/polymer#~<var>currentVersion</var>"
```
robdodson: does this do magical stuff and insert the latest version?

robdodson:
LGTM if those variable tags do what they seem to do :D

1.0/docs/start/getting-the-code.md
arthurevans: No, it's just an HTML `<var>` tag, indicating replacement text. The paragraph below explains it. Rather than using template magic, I was just going to do this, to explain that the version will vary.

robdodson: Could you simplify it to:

`"polymer": "Polymer/polymer#~1.X.X"` maybe?

arthurevans:
OK, changed to 1.X.X and updated the comment. Better?

robdodson:
Does it still need the var tags and explanation? I think it's implied that 1.x.x means a version range.

ebidel:
Should we stick with ^ instead so people get minor updates.

robdodson:
@ebidel when you run `bower install --save Polymer/polymer` bower will write the `~` version to your bower.json. It's my understanding that since we have values for all 3 digits, the `^` and `~` behavior would be the same in this case... i think... :smile_cat:

ebidel:
Maybe it's more of an issue with element updates, but I've seen folks not get the minor bumps from dependencies.

robdodson:
Ah I'm wrong :P

> In the simplest terms, the tilde matches the most recent minor version (the middle number). ~1.2.3 will match all 1.2.x versions but will miss 1.3.0.

> The caret, on the other hand, is more relaxed. It will update you to the most recent major version (the first number). ^1.2.3 will match any 1.x.x release including 1.3.0, but will hold off on 2.0.0.

robdodson:
Hm gross check this out: https://github.com/bower/bower/issues/2121

> On Windows, ^ is a shell escape character. I think you need to either quote it, or double it up: bower install --save pkg#^^1.2.0.

robdodson:
Opened a bower issues to try to discuss this... https://github.com/bower/bower/issues/2144

robdodson:
PR to change the behavior is here: https://github.com/bower/bower/pull/2145

arthurevans:
So, should we change the install advice here?

My objection to the `#^1.2.0` version is that it suggests the 2nd digit is significant, and people keep wanting to update the docs, because they're "out of date".

Maybe we should change this back to `#^1.0` and explain that this means "latest stable release >= 1.0 and < 2.0"?

robdodson:
I think the advice as written is fine. I may even swap out the tilde for caret since that'll be in the next release of bower anyway.

arthurevans:
Updated now that the new bower is live, and added a note for folks using the older version. WTYT @robdodson @ebidel ?

arthurevans:
Silence == consent. Merging!

PR #1501, opened 2016-01-15
Remove legacy reference to python webserver

thierer:
The "creating reusable elements" section of the "Get started" docs contain a reference to the python web server which seems to be a legacy from the pre-polyserve days and is confusing because the example doesn't actually use python but polyserve...


arthurevans:
Good catch. LGTM. Thanks for the fix!

PR #1502, opened 2016-01-20
remove para about overriding callbacks, fixes #1455

### LINKED ISSUE #1455: is redefining `createdCallback` really the same as redefining `callback()`?
In the [docs](https://www.polymer-project.org/1.0/docs/devguide/registering-elements.html#lifecycle-callbacks), I read:

> You can fallback to using the low-level methods if you prefer (in other words, you can simply implement createdCallback in your prototype).

Are you sure that's the case? A Polymer's element would normally have, a its `attachedCalback()`
    
    function() {
      if (this._readied) {
        this._beforeAttached();
        baseAttachedCallback.call(this);
      } else {
        this._attachedPending = true;
      }
    }

If I rewrite this, I would lose the `_beforeAttached()` as well as the check on `this._readied`. Is it really safe to rewrite this function, as stated in the documentation? If so, what are the "practical" differences between the two? (I followed what happens in Polymer's `attachedCallback()` and a few things seem to be happening, especially waiting for the element to be styled...

kaycebasques:
PTAL @arthurevans

### DIFF
--- a/1.0/docs/devguide/registering-elements.md
+++ b/1.0/docs/devguide/registering-elements.md
@@ -185,9 +185,6 @@ in turn call shorter-named lifecycle methods on your prototype.
 - `detached` instead of `detachedCallback`
 - `attributeChanged` instead of `attributeChangedCallback`
 
-You can fallback to using the low-level methods if you prefer (in other
-words, you can simply implement `createdCallback` in your prototype).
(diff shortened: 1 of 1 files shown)


arthurevans:
LGTM. Thanks!

PR #1504, opened 2016-01-20
add sauce labs video to tests guide

kaycebasques:
@robdodson can you dummy check and make sure that the video loads correctly on your end?

I'd like to eventually expand this out to a section but it looks like a big undertaking.

### DIFF
--- a/1.0/tools/tests.md
+++ b/1.0/tools/tests.md
@@ -404,7 +404,26 @@ WCT.loadSuites([
 This sample runs `basic-test.html` twice, once using shady DOM and once
 using native shadow DOM (if the browser supports it).
 
+## Automated testing in the cloud
+
+It's important to get a good testing setup in place for your project as 
+early as possible. Using services like Travis for continuous integration, 
+and Sauce Labs for cross-browser testing means you can be confident that 
+changes you push to your project will work well on different platforms and 
+devices. For guidance on setting up these tools check out the Polycast below.
+
+<div class="yt-embed">
+  <google-youtube
+    videoid="afy_EEq_4Go"
+    autoplay="0"
+    rel="0"
+    fluid>
+  </google-youtube>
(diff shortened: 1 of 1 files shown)


robdodson:
sure lemme try

robdodson:
Just started reviewing :) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

robdodson:
I think this is a fine start. I agree that breaking it out into its own section is really tough. We could maybe tack a section on to the end of this page that just calls this out a bit more.

## Automated Testing in the Cloud

> It's important to get a good testing setup in place for your project as early as possible. Using services like Travis for continuous integration, and Sauce Labs for cross-browser testing means you can be confident that changes you push to your project will work well on different platforms and devices. For guidance on setting up these tools check out the Polycast below.

> // video embed

or something like that?

arthurevans:
@kaycebasques, could you add a section like Rob suggested? I really like mentioning Travis explicitly in the text--someone was asking about that today.

TimvdLippe:
@arthurevans yes that someone was me ;p
Yesterday I have spent quite some time setting up Travis and then I discovered https://github.com/Polymer/tools/tree/master/travis existed. So a small note like `A default Travis configuration can be copied from here: <link>` would be greatly appreciated!

robdodson:
this is mentioned in the video for what it's worth :D

TimvdLippe:
@robdodson Yes this is true, but a copy-pastable version is a lot easier to extract from, if you for example know most of the contents of the video and are only looking for the default travis configuration. Just a tiny link could save a lot of time for me at least.

robdodson:
oh yeah I agree, i'm just tooting my own horn :postal_horn:

kaycebasques:
I'll update tomorrow

PR #1479, opened 2015-12-27
Improved section on registering elements

mercmobily:
Hi,

And... here it is. I tried to make the language less "chatty" and did my best to explain exactly what goes on when elements are created.

I also improved it as much as possible. I really think it's a much better section now.

Thank you!

Merc.

PR #1506, opened 2016-01-21
Improvements to registration chapter

arthurevans:
Includes changes from #1479 (@mercmobily, the GoogleBot will probably ask you to approve this PR because it includes changes you made in another PR ... Don't ask me why. It just does.)

Changes can be previewed here:
https://init-order-dot-polymer-project.appspot.com/1.0/docs/devguide/registering-elements.html#lifecycle-callbacks

### DIFF
--- a/1.0/docs/devguide/registering-elements.md
+++ b/1.0/docs/devguide/registering-elements.md
@@ -177,16 +177,74 @@ all imports in the document have finished loading.
 ## Lifecycle callbacks {#lifecycle-callbacks}
 
 Polymer's Base prototype implements the standard Custom Element lifecycle
-callbacks to perform tasks necessary for Polymer's built-in features.  The hooks
-in turn call shorter-named lifecycle methods on your prototype.
-
-- `created` instead of `createdCallback`
-- `attached` instead of `attachedCallback`
-- `detached` instead of `detachedCallback`
-- `attributeChanged` instead of `attributeChangedCallback`
+callbacks to perform tasks necessary for Polymer's built-in features.
+{{site.project_title}} in turn calls shorter-named lifecycle methods on your
+prototype.
 
 Polymer adds an extra callback, `ready`, which is invoked when Polymer has
 finished creating and initializing the element's local DOM.
+
+<table>
+  <tr>
+    <th>Callback</th>
+    <th>Description</th>
+  </tr>
+  <tr>
+    <td><code>created</code></td>
+    <td>Called when the element has been created, but before property values are
+       set and local DOM is initialized.
+      <p>Use for one-time set-up before property values are set.
+      </p>
+      <p>Use instead of <code>createdCallback</code>.
+      </p>
+    </td>
+  </tr>
+  <tr>
+    <td><code>ready</code></td>
+    <td>Called after property values are set and local DOM is initialized.
+      <p>Use for one-time configuration of your component after local
+        DOM is initialized. (For configuration based on property values, it
+        may be preferable to use an <a href="properties.html#multi-property-observers">observer</a>.)
+      </p>
+    </td>
+  </tr>
+  <tr>
+    <td><code>attached</code></td>
+    <td>Called after the element is attached to the document.
+      <p>Uses include accessing computed style information, and adding
+        document-level event listeners. (If you use declarative
+        event handling, such as <a href="events.html#annotated-listeners">annotated
+        event listeners</a> or the
+        <a href="events.html#event-listeners"><code>listeners</code> object</a>,
+        {{site.project_title}} automatically adds listeners on attach and removes
+        them on detach.)
+      </p>
+      <p>Use instead of <code>attachedCallback</code>.
+      </p>
+    </td>
+  </tr>
+  <tr>
+    <td><code>detached</code></td>
+    <td>Called after the element is attached to the document.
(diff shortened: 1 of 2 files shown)


mercmobily:
Hi,

Looks fantastic. This big has CLA:NO but I don't have a way to actually sign it off...?

Merc.

mercmobily:
I got an email saying:

We found a Contributor License Agreement for you (the sender of this pull request) and all commit authors, but as best as we can tell these commits were authored by someone else. If that's the case, please add them to this pull request and have them confirm that they're okay with these commits being contributed to Google. If we're mistaken and you did author these commits, just reply here to confirm.

I guess all I have to say is "I AM TOTALLY OK"? :D

arthurevans:
Thanks Merc!

1.0/docs/devguide/registering-elements.md
```diff
@@ -177,17 +177,75 @@ all imports in the document have finished loading.
...
+        {{site.project_title}} automatically adds listeners on attach and removes
+        them on detach.)
+      </p>
+      <p>Use instead of <code>attachedCallback</code>.
+      </p>
+    </td>
+  </tr>
+  <tr>
+    <td><code>detached</code></td>
+    <td>Called after the element is attached to the document.
```
bicknellr: s/attached to/detached from/

bicknellr: It might be useful to reword these so it's more clear that attached and detached aren't one-off callbacks, but happen every time the element is attached / detached. I think the use of 'Called *after*' above in the explanation of ready kind of primed me to read it that way. 'Called *when*', like in attributeChanged, seems to imply that this can occur multiple times, but it also loses the important detail that it occurs after attachment / detachment. Maybe there's another way to state it.

1.0/docs/devguide/registering-elements.md
```diff
@@ -177,17 +177,75 @@ all imports in the document have finished loading.
...
+    <td><code>ready</code></td>
+    <td>Called after property values are set and local DOM is initialized.
+      <p>Use for one-time configuration of your component after local
+        DOM is initialized. (For configuration based on property values, it
+        may be preferable to use an <a href="properties.html#multi-property-observers">observer</a>.)
+      </p>
+    </td>
+  </tr>
+  <tr>
+    <td><code>attached</code></td>
```
bicknellr: For elements that are written explicitly in the template (/ initial state) of their container, I assume that attached follows ready, but this could be good to make explicit.

bicknellr: Oh, this is mentioned below in 'Ready callback and local DOM initialization'.

1.0/docs/devguide/registering-elements.md
```diff
@@ -232,56 +308,104 @@ local DOM when the element is constructed.
...
+-   Local DOM initialized (This means that **local DOM** children are created,
+    their property values are set as specified in the template, and `ready`
+    has been called on them).
+-   `ready` callback.
+-   [`factoryImpl` callback](#custom-constructor).
+-   `attached` callback.
 
+Note that while the life cycle callbacks listed above will be called in the
+described order for any given element, the  **initialization timing between
+elements may vary** depending on many factors, including whether or not the
```
bicknellr: Can't you at least rely on the parent being initialized before the descendants in its template?

bicknellr: Once again, explained later (in 'Initialization timing for light DOM children').

1.0/docs/devguide/registering-elements.md
```diff
@@ -232,56 +308,104 @@ local DOM when the element is constructed.
...
+    behavior and  polyfill behavior.
+
+#### Initialization timing for siblings
+
+There are no guarantees with regard to initialization timing between sibling
+elements.
+
+This means that siblings may become `ready` in any order.
+
+For accessing sibling elements when an element initializes, you can call `async`
```
bicknellr: Is there a way to listen for an element's attached callback, rather than deferring until the next tick?

1.0/docs/devguide/registering-elements.md
```diff
@@ -232,56 +308,104 @@ local DOM when the element is constructed.
...
+    adding and removing menu items with minimal overhead.
+
+*   To react when children are added and removed, use the
+    [`observeNodes` method](local-dom.html#observe-nodes).
+
+#### Initialization timing for local DOM children
+
+In terms of local DOM and initialization timing, local DOM children are created,
+their property values are set as specified in the template, and `ready` is
+called on them _before_ their parent's `ready` callback is called.
```
bicknellr: Haha, this is the opposite of what I expected, but makes more sense after thinking about it. I'm pretty sure I ended up relying on this behavior for a bug fix or two but still forgot.

1.0/docs/devguide/registering-elements.md
arthurevans: There is not. I would like to improve this section with better practices, but I don't have anything ready to go here. I

arthurevans:
Updated. Thanks for the comments, @bicknellr! Also updated the information about the registration callbacks based on email convo. with @kevinpschaaf.

robdodson:
lgtm

PR #1507, opened 2016-01-22
Update styling.md

dschuyler:
With the > on the left of ::content, the > will be removed -- is that the intent?  I could easily be wrong here, but I'm wondering if the > should go to the right of the ::content.


robdodson:
yeah this looks like a typo on our part. I _think_ this change LGTM but @arthurevans would you mind double checking?

arthurevans:
LGTM2. Thanks!

PR #1509, opened 2016-01-22
Update styling.md

dschuyler:
Hi, I found another case were the > looks like it should be after ::content.
The prior change can be seen on line 90 of this file.


arthurevans:
Either one would make sense, but let's keep it consistent.

PR #1511, opened 2016-01-22
Update styling.md

dschuyler:
Avoiding /deep/ and ::shadow is mentioned elsewhere in the document, but it's helpful to mention it directly in the code.

### DIFF
--- a/1.0/docs/devguide/styling.md
+++ b/1.0/docs/devguide/styling.md
@@ -550,7 +550,8 @@ Example:
           box-sizing: border-box;
         }
 
-        /* Can use /deep/ and ::shadow combinators */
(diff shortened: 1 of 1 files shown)


robdodson:
It seems more like we just shouldn't be showing `/deep/` or `::shadow` anywhere. They're going to get pulled out of Chrome at some point

dschuyler:
@robdodson, cool with me.  Maybe do both (add comment and then remove /deep/)?  This comment could be an interim change until a larger change is made to the doc?

robdodson:
I feel like we should just axe it now.. @arthurevans wdyt?

arthurevans:
Thanks for the PR! Fixed this in #1525, so closing this one.

PR #1515, opened 2016-01-26
document notifySplices and fix render; fixes #1439

### LINKED ISSUE #1439: Correct documentation of dom-repeat.render(); add docs for notifySplices
_From @wuyuanyi135 on November 19, 2015 4:47_

I was trying to mutate a items array of a dom-repeat with javascript and see if the template will update after the change. 

This is my test code, and I host it on c9 on Nov 19, 2015. Hope [this](https://myweb-wuyuanyi.c9users.io/filetable) helps.
```html
<!DOCTYPE html>

<link rel="import" href="../polymer/polymer.html" />

<dom-module id="file-list">
    <template>
        Your files:
        <table>
            <template id="table_template" is="dom-repeat" items="{{fileItems}}">
                <tr>
                    <td>{{index}}</td>
                    <td>{{item.fileName}}</td>
                    <td>{{item.mimeType}}</td>
                </tr>
            </template>

        </table>
        <input type="button" value="Add one!" on-click="buttonClicked" />
    </template>

    <script>
        Polymer({
            is: "file-list",

            properties: {
                fileItems: {
                    value: [{
                        type: Array,
                        fileName: "filename 1",
                        mimeType: "mime 1"
                    }, {
                        fileName: "filename 2",
                        mimeType: "mime 2"
                    }],
                    notify: true
                }
            },

            buttonClicked: function() {
                //this.fileItems.push({fileName:"filename x",mimeType:"mimeType x"});
                //this.$.table_template.render();
                this.push("fileItems",{fileName:"filename x",mimeType:"mimeType x"});
            }
        });
    </script>
</dom-module>
```

The [documentation page](https://www.polymer-project.org/1.0/docs/devguide/templates.html#dom-repeat) indicates that I need to use polymer function `this.push("fileItems",object)` rather than `fileItems.push(object)`, otherwise I need manually call `this.$.table_template.render()` to update the template.

What happened is, when I append an object to `fileItems` array, the `render()` throws an exception
```
Uncaught TypeError: Cannot read property '0' of undefined
Polymer.Collection._parseKey @ polymer.html:3900
Polymer.Collection.getItem @ polymer.html:3919
Polymer._stampInstance @ polymer.html:4328
Polymer._insertInstance @ polymer.html:4338
Polymer._applyFullRefresh @ polymer.html:4201
Polymer._render @ polymer.html:4134
Debouncer.complete @ polymer.html:1231
Polymer.Base.extend.flush @ polymer-mini.html:1152
Polymer.Templatizer._flushTemplates @ polymer.html:3634
Polymer.render @ polymer.html:4129
```

I roughly traced the problem, and find that in function `getKey` in line 3883 of `polymer.html`, `key = this.omap.get(item);` was executed. But `this.omap` only contains 2 elements (initial value). Therefore, the new appended elements cause `key` to be `undefined` and then lead to the exception. 

```javascript
getKey: function (item) {
    var key;
    if (item && typeof item == 'object') {
        key = this.omap.get(item);
    } else {
        key = this.pmap[item];
    }
    if (key != undefined) {
        return '#' + key;
    }
}
```

Did i use the proper way to invoke render()? How can I update `this.omap`?

_Copied from original issue: Polymer/polymer#3022_

kaycebasques:
@arthurevans WDYT?

### DIFF
--- a/1.0/docs/devguide/properties.md
+++ b/1.0/docs/devguide/properties.md
@@ -613,6 +613,17 @@ Example:
       </script>
     </dom-module>
 
+#### Using native array mutation methods {#notifysplices}
+
+Whenever possible you should always use Polymer's 
+[array mutation methods](#array-mutation). However, this isn't always
+possible. For example, you may be using a third-party library
+that does not use Polymer's array mutation methods. 
+In these scenarios you can call 
+[`notifySplices`](/{% polymer_version_dir %}/api/#Polymer.Base:method-notifySplices) 
+after each mutation to ensure that any Polymer elements observing the array 
+are properly notified of the changes.
+
 ## Property change notification events (notify) {#notify}
 
 When a property is set to `notify: true`, an event,
--- a/1.0/docs/devguide/templates.md
+++ b/1.0/docs/devguide/templates.md
@@ -66,14 +66,12 @@ instances, which update via the normal [structured data notification system
 ](#path-binding).
 
 Mutations to the `items` array itself (`push`, `pop`, `splice`, `shift`,
-`unshift`) must be performed using methods provided on Polymer elements, such
-that the changes are observable to any elements bound to the same array in the
-tree. For example:
-
-    this.push('employees', { first: 'Jack', last: 'Aubrey' });
-
-You can use the `render` method to force a `dom-repeat` to update (for example, 
-if you're using a third-party library that mutates the array).
+`unshift`), should be performed using Polymer's
+[array mutation methods](properties.html#array-mutation).
+These methods ensure that any elements observing the array are kept in sync.
+If you can't avoid using the native `Array.prototype` methods, make sure to
+call [`notifySplices`](properties.html#notifysplices) to ensure that any
+elements watching `items` are properly updated.
 
 ### Handling events in `dom-repeat` templates {#handling-events}
 
@@ -147,8 +145,16 @@ To filter or sort the _displayed_ items in your list, specify a `filter` or
 In both cases, the value can be either a function object, or a string identifying a 
 function defined on the host element.
 
-By default, the `filter` and `sort` functions only run when the array itself
-is mutated (for example, by adding or removing items).
+By default, the `filter` and `sort` functions only run when one of the
+following occurs: 
+
+*   The array itself is mutated (for example, by adding or removing items).
+*   The `filter` or `sort` function is changed.
+
+To re-run the `filter` or `sort` when an unrelated piece of data changes, 
+call [`render`](#synchronous-renders). For example, if your element has a 
+`sortOrder` property that changes how the `sort` function works, you can 
+call `render` when `sortOrder` changes.
 
 To re-run the `filter` or `sort` functions when certain sub-fields
 of `items` change, set the `observe` property to a space-separated list of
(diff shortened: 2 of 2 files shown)


1.0/docs/devguide/templates.md
```diff
@@ -277,6 +278,11 @@ different name for the index property.
     </template>
     {% endraw %}
 
+### Forcing synchronous renders {#synchronous-renders}
+
+Call [`render`](/{{versionPrefix}}/api/#dom-repeat:method-render) to force
+a `dom-repeat` template to synchronously render any changes to its data. 
+Normally changes are batched and rendered asynchronously.
```
arthurevans: We should add the two use cases here:

https://github.com/Polymer/docs/issues/1439#issuecomment-158156925

Hmm... the first one isn't really a use case, it's more of a general description. I think you could replace it with these two:

- For unit testing, to ensure items have rendered before checking the generated DOM.
- To ensure a list of items have rendered before scrolling to a specific item.

And explicitly note that it does _not_ cause the dom-repeat to pick up changes to the array that were made without the Polymer array mutation methods. If you (or a third-party library) has mutated the array without using the Polymer array mutation methods, you need to use notifySplices (link to section). 

This is one of the #1 things that confuses people about using render.

1.0/docs/devguide/properties.md
```diff
@@ -613,6 +613,18 @@ Example:
       </script>
     </dom-module>
 
+#### Using native array mutation methods {#notifysplices}
+
+Whenever possible you should always use Polymer's 
+[array mutation methods](#array-mutation).
+However, in some scenarios you might be forced to mutate arrays using the native
+`Array.prototype` methods. For example, you may be using a third-party library
```
arthurevans: This is a little wordy. What do you think of just:

"However, this isn't always possible." for the 2nd sentence here?

And maybe "that is not aware of" => "that does not use"?

1.0/docs/devguide/templates.md
```diff
@@ -150,6 +148,9 @@ function defined on the host element.
 By default, the `filter` and `sort` functions only run when the array itself
 is mutated (for example, by adding or removing items).
 
+If you change the `filter` or `sort` function and want to re-process the data,
```
arthurevans: These lines are incorrect. If you _change_ the `filter` or `sort` function, they'll be re-run automatically. You can use `render` to re-run the `filter` or `sort` function when some unrelated piece of data changes (for example, if your element has a `sortOrder` property that changes how the `sort` function works, you can call `render` when `sortOrder` changes.

1.0/docs/devguide/templates.md
```diff
@@ -150,6 +148,9 @@ function defined on the host element.
 By default, the `filter` and `sort` functions only run when the array itself
 is mutated (for example, by adding or removing items).
```
arthurevans: How about:

"when ..." =>
"when one of the following occurs:

*   The array itself is mutated (for example, by adding or removing items).
*   The `filter` or `sort` function is changed.

arthurevans:
Done reviewing. Left a few comments.

1.0/docs/devguide/properties.md

1.0/docs/devguide/templates.md
kaycebasques: :heavy_check_mark:

1.0/docs/devguide/templates.md
kaycebasques: :heavy_check_mark:

kaycebasques:
OK @arthurevans WDYT?

1.0/docs/devguide/templates.md
```diff
@@ -277,6 +283,23 @@ different name for the index property.
...
+
+Call [`render`](/{% polymer_version_dir %}/api/#dom-repeat:method-render) 
+to force a `dom-repeat` template to synchronously render any changes to its 
+data. Normally changes are batched and rendered asynchronously. This is 
+useful in two scenarios:
+
+*   For unit testing, to ensure items have rendered before checking the 
+    generated DOM.
+*   To ensure a list of items have rendered before scrolling to a 
+    specific item.
```
arthurevans: *   To rerun the `sort` or `filter` functions when an a piece of data changes _outside_ the array (for example, sort order or filter criteria).

1.0/docs/devguide/templates.md
```diff
@@ -277,6 +283,23 @@ different name for the index property.
     </template>
     {% endraw %}
 
+### Forcing synchronous renders {#synchronous-renders}
+
+Call [`render`](/{% polymer_version_dir %}/api/#dom-repeat:method-render) 
+to force a `dom-repeat` template to synchronously render any changes to its 
+data. Normally changes are batched and rendered asynchronously. This is
```
arthurevans: Note that given the order of these sentences, "This is useful..." appears to refer to "Normally changes are batched..."

So probably, "This is useful in two scenarios:" => "Synchronous rendering has a performance cost, but can be useful in a few scenarios:"

arthurevans:
Caught a couple of other things on 2nd pass. Left comments, otherwise LG.

1.0/docs/devguide/templates.md
kaycebasques: :heavy_check_mark:

1.0/docs/devguide/templates.md
kaycebasques: :heavy_check_mark:

PR #1519, opened 2016-01-29
fix prop change event docs, fixes #1437

### LINKED ISSUE #1437: Property changed notification documentation
_From @tpluscode on September 20, 2015 14:46_

Hi. I've found in the [documentation](https://www.polymer-project.org/1.0/docs/devguide/properties.html#notify) that a `propertyName-changed` event should be fired for properties with `notify: true`. However it seems that a `property-name-changed` event is fired instead

_Copied from original issue: Polymer/polymer#2461_

kaycebasques:
it looks like the documentation error was already fixed (docs said that `propName-changed` event was fired when it should have said `prop-name-changed`) but I added an example for clarity.

### DIFF
--- a/1.0/docs/devguide/data-binding.md
+++ b/1.0/docs/devguide/data-binding.md
@@ -268,13 +268,13 @@ When you configure a declared property with the `notify` flag set to `true`,
     `this.firstName` will fire `first-name-changed`. Listeners will receive an event object 
     whose `e.detail.value` attribute has the changing property's new value.
 
-When using a {{site.project_title}} element with other elements or frameworks, you can 
-manually attach an <code>on-<var>property</var>-changed</code> listener to an element to
-be notified of property changes, and take the necessary actions based on the new value. 
+You can manually attach a <code><var>property</var>-changed</code> 
+listener to an element to [notify external elements, frameworks,
+or libraries](/1.0/docs/devguide/properties.html#notify-external) of property 
+changes.
 
-This is essentially what {{site.project_title}} does when you create a two-way data
-binding.
-
+This is essentially what {{site.project_title}} does when you create a 
+two-way data binding.
 
 ### Two-way binding to native elements {#two-way-native}
 
--- a/1.0/docs/devguide/properties.md
+++ b/1.0/docs/devguide/properties.md
@@ -615,14 +615,22 @@ Example:
 
 ## Property change notification events (notify) {#notify}
 
-When a property is set to `notify: true`, an event,
-<code><var>property-name</var>-changed</code>, is fired whenever the property
-value changes. These events are used by the two-way data binding system, and can
-also notify external scripts and frameworks to respond to changes in the element.
+When a property is set to `notify: true`, an event is fired whenever the 
+property value changes. The event name is:
 
-For more on property change notifications and data binding, see  [Property
-change notification and two-way binding](data-binding.html#property-notification).
+<code><var>property-name</var>-changed</code>
 
+Where <code><var>property-name</var></code> is the dash-case version of 
+the property name. For example, a change to `this.firstName` fires 
+`first-name-changed`. 
+
+These events are used by the two-way data binding system. External 
+scripts can also listen for events (such as `first-name-changed`) 
+directly using `addEventListener`.
+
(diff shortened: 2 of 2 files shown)


kaycebasques:
Actually, I missed [Addy's comment](https://github.com/Polymer/docs/issues/1437#issuecomment-157898186)... let me check if that needs to be documented and I'll ping you when this is ready.

kaycebasques:
OK @arthurevans, all set.

1.0/docs/devguide/data-binding.md
```diff
@@ -268,13 +268,13 @@ When you configure a declared property with the `notify` flag set to `true`,
     `this.firstName` will fire `first-name-changed`. Listeners will receive an event object 
     whose `e.detail.value` attribute has the changing property's new value.
 
-When using a {{site.project_title}} element with other elements or frameworks, you can 
-manually attach an <code>on-<var>property</var>-changed</code> listener to an element to
```
kaycebasques: note that this was a bug. should have been `property-changed`, not `on-property-changed`.

1.0/docs/devguide/properties.md
```diff
@@ -615,14 +615,44 @@ Example:
...
-value changes. These events are used by the two-way data binding system, and can
-also notify external scripts and frameworks to respond to changes in the element.
-
-For more on property change notifications and data binding, see  [Property
-change notification and two-way binding](data-binding.html#property-notification).
-
+When a property is set to `notify: true`, an event is fired whenever the 
+property value changes. The event syntax is:
+
+    property-name-changed
```
arthurevans: We should use `<code><var>property-name</var>-changed</code>` here (and `<code><var>property-name</var></code>` below) for consistency.

1.0/docs/devguide/properties.md
```diff
@@ -615,14 +615,44 @@ Example:
...
-When a property is set to `notify: true`, an event,
-<code><var>property-name</var>-changed</code>, is fired whenever the property
-value changes. These events are used by the two-way data binding system, and can
-also notify external scripts and frameworks to respond to changes in the element.
-
-For more on property change notifications and data binding, see  [Property
-change notification and two-way binding](data-binding.html#property-notification).
-
+When a property is set to `notify: true`, an event is fired whenever the 
+property value changes. The event syntax is:
```
arthurevans: event syntax => event name.

1.0/docs/devguide/properties.md

kaycebasques:
OK @arthurevans feedback incorporated

1.0/docs/devguide/properties.md
```diff
@@ -615,14 +615,45 @@ Example:
...
+
+These events are used by the two-way data binding system, and can also 
+[notify external scripts and frameworks](#notify-external) 
+to respond to changes in the element.
+
+For more on property change notifications and data binding, see 
+[Property change notification and two-way 
+binding](data-binding.html#property-notification).
+
+### Notify external scripts of property changes {#notify-external}
```
arthurevans: I'm not convinced this is necessary (or necessarily fits here). I'd like @robdodson and @ebidel to weigh in here. We do have some total beginners using Polymer, but I'm not sure adding this task here is helpful. Thoughts?

1.0/docs/devguide/properties.md
```diff
@@ -615,14 +615,45 @@ Example:
...
-change notification and two-way binding](data-binding.html#property-notification).
+<code><var>property-name</var>-changed</code>
 
+Where <code><var>property-name</var></code> is the dash-case version of 
+the property name. For example, a change to `this.firstName` fires 
+`first-name-changed`.
+
+These events are used by the two-way data binding system, and can also 
+[notify external scripts and frameworks](#notify-external) 
+to respond to changes in the element.
```
robdodson: I think you could maybe remove the example below and replace it with an extra sentence here.

"External scripts can also listen for the "first-name-changed" event directly using addEventListener."

kaycebasques:
I removed the unnecessary section and added a sentence per Rob's suggestion

arthurevans:
LGTM.

PR #1522, opened 2016-02-03
list observer dependencies as args, fixes #1492

### LINKED ISSUE #1492: Properties: make it clear observers should be used for multiple properties
From http://rvcode.com/javascript/2016/01/09/polymer-mind-the-properties-init-order.html.

The misconception is that there's an ordering to how property changed handlers are called. If you move properties around in `properties`, the order of callbacks can change. For observers that also need other properties, `observers` with those dependencies are the way to go. 

Not sure how, but maybe this needs to be called out better in the docs if people are running into the same issues early on.

kaycebasques:
whaddayouthink, @arthurevans?

### DIFF
--- a/1.0/docs/devguide/properties.md
+++ b/1.0/docs/devguide/properties.md
@@ -324,9 +324,11 @@ Example:
 
     });
 
-**Compatibility note:** The argument order for change handlers is currently the
-**opposite** of the order used in 0.5.
-{: .alert .alert-info }
+**Warning:** A single property observer shouldn't rely on any other properties,
+sub-properties, or paths because the observer can be called while these 
+dependencies are undefined. See [Always include dependencies
+as observer arguments](#dependencies) for details.
+{: .alert alert-warning }
 
 Property change observation is achieved in Polymer by installing setters on the
 custom element prototype for properties with registered interest (as opposed to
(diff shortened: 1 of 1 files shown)


1.0/docs/devguide/properties.md
```diff
@@ -324,14 +324,69 @@ Example:
 
     });
 
-**Compatibility note:** The argument order for change handlers is currently the
```
kaycebasques: seemed unnecessary to continue to have 0.5 warnings

1.0/docs/devguide/properties.md
```diff
@@ -324,14 +324,69 @@ Example:
...
-**opposite** of the order used in 0.5. 
-{: .alert .alert-info }
-
 Property change observation is achieved in Polymer by installing setters on the
 custom element prototype for properties with registered interest (as opposed to
 observation via `Object.observe` or dirty checking, for example).
 
+### Always include dependencies as observer arguments
```
arthurevans: This heading (and some of the text below) doesn't make sense for single-property observers, since you can't include any additional dependencies.
 
In the current organization, it feels like this should perhaps be two sections.  Maybe a short warning in one section cross-referencing a more general section.

arthurevans:
Sorry @kaycebasques, I thought I'd provided feedback on this way back when.

kaycebasques:
Hey @arthurevans I shuffled the section lower in the doc and added a warning + link in the single-prop observer section. WDYT?

kaycebasques:
I don't know why that unrelated commit (`mention on-tap for indexing`) is showing up in the commit history in this PR. Concerning...

1.0/docs/devguide/properties.md
```diff
@@ -324,9 +324,10 @@ Example:
 
     });
 
-**Compatibility note:** The argument order for change handlers is currently the
-**opposite** of the order used in 0.5.
-{: .alert .alert-info }
+**Warning:**A single property observer shouldn't rely on any properties,
```
arthurevans: Missing space (before A). 
"rely on any" => "rely on any other"

1.0/docs/devguide/properties.md
```diff
@@ -324,9 +324,10 @@ Example:
 
     });
 
-**Compatibility note:** The argument order for change handlers is currently the
-**opposite** of the order used in 0.5.
-{: .alert .alert-info }
+**Warning:**A single property observer shouldn't rely on any properties, 
+sub-properties, or paths because [the observer can be called while these 
+dependencies are undefined](#dependencies).
```
arthurevans: Please replace the blind link with an attributed one. Like "the observer can be called while these 
 dependencies are undefined. See [Always include dependencies as observer arguments](#dependencies)
for details."

General idea: that an inline, context-free link on "foo" should only go to the definitive definition of "foo"--like an API reference page or concepts section.

arthurevans:
Added a few notes, otherwise LGTM.

arthurevans:
Re: that commit... The related PR for the `on-tap` commit seems to have gone missing, too. It looks like maybe you were on that branch when you created this one? Or possibly you merged that commit into your local master and rebased on that...

kaycebasques:
¯\_(ツ)_/¯

Sounds like things that I would do...

Issue #1524, opened 2016-02-05
Clarify section on scopeSubtree


arthurevans:
It should be more clear that `scopeSubtree` is only for nodes that _don't_ have any local DOM scoping applied. Calling it on a tree with Polymer elements in it will break the local DOM scoping. See comment here, which includes text that could be used as a rough draft:

https://github.com/Polymer/polymer/issues/3378#issuecomment-180479650

arthurevans:
Closed in #1525.

Issue #1355, opened 2015-09-16
Can't style distributed content with custom properties


arthurevans:
This should be documented. See:

https://github.com/Polymer/polymer/issues/2188

arthurevans:
In general, custom properties limitations section needs some attention. Perhaps a bullet-list summary of the limitations up front, followed by individual headings with detailed explanations of the different issues.

arthurevans:
See also: https://github.com/Polymer/polymer/issues/2149#issuecomment-124731041

arthurevans:
Closed in #1525.

PR #1525, opened 2016-02-08
Styles updates: Fixes #1188, #1355, #1524.

### LINKED ISSUE #1188: Update docs to say ::shadow and /deep/ are deprecated in Polymer 1.0
Here: https://www.polymer-project.org/1.0/docs/devguide/styling.html#xscope-styling

Note, users are still trying to use these [features](https://github.com/Polymer/polymer/issues/1809). We support them but they should be considered deprecated.

arthurevans:
Also temporary fix for malfunctioning API doc links. They weren't opening because of the AJAX-ified links. We either need to add an exception for `/1.0/api/` links or add `{:target="_blank"}` after all of them. I did the latter here rather than make a JS change, because I think we should consider how best to handle these links. But for now, unbreaking stuff seemed like the thing to do.

@robdodson @kaycebasques PTAL.


robdodson:
I think this seems ok.

Btw, do you have a plugin in your text editor that shaves off trailing white space? These diffs are hard to go through because there are so many white space changes. Would it make sense to do a pass over the site and rip out all the trailing white space so we can clean up our diffs?

arthurevans:
Yeah, I think it's something that's turned on in Sublime Text. It probably does make sense to make a pass over the site so we can reduce these diffs... Many PRs we get from the outside seem to have trailing whitespace removed, as well.

Issue #1533, opened 2016-02-10
The "Dynamism limitations" and "Inheritance Limitations" sections in Styling Local Dom are hard to follow


mercmobily:
# Dynamism limitations

I don't like writing issues in the documentation without fixing them (you should know that by now). However, for this one, all I can provide is code that shows that what I *think* the section is trying to say is wrong.

Here is the code:

````
<link rel="import" href="/bower_components/polymer/polymer.html">

<dom-module id="my-toolbar">
  <template>
    <style>
      :host {
        padding: 4px;
        background-color: gray;
      }
      .title {
        color: var(--my-toolbar-title-color);
      }
    </style>
    <span class="title">{{title}}</span>
  </template>
  <script>
    Polymer({
      is: 'my-toolbar',
      properties: {
        title: String
      }
    });
  </script>
</dom-module>

<dom-module id="x-custom">
  <template>
    <style>
      :host {
        --my-toolbar-title-color: black;
      }
      .funky {
        --my-toolbar-title-color: red;
      }
    </style>
    <my-toolbar id="normal" title="My awesome app, normal"></my-toolbar>
    <my-toolbar id="mega" class="funky" title="My awesome app, super"></my-toolbar>

    <button on-tap="changeTheme">Change theme</button>
  </template>
  <script>
    Polymer({
      is: 'x-custom',
      changeTheme: function() {
        this.$.normal.className += " funky"
        this.updateStyles();
      }
    });
  </script>
</dom-module>
````
So, adding a class doesn't change the color on the page (try to run this code without `updateStyles()`...)

Or maybe I totally misunderstood this section which had the following problems:

* No code to run (!)
* It's unclear what the limitation really is
* It seems to contradict what's been said before

# Inheritance limitations

I don't understand the limitations... and to be honest this part simply confused me.
This sentence: 

> "Unlike normal CSS inheritance which flows from parent to child, custom properties in Polymer’s shim can only change when inherited by a custom element from rules that set properties in scope(s) above it, or in a :host rule for that scope"

I am sure it's formally correct, and once I understand things I might be able to follow it, but while this section actually has code, it's a totally different type of element which in the local DOM's style part it sets `--custom-color` and then a CSS rule within there also _uses_ `--custom-color` as an attribute (?!?). 

This part:

> Unlike normal CSS inheritance which flows from parent to child, custom properties in Polymer’s shim can only change when inherited by a custom element from rules that set properties in scope(s) above it, or in a :host rule for that scope. 

[...]

PR #1537, opened 2016-02-15
Styling guide worked on extensively -- first draft

mercmobily:
Following up from #1533 , I decided that asking questions and asking you guys to "rewrite a section" was less than ideal .So,I sat down over the week end (well, actually, into Monday morning) and worked on an extensive rewrite.

After wasting some of your time, hopefully you will find here something valuable to use or take apart.

In general:

 * I made the examples more uniform and more comprehensive, always using `<my-element>` and `<my-toolbar>`

* I restructured it a little, so that there is always a "normal" explanation and then a section on the shim's limitations.

* Whenever possible, I tried my best to explain things (like the dynamism and inheritance limitations) using code that matched what the reader had read previously in the chapter

* I explained inheritance limitations providing -- again -- an example similar to the previous ones, but adding a <span> to the mix AND (not sure here if it made sense) trying to explain, "give or take", what happens behind the scenes and emphasising that Polymer elements are treated differently.

As usual: no hard feelings if you scrap it all. It was a good exercise for me anyway.

### DIFF
--- a/1.0/docs/devguide/styling.md
+++ b/1.0/docs/devguide/styling.md
@@ -8,11 +8,27 @@ subtitle: Developer guide
 
 {% include toc.html %}
 
-
 Polymer uses [Shadow DOM styling
 rules](http://www.html5rocks.com/en/tutorials/webcomponents/shadowdom-201/) for
-providing scoped styling of the element's local DOM.  Scoped styles should be
-provided via `<style>` tags placed inside the element's local DOM `<template>`.
+providing scoped styling of the element's local DOM. This allows elements to
+create style rules that will only apply to the element's light DOM or local
+DOM, as well as advanced theming with CSS custom properties and CSS mixins.
+
+Scoped styles should be provided via `<style>` tags placed inside the element's
+local DOM `<template>`.
+
+To place styles outside of the element, or share styles between elements, you
+can create a [style module](#style-modules).
+
+**NOTE: I think a nice introduction, without code, works a lot better. I also added a brief explanation.**
+
+# Styling the local DOM
+
+**NOTE: I think it's nice to introduce the "super-basic" example of shadow dom styling, without ::content**
+
+Here is a basic example of an element that uses `<style>` to style its local
+DOM. Note that any other element of class `child-element` within the page will
+not be affected:
 
 
     <dom-module id="my-element">
@@ -24,92 +40,148 @@ provided via `<style>` tags placed inside the element's local DOM `<template>`.
             display: block;
             border: 1px solid red;
           }
-          #child-element {
+          /* In shady DOM this will become `.child-element.my-element {` */
+          .child-element {
             background: yellow;
           }
-          /* styling elements distributed to content (via ::content) requires */
-          /* selecting the parent of the <content> element for compatibility with */
-          /* shady DOM . This can be :host or a wrapper element. */
-          .content-wrapper ::content > .special {
-            background: orange;
-          }
+
         </style>
 
-        <div id="child-element">In local DOM!</div>
-        <div class="content-wrapper"><content></content></div>
+        <div class="child-element">In local DOM!</div>
+        <content></content>
 
       </template>
 
       <script>
-
           Polymer({
               is: 'my-element'
           });
-
       </script>
 
     </dom-module>
 
-To place styles outside of the element, or share styles between elements, you can create
-a [style module](#style-modules).
+In the page, you might have:
+
+    <my-element>
+        <!-- In shady DOM the class is "child-element style-scope my-element" -->
+        <div class="child-element">In light DOM!</div>
+    </my-element>
+
+And see that only `In local DOM` has a yellow background.
 
 **Note:**  Prior to Polymer 1.1, the recommendation was to place `<style>` tags
 inside the `<dom-module>` for an element (but _outside_ the `<template>`). This
  is still supported, but is no longer recommended.
 {: .alert .alert-info }
 
+**NOTE: I like introducing what shadow DOM styling SHOULD be like, and then explain what we should do with shady DOM, and why.**
 
-### Styling distributed children (::content)
+#### Shim limitations
 
-Under shady DOM, the `<content>` tag doesn't appear in the DOM tree. Styles are rewritten to remove the
-`::content` pseudo-element, **and any combinator immediately to the left of `::content`.**
-
-This implies:
-
-*   You must have a selector to the left of the `::content` pseudo-element.
-
-        :host ::content div
-
-    Becomes:
-
-        x-foo div
-
-    (Where `x-foo` is the name of the custom element.)
-
-*   To limit styles to elements inside the ::content tag, add a wrapper element around the
-    `<content>` element. This is especially important when using a child combinator (`>`) to
-    select top-level children.
-
-        <dom-module id="my-element">
-
-          <template>
-
-            <style>
-              .content-wrapper ::content > .special {
-                background: orange;
-              }
-            </style>
-
-            <div class="content-wrapper"><content></content></div>
-
-          </template>
-
-        </dom-module>
-
-    In this case, the rule:
-
-        .content-wrapper ::content > .special
-
-    Becomes:
-
-        .content-wrapper > .special
+There is no need to do any special workaround in this simple case, since shady
+DOM will automatically change the style as well as the element's class to make
+sure that styling in the local DOM doesn't leak out of `my-element` by adding a
+required class, `my-element`, to the created element (and to the selector).
 
 
+### Styling distributed children (`::content`)
+
+In shadow DOM, you can style an element's light DOM using the `::content` pseudo element. So, you can do this:
+
+    <dom-module id="my-element">
+
+      <template>
+
+        <style>
+          /* WARNING: This does NOT work in shady DOM */
+          ::content > .special {
+            background: orange;
+          }
+        </style>
+
+       <div class="special">This should not be affected</div>
+       <content></content>
+      </template>
+
+      <script>
+          Polymer({
+              is: 'my-element'
+          });
+      </script>
+
+    </dom-module>
+
+
+In shadow DOM, the `.special` class will only ever affect elements in the light
+DOM. So when you use your element:
+
+    <my-element>
+        <div class="special">This should be special!</div>
+    </my-element>
+
+Only `This should be special!` will have an orange background.
+
+#### Shim limitations
+
+Shady DOM is much more limited than shadow DOM.
+
+Under shady DOM,  the `<content>` tag doesn't appear in the DOM tree, and styles
+are rewritten to remove the `::content` pseudo-element **as well as and any
+combinator immediately to the left of `::content`.** (E.g. `> :: content` where `>`
+is the combinator before it).
+
+Using shady DOM, the example above sees the style `::content > .special {`
+is rewritten as `my-element > .special`. This means that the rule will not apply
+just to the light DOM (as it should), but to the local DOM too (which it
+shouldn't).
+
+**NOTE: The original documentation is very confusing here, as it says that you must have a selector before `::content` (for example `:host`), but in my experience that doesn't actually do anything since BOTH `:host ::content div` and `::content div` becomes `element div` (which is obviously not enough). I wonder if adding `element` was a recent-ish addition, and earlier versions of polymer had `::content div` turn into a dangerous, leaky `div`!**
+
+This means that if you want to target only and only the light DOM, you will need
+to wrap `<content></content>` in a `div`, and target that specific `div` in your
+rules:
+
+    <dom-module id="my-element">
+
+      <template>
+
+        <style>
+          .content-wrapper ::content > .special {
+            background: orange;
+          }
+        </style>
+
+        <div class=".special">I am not special!</div>
+        <div class="content-wrapper"><content></content></div>
+
+      </template>
+
+      <script>
+          Polymer({
+              is: 'my-element'
+          });
+      </script>
+
+    </dom-module>
+
+In this case, the rule:
+
+    .content-wrapper ::content > .special
+
+Becomes:
+
+    .content-wrapper.my-element  > .special
+
+**NOTE: I have a bit of a problem here as the chapter hasn't yet introduced**
+**custom properties, and yet it warns "Custom properties can't style distributed**
+**children. I am commenting this warning out for now**
+
+<!--
 **Custom properties can't style distributed children.** The {{site.project_title}}
 [custom properties]({#xscope-styling-details) shim doesn't support styling
 distributed children.
 {: .alert .alert-info }
-
+-->
 
 ## Cross-scope styling {#xscope-styling}
 
@@ -138,6 +210,10 @@ provided to address the theming problem was the `/deep/` combinator and `::shado
 pseudo-element, which allowed writing rules that pierce through the Shadow DOM
 encapsulation boundary. However, these proved problematic and have been deprecated.
 {: .alert .alert-info }
+
+The recommended current way of affecting the look of an element's local DOM is by using
+custom CSS properties and custom CSS mixins, which offer great flexibility without
+paying the price of giving up encapsulation.
 
 ### Custom CSS properties {#xscope-styling-details}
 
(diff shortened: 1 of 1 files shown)


arthurevans:
Heh. Ironically, I started to do some rewrites on the plane home. Let me take a look your changes and figure out what we should do.

1.0/docs/devguide/styling.md
```diff
@@ -8,11 +8,27 @@ subtitle: Developer guide
 
 {% include toc.html %}
 
-
 Polymer uses [Shadow DOM styling
 rules](http://www.html5rocks.com/en/tutorials/webcomponents/shadowdom-201/) for
-providing scoped styling of the element's local DOM.  Scoped styles should be
-provided via `<style>` tags placed inside the element's local DOM `<template>`.
+providing scoped styling of the element's local DOM. This allows elements to
+create style rules that will only apply to the element's light DOM or local
```
arthurevans: 2 points here:

-   "light DOM" is not correct. This should probably be "the element's local DOM and distributed children." 

    Explanation: there are only two, specific places where scoped rules can touch stuff outside the element's local DOM: you can style the host element itself, with `:host`, and you can style distributed children, with `::content`. The distinction between light DOM children and distributed children is subtle, but important. If you don't have a `<content>` tag, you can't style your light DOM children, for instance. If you use `<content select=`, you can't style anything that doesn't match one of your selectors, because it doesn't get distributed.

-   The CSS custom properties shouldn't be mentioned in the same sentence. They're not related to the shadow DOM spec, they're a completely different specification that happens to be useful for styling shadow DOM. (We also talk about "the shady DOM shim" and "the custom properties shim" as separate entities.)

1.0/docs/devguide/styling.md
```diff
@@ -8,11 +8,27 @@ subtitle: Developer guide
...
+DOM, as well as advanced theming with CSS custom properties and CSS mixins.
+
+Scoped styles should be provided via `<style>` tags placed inside the element's
+local DOM `<template>`.
+
+To place styles outside of the element, or share styles between elements, you
```
arthurevans: This line is not quite right. Style modules are for sharing styles, yes. The `custom-style` element is what you use for defining local DOM aware style rules outside of an element.

1.0/docs/devguide/styling.md
```diff
@@ -8,11 +8,27 @@ subtitle: Developer guide
...
+can create a [style module](#style-modules).
+
+**NOTE: I think a nice introduction, without code, works a lot better. I also added a brief explanation.**
```
arthurevans: Yeah, I agree. 

(Longer-term plan, BTW, is to separate conceptual docs ("How shadow DOM scoping works") from task-oriented docs ("Style a local DOM element"). But working on that ATM for data binding; styling is probably a ways off.

1.0/docs/devguide/styling.md
```diff
@@ -24,92 +40,148 @@ provided via `<style>` tags placed inside the element's local DOM `<template>`.
             display: block;
             border: 1px solid red;
           }
-          #child-element {
+          /* In shady DOM this will become `.child-element.my-element {` */
```
arthurevans: We haven't really talked about how shady DOM shims things yet, so it's a little confusing to say, "In shady DOM this becomes ..." This is an implementation detail of how the shady DOM styling works, so if we discuss it, it should be in the context of the shim implementation/shim limitations.

1.0/docs/devguide/styling.md
```diff
@@ -24,92 +40,148 @@ provided via `<style>` tags placed inside the element's local DOM `<template>`.
...
+        <div class="special">This should be special!</div>
+    </my-element>
 
-        x-foo div
+Only `This should be special!` will have an orange background.
 
-    (Where `x-foo` is the name of the custom element.)
+#### Shim limitations
 
-*   To limit styles to elements inside the ::content tag, add a wrapper element around the
```
arthurevans: This deleted paragraph concisely describes why we're using the workaround starting at L133.

1.0/docs/devguide/styling.md
```diff
@@ -24,92 +40,148 @@ provided via `<style>` tags placed inside the element's local DOM `<template>`.
...
-    `<content>` element. This is especially important when using a child combinator (`>`) to
-    select top-level children.
+Shady DOM is much more limited than shadow DOM.
 
-        <dom-module id="my-element">
+Under shady DOM,  the `<content>` tag doesn't appear in the DOM tree, and styles
```
arthurevans: It is really confusing to insert this paragraph here just before describing a different limitation, which _also_ involves the child combinator. (This paragraph talks about `:host > ::content foo`, as opposed to `:host ::content > foo`, which are different selectors and see different transforms.

Honestly I think this is a much rarer case and I would not put it first.

1.0/docs/devguide/styling.md
```diff
@@ -24,92 +40,148 @@ provided via `<style>` tags placed inside the element's local DOM `<template>`.
...
+are rewritten to remove the `::content` pseudo-element **as well as and any
+combinator immediately to the left of `::content`.** (E.g. `> :: content` where `>`
```
arthurevans: E.g. => for example (we avoid Latin).

1.0/docs/devguide/styling.md
```diff
@@ -24,92 +40,148 @@ provided via `<style>` tags placed inside the element's local DOM `<template>`.
...
+is rewritten as `my-element > .special`. This means that the rule will not apply
+just to the light DOM (as it should), but to the local DOM too (which it
+shouldn't).
 
-            <style>
-              .content-wrapper ::content > .special {
-                background: orange;
-              }
-            </style>
+**NOTE: The original documentation is very confusing here, as it says that you must have a selector before `::content` (for example `:host`), but in my experience that doesn't actually do anything since BOTH `:host ::content div` and `::content div` becomes `element div` (which is obviously not enough). I wonder if adding `element` was a recent-ish addition, and earlier versions of polymer had `::content div` turn into a dangerous, leaky `div`!**
```
arthurevans: Yeah, I think  the doc is confusing here. Several things got conflated. 

At one point, you just needed a selector there.

But if you're using a child combinator, or if you have anything else in the local DOM that could be affected by these styles, you should use the wrapper approach. 

You highlighted one of the possible issues, here's the other one:

Rewriting `::content > .special` as `my-element > special` is potentially wrong because `>` selects _direct children`. So consider this element template:

    <template>
      <div id="main">
        <content></content>
      </div>
    </template>

Used like this:

    <my-element>
      <div class="special">I'm special!</div>
      <div>I'm not.</div>
    </my-element>
      
If I use the rule `::content > .special`, and in shady DOM that gets transformed to `my-element > .special` ... And it no longer selects anything, because the _rendered_ tree looks like this:

    <my-element>
       <div id="main">
         <div class="special">I'm special!</div>
         <div>I'm not.</div>
       </div>
    </my-element>

So, short version: If you have a very simple local DOM template like this:

    <template>
      <content></content>
    </template>

You can get away with just `::content selector`.

If your element has other local DOM contents that could match `selector`, or if you want to select only the top-level distributed children (using the child combinator (`>`)), then you need to use a wrapper.

mercmobily: I think this should be (nicely) explained in the documentation. I mean, it looks like too much detail, but.. isn't that where the devil is? I will be happy to include this in the explanation, if you're OK with it.

1.0/docs/devguide/styling.md
mercmobily: I know... my point being "too concisely?". Anyway, let me know if you'd like to keep that line :D

1.0/docs/devguide/styling.md
mercmobily: Very true.

1.0/docs/devguide/styling.md
mercmobily: Yes, yes and yes.

1.0/docs/devguide/styling.md
mercmobily: OK

arthurevans:
Thanks for all this work, Merc.

Not going to finish reviewing today. Let me finish reviewing and think this whole thing over before you put too much more time into this. For the custom properties shim in particular, I think we can make things a lot clearer with some structural changes.

mercmobily:
OK no worries. I was hoping to create something "nearly finished" -- and obviously failed again! I hope I didn't just create extra work in the process of trying to help...

1.0/docs/devguide/styling.md
```diff
@@ -342,112 +421,195 @@ Example usage of `my-toolbar`:
...
-Below are current limitations of the shim. Improvements to performance and
-dynamism will continue to be explored.
+For the following example, adding/removing the `urgent` class on the
+`<my-toolbar>` element will have the desired effect, since the dynamism is
+related to *application* of a custom property.
 
-#### Dynamism limitations
+**NOTE: as a selector, I used `:host.urgent span.title` rather than
+**:host(.urgent) span.title mainly because the :host(.selector) syntax hasn't
+**been used at all in the guide before now. I hope it's OK and it's "formally"
```
arthurevans: `:host(.urgent)` is correct. `:host.urgent` won't work under shadow DOM (and probably shouldn't work under shady DOM).

1.0/docs/devguide/styling.md
```diff
@@ -342,112 +421,195 @@ Example usage of `my-toolbar`:
...
+You can directly modify a {{site.project_title}} element's custom property by
+setting key-value pairs in
+[`customStyle`](../../api/#Polymer.Base:property-customStyle) on the element.
+This is analogous to setting `style` on an element.
 
-You can  directly modify a {{site.project_title}} element's custom property by setting
-key-value pairs in [`customStyle`](../../api/#Polymer.Base:property-customStyle) on the
-element (analogous to setting `style`) and then calling `updateStyles`. Or you can pass
-a dictionary of property names and values as an argumen to `updateStyles`.
+Note that in shady DOM you then need to call `updateStyles()` in order for the
```
arthurevans: This is a limitation of the custom properties shim and is not related to shady DOM.

1.0/docs/devguide/styling.md
```diff
@@ -24,92 +40,148 @@ provided via `<style>` tags placed inside the element's local DOM `<template>`.
...
+    .content-wrapper ::content > .special
 
-        .content-wrapper > .special
+Becomes:
 
+    .content-wrapper.my-element  > .special
 
+**NOTE: I have a bit of a problem here as the chapter hasn't yet introduced**
+**custom properties, and yet it warns "Custom properties can't style distributed**
+**children. I am commenting this warning out for now**
```
arthurevans: This is true, but it's also true that people don't tend to read docs in order. This is a large enough limitation (and a big enough source of confusion) that I wanted to put it up front here.

mercmobily:
Shall I get started on this Arthur?

1.0/docs/devguide/styling.md
```diff
@@ -8,11 +8,27 @@ subtitle: Developer guide
...
+
+Scoped styles should be provided via `<style>` tags placed inside the element's
+local DOM `<template>`.
+
+To place styles outside of the element, or share styles between elements, you
+can create a [style module](#style-modules).
+
+**NOTE: I think a nice introduction, without code, works a lot better. I also added a brief explanation.**
+
+# Styling the local DOM
```
arthurevans: This should be an H2 (`##`).

1.0/docs/devguide/styling.md
```diff
@@ -24,92 +40,148 @@ provided via `<style>` tags placed inside the element's local DOM `<template>`.
...
+And see that only `In local DOM` has a yellow background.
 
 **Note:**  Prior to Polymer 1.1, the recommendation was to place `<style>` tags
 inside the `<dom-module>` for an element (but _outside_ the `<template>`). This
  is still supported, but is no longer recommended.
 {: .alert .alert-info }
 
+**NOTE: I like introducing what shadow DOM styling SHOULD be like, and then explain what we should do with shady DOM, and why.**
+
+#### Shim limitations
```
arthurevans: Like these, but how about renaming them to "Shady DOM limitations" since we have both shady DOM and the custom properties shims to discuss in this chapter.

arthurevans:
Hi Merc,

Sorry to be slow on this, but I was out of the country for meetings last week and got called up for jury duty this week, so I'm way, way behind.

I don't have the bandwidth to deal with the custom properties stuff right now. I think it needs more structural work. 

Could you break out the first half of the PR? I think we're closer on that, and it's a more manageable chunk to review.

1.0/docs/devguide/styling.md
```diff
@@ -24,92 +40,148 @@ provided via `<style>` tags placed inside the element's local DOM `<template>`.
...
+
+There is no need to do any special workaround in this simple case, since shady
+DOM will automatically change the style as well as the element's class to make
+sure that styling in the local DOM doesn't leak out of `my-element` by adding a
+required class, `my-element`, to the created element (and to the selector).
+
+
+### Styling distributed children (`::content`)
+
+In shadow DOM, you can style an element's light DOM using the `::content` pseudo element. So, you can do this:
```
arthurevans: again, light DOM => distributed children throughout. pseudo-element is hyphenated.

1.0/docs/devguide/styling.md
```diff
@@ -24,92 +40,148 @@ provided via `<style>` tags placed inside the element's local DOM `<template>`.
...
 inside the `<dom-module>` for an element (but _outside_ the `<template>`). This
  is still supported, but is no longer recommended.
 {: .alert .alert-info }
 
+**NOTE: I like introducing what shadow DOM styling SHOULD be like, and then explain what we should do with shady DOM, and why.**
+
+#### Shim limitations
+
+There is no need to do any special workaround in this simple case, since shady
+DOM will automatically change the style as well as the element's class to make
```
arthurevans: Run-on sentence. Maybe something like:

Shady DOM emulates shadow DOM style encapsulation by adding classes to elements inside the local DOM tree at runtime, and by rewriting, or _shimming_ the style rules. For example, shady DOM adds the classes `style-scope` and `my-element` to the local DOM tree for `<my-element>`. The selector `.child-element` is rewritten as `.my-element.child-element`.

Shady DOM only shims style rules that appear _inside_ an element's local DOM, or in a [`custom-style` element ](link forward). Styles defined at the document level inside a normal `<style>` tag are **not** shimmed and can leak into local DOM trees. To avoid this issue, use a `custom-style` element for document-level styling.

1.0/docs/devguide/styling.md
```diff
@@ -24,92 +40,148 @@ provided via `<style>` tags placed inside the element's local DOM `<template>`.
...
+          });
+      </script>
 
-This implies:
+    </dom-module>
 
-*   You must have a selector to the left of the `::content` pseudo-element.
 
-        :host ::content div
+In shadow DOM, the `.special` class will only ever affect elements in the light
```
arthurevans: `.special` class => the `::content .special` selector

We use present tense (almost) everywhere, so "will only ever affect elements in the light DOM" => only matches distributed children. If you want to emphasize this more, could be "matches distributed children and _only_ distributed children."

mercmobily:
I am back... I locked myself in a cave till I finished the server side of the story of my project.
Now... I am eager to get this finished. Arthur, what would you like me to do? Shall I try and apply _all_ of the request you made, and see where we are at?

(Hopefully, at that point the review on your side won't take as much energy...)

I am happy to pounce on it whenever you want me to.

mercmobily:
(I will fix the conflicts too, the file has had a few commits since I made this PR)

arthurevans:
Welcome back @mercmobily! I was afraid I'd scared you off entirely.

Unfortunately, I've recently entered my own cave over here. If you can do just the first half (overview portion) in the next couple of days I can review it. Otherwise, I'm not going to get a look at it until after IO and I can guarantee more merge conflicts at that time :).

mercmobily:
Hi,

Damn. I was away those couple of days... I am too late right? :(
I am happy to go through it and resolve any conflicts really...

I think it's best if we wait till after IO.

Merc.
ᐧ

arthurevans:
Hi @mercmobily ... the new site is finally on the master branch here, so we had to close this (more explanation below):

mercmobily:
Hi,

No worries, thank you Arthur!

merc.

PR #1540, opened 2016-02-16
Fix broken links to Selenium SafariDriver

tony19:
Step 1 in "Test your elements" Quick Start describes SafariDriver with a link to a `http://code.google.com` page that no longer exists. This patch updates the message with links to the corresponding GitHub wiki pages and provides a direct link to the SafariDriver extension from Selenium.

I've also requested a similar [update for wct](https://github.com/Polymer/web-component-tester/pull/289).

### DIFF
--- a/1.0/tools/tests.md
+++ b/1.0/tools/tests.md
@@ -59,9 +59,9 @@ Follow the steps below to get set up, or watch the Polycast:
 
         npm install -g web-component-tester
 
-    On Mac OS X you need to [manually install][selenium] the latest Safari 
-    extension for Selenium. See the 
-    [Web Component Tester Polycast][workaround-example] for 
+    On Mac OS X you need to [manually install][selenium] the latest SafariDriver
+    extension for Selenium [(`SafariDriver.safariextz`)][safaridriver]. See the
+    [Web Component Tester Polycast][workaround-example] for
     a demonstration.
 
 2.  `cd` to the base directory of the element.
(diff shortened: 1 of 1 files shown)


1.0/tools/tests.md
```diff
@@ -59,9 +59,9 @@ Follow the steps below to get set up, or watch the Polycast:
 
         npm install -g web-component-tester
 
-    On Mac OS X you need to [manually install][selenium] the latest Safari 
-    extension for Selenium. See the 
-    [Web Component Tester Polycast][workaround-example] for 
+    On Mac OS X you need to [manually install][selenium] the latest SafariDriver
```
robdodson: should this be "and the latest SafariDrive"

tony19: I think "SafariDriver" is the actual name of the Safari extension, and I was trying to be a little more explicit/helpful. Since I'm already indicating the filename (SafariDriver.safariextz) in the same sentence, I guess this change is overkill. I can restore this to "Safari" if you'd prefer.

arthurevans: @tony19 I think the change is good. The wording here is a little confusing, but that's our fault. 

I'm a little iffy about linking to the current version of the driver, because we're unlikely to notice when that version changes. However, without the direct link it's nearly impossible to find, because currently it's _only_ in the 2.48 folder, not in any of the four subsequent releases.

Thoughts on this @robdodson? Inclined to leave the link for now, and work on the wording here later.

robdodson:
sgtm!

arthurevans:
Thanks for the PR!

PR #1542, opened 2016-02-16
Fixes #1539: Update events.md to have messages variable be actually visible in event listeners example

### LINKED ISSUE #1539: The examples in events.html, gestures section, don't do anything with this.message
Hi,

Have a look at the code example here: https://www.polymer-project.org/1.0/docs/devguide/events.html#gestures

The example currently sets this.message, but never does anything with it (!). Either we use `console.log`, or a `message` property that gets bound to the template.

If you go for the second choice, here is the code I propose. If it looks good to you, I will create a PR. If I am missing something, please say so :dancer: 

````
<dom-module id="drag-me">

  <template>

    <style>
      #dragme {
        width: 50px;
        height: 50px;
        background: gray;
      }
    </style>

    <div id="dragme" on-track="handleTrack"></div>
    <p>{{message}}</p>

  </template>

  <script>

    Polymer({

      is: 'drag-me',
      properties: {
        message: String
      },
      
      handleTrack: function(e) {
        switch(e.detail.state) {
          case 'start':
            this.message = 'Tracking started!';
            break;
          case 'track':
            this.message = 'Tracking in progress... ' +
              e.detail.x + ', ' + e.detail.y;
            break;
          case 'end':
            this.message = 'Tracking ended!';
            break;
        }
      }

    });

  </script>

</dom-module>

lozandier:
Fixed a quirk issue associated with the static generator used for the docs that requires `{{}}` to be escaped with `{%raw%}` & `{%endraw%}`

### DIFF
--- a/1.0/docs/devguide/events.md
+++ b/1.0/docs/devguide/events.md
@@ -217,7 +217,7 @@ Example:
           }
         </style>
 
-        <div id="dragme" on-track="handleTrack">{{message}}</div>
+        <div id="dragme" on-track="handleTrack">{%raw%}{{message}}{%endraw%}</div>
 
       </template>
 
@@ -265,7 +265,7 @@ Example with `listeners`:
           }
         </style>
 
-        <div id="dragme">{{message}}</div>
+        <div id="dragme">{%raw%}{{message}}{%endraw%}</div>
 
       </template>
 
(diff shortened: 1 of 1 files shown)


arthurevans:
LGTM. Thanks!

PR #1546, opened 2016-02-17
Update events.md to have event retargeting code example

lozandier:
## PR Goals
- [x] Add an explicit example of event-re-targeting in Polymer to be more consistent in the document flow of `events.md` regarding Polymer-specific handling of behavior of this aspect of Shadow DOM polyfilled by Shady DOM
- [x] Write a forced re-targeting situation after discussing this with @sjmiles; likely with the `click`event to be simple.

## Side-effects
- A slightly longer events.md file

### DIFF
--- a/1.0/docs/devguide/events.md
+++ b/1.0/docs/devguide/events.md
@@ -323,3 +323,38 @@ normalized event has the following properties:
     (equivalent to `event.path` under shadow DOM).
 
 
+Example:
+    <dom-module id="event-retargeting">
+      <template>
+        <button id="myButton">Click Me</button>
+      </template>
+    </dom-module>
+
+    <script>
+
+      Polymer({
+
+          is: 'event-retargeting',
+
+          listeners: {
+            'myButton.click': 'handleTap',
+          },
+
+          handleTap(e) {
+            console.info(e.target.id + ' was tapped.');
+          }
+
+        });
+
+    </script>
+
+  </dom-module>
+
+  <script>
+    document.querySelector('event-retargeting').addEventListener('click', function(){
+      var normalizedEventObject = Polymer.dom(event);
+      console.info('rootTarget is:', normalizedEventObject.rootTarget); // logs #myButton
+      console.info('localTarget is:', normalizedEventObject.localTarget); // logs an instance of event-targeting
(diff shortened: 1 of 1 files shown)


arthurevans:
Hi @lozandier. Thanks for the PR--in this case, I'm not sure the sample demonstrates retargeting in a useful way. Because of the way you've specified the listener, you're listening directly to events on the button, which are therefore never retargeted.

We have another (very old) PR with a sample using localTarget, but it doesn't really fit with the code style we use on the site, and it doesn't explain what's going on:

https://github.com/Polymer/docs/pull/1320/files

In that example, the actual event target is probably going to be an element in the `<paper-button>`'s local DOM, but what the user wants is the paper button itself, hence `localTarget`. Except, as it turns out, that's not a great example because the rootTarget returns the `<paper-button>`, too.

lozandier:
Hi, @arthurevans:

You're correct; my PR is incomplete (as indicated by the 1/2 checkbox of its goals). After asking a clarifying question that @sjmiles thoroughly answered, I was going to find time tonight or tomorrow my intent of this PR to have useful retargeting example. 

To be simple, probably with the `click` event instead of `tap` since the `tap` event can be pretty ugly to have access to outside the definition of an element's local DOM via `Polymer.Base._listen`

lozandier:
@arthurevans I've updated the PR to be a complete attempt of meeting its goals; feel free to review the example I came up with as a reasonable example. It seemed to have checked out fine with me when [making the Codepen for it](http://codepen.io/lozandier/pen/0c1396c5b2820991484e3eb231005fc1?editors=1010). 

<small>
Tip: You can import pens outright with Codepen using [[[urlOfPen]]]; if the Polymer pen entirely using the HTML tab to define the element you can technically use `import` as usual with the `.html` suffix applied. 
</small>

lozandier:
@arthurevans Whoops, at first accidentally replaced the tracking example with the updated re-targeting example

arthurevans:
LGTM. Only comments I have are formatting, so I'm going to go ahead and merge this and tweak it a bit. Thanks for the PR!

lozandier:
:+1:

PR #1550, opened 2016-02-19
mention on-tap for indexing

### LINKED ISSUE #1279: Ensure on-tap gets indexed somewhere
Currently, searching for "on-tap" doesn't yield any useful results in the 1.0 docs.

However, since we use it often in our code samples, people see it any try to find it in the docs. Not knowing that "on-<event>" is a general pattern and "tap" is a polymer gesture event, they expect to find documentation for a specific "on-tap" attribute.

Because it's a FAQ, we should add a note and make sure search sends people to the correct section.

kaycebasques:
fixes #1279

### DIFF
--- a/1.0/docs/devguide/events.md
+++ b/1.0/docs/devguide/events.md
@@ -53,13 +53,9 @@ Example:
 
     </dom-module>
 
-
-
-
-
 ## Annotated event listener setup {#annotated-listeners}
 
-To add event listeners to local-DOM children, use
+To add event listeners to local DOM children, use
 <code>on-<var>event</var></code>  annotations in your template. This often
 eliminates the need to give an element an `id` solely for  the purpose of
 binding an event listener.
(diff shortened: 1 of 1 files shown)


1.0/docs/devguide/events.md
```diff
@@ -165,8 +158,8 @@ interactions automatically when a declarative listener is added for the event
 type.  These events fire consistently on both touch and mouse environments,
 so we recommend using these events instead of their mouse- or
 touch-specific event counterparts. This provides better interoperability with both touch and
-mouse devices.  For example, `tap` should be used instead of
-`click` for the most reliable cross-platform results.
+mouse devices.  For example, `on-tap` should be used instead of
```
arthurevans: We should just use `tap` and `click` here, I think--those are the proper event names.

arthurevans:
one minor comment, otherwise LGTM.

PR #1553, opened 2016-02-23
Update Polymer version to latest stable release

iryston:
Update Polymer version to latest stable release v1.3.0


iryston:
I signed it!

arthurevans:
Thanks for the PR. There's actually no need to update this here, the pattern "#^1.2.0" matches the latest stable versions >= 1.2.0 and < 2.0.0.

The latest versions of bower actually do the right thing if you omit the version from the install command, so we'll be simplifying these shortly (see https://github.com/Polymer/docs/pull/1488).

PR #1554, opened 2016-02-23
Don't mention deprecated layout classes

rictic:
Remove mention of layout classes from migration guide, as they're deprecated,
and will eventually go away.


arthurevans:
Thanks for the PR! I addressed this instead by updating the docs to refer to the new style modules in #1571.

PR #1556, opened 2016-02-25
Update "Create a new page" examples to reflect current PSK.

michael-the1:
Match current PSK to tutorial examples.

While learning Polymer, I noticed some of the examples don't match the current PSK. They seem to contain good practices too (`[[one-way binding]]`, `{{baseUrl}}`) so I updated the examples to match.


kaycebasques:
note to self: need to add raw tags at line 104 so that Jekyll does not eat the double braces

kaycebasques:
looks good to me, thanks for keeping the doc fresh @michael-the1

michael-the1:
The pleasure was mine entirely.

PR #1562, opened 2016-02-26
add dom-repeat perf props, fixes #1503

### LINKED ISSUE #1503: template dom-repeat: mention delay, targetFramerate, etc
https://www.polymer-project.org/1.0/docs/devguide/templates.html doesn't mention the perf props `delay`, `intialCount`, `renderedItemCount`, `targetFramerate` etc.

http://polymer.github.io/polymer/

kaycebasques:
@addyosmani @ebidel is this good enough for #1503?

### DIFF
--- a/1.0/docs/devguide/templates.md
+++ b/1.0/docs/devguide/templates.md
@@ -300,8 +300,32 @@ rendering has a performance cost, but can be useful in a few scenarios:
 `render` **only** works with changes made with Polymer's
 [array mutation methods](properties.html#array-mutation).
 If you or a third-party library mutate the array without Polymer's methods,
-you need to call [`notifySplices`](#notifysplices) to ensure that any elements
+you need to call [`notifySplices`](properties.html#notifysplices) to ensure that any elements
 watching the array are properly notified.
+
+### Improve performance for large lists {#large-list-perf}
+
+By default, `dom-repeat` tries to render all of the list items at once. If 
+you try to use `dom-repeat` to render a very large list of items, the UI may 
+freeze while it's rendering the list. If you encounter this problem, enable 
+"chunked" rendering by setting 
+[`initialCount`](/{% polymer_version_dir %}/api/#dom-repeat:property-initialCount){:target="_blank"}. 
+In chunked mode, 
+`dom-repeat` renders `initialCount` items at first, then renders the rest of 
+the items incrementally one chunk per animation frame. This lets the UI thread 
+handle user input between chunks. You can keep track of how many items have 
+been rendered with the 
+[`renderedItemCount`](/{% polymer_version_dir %}/api/#dom-repeat:property-renderedItemCount){:target="_blank"} 
+read-only property.
+
+`dom-repeat` adjusts the number of items rendered in each chunk to try and 
+maintain a target framerate. You can further tune rendering by setting 
+[`targetFramerate`](/{% polymer_version_dir %}/api/#dom-repeat:property-targetFramerate){:target="_blank"}.
+
+You can also set a debounce time that must pass before a `filter` or `sort`
+function is re-run by setting the 
+[`delay`](/{% polymer_version_dir %}/api/#dom-repeat:property-delay){:target="_blank"}
+property.
 
 ## Array selector (array-selector) {#array-selector}
 


1.0/docs/devguide/templates.md
```diff
@@ -303,6 +303,30 @@ If you or a third-party library mutate the array without Polymer's methods,
...
+templates. 
+
+* [`delay`](/1.0/api/#dom-repeat:property-delay). Set a debounce time 
+  that must pass before a `filter` or `sort` function is re-run.
+* [`initialCount`](/1.0/api/#dom-repeat:property-initialCount). Define an 
+  initial number of template instances to render after setting the `items`
+  array, and put the `dom-repeat` template into chunking mode. The remaining
+  items are created and rendered incrementally at each animation frame until
+  all instances have been rendered.
+* [`renderedItemCount`](/1.0/api/#dom-repeat:property-renderedItemCount). The
```
arthurevans: This doesn't belong in this list: you can't set it and it doesn't affect performance. It's a read-only property that tells you how many items the dom-repeat has rendered so far.

1.0/docs/devguide/templates.md
```diff
@@ -303,6 +303,30 @@ If you or a third-party library mutate the array without Polymer's methods,
 you need to call [`notifySplices`](#notifysplices) to ensure that any elements
 watching the array are properly notified.
 
+### Performance properties {#dom-repeat-perf}
```
arthurevans: I think the task here would be "Improve performance for large lists" or something of that nature.

Main point: By default, `dom-repeat` tries to render all of the list items at once. If you try to use `dom-repeat` to render a very large list of items, the UI may freeze while it's rendering the list. If you encounter this problem, enable "chunked" rendering by setting `initialCount`. In chunked mode, `dom-repeat` renders `initialCount` items at first, then renders the rest of the items incrementally one chunk per animation frame. This lets the UI thread to handle user input between chunks. 

`dom-repeat` adjusts the number of items rendered in each chunk to try and maintain a target framerate.
You can further tune rendering by setting `targetFramerate` and `delay`.

1.0/docs/devguide/templates.md
```diff
@@ -303,6 +303,30 @@ If you or a third-party library mutate the array without Polymer's methods,
...
+
+Use the following properties to affect the performance of `dom-repeat` 
+templates. 
+
+* [`delay`](/1.0/api/#dom-repeat:property-delay). Set a debounce time
```
arthurevans: Make sure these links work correctly. I think we had to do something wacky before like include the full URL.

1.0/docs/devguide/templates.md
kaycebasques: @ebidel is there a reason you mentioned `renderedItemCount` in #1503?

1.0/docs/devguide/templates.md
kaycebasques: they're linking out correctly on my local build

1.0/docs/devguide/templates.md
```diff
@@ -303,6 +303,28 @@ If you or a third-party library mutate the array without Polymer's methods,
...
+### Improve performance for large lists {#large-list-perf}
+
+By default, `dom-repeat` tries to render all of the list items at once. If 
+you try to use `dom-repeat` to render a very large list of items, the UI may 
+freeze while it's rendering the list. If you encounter this problem, enable 
+"chunked" rendering by setting 
+[`initialCount`](/1.0/api/#dom-repeat:property-initialCount). In chunked mode, 
+`dom-repeat` renders `initialCount` items at first, then renders the rest of 
+the items incrementally one chunk per animation frame. This lets the UI thread 
+to handle user input between chunks. You can keep track of how many items have
```
arthurevans: delete "to"

1.0/docs/devguide/templates.md
```diff
@@ -303,6 +303,28 @@ If you or a third-party library mutate the array without Polymer's methods,
 you need to call [`notifySplices`](#notifysplices) to ensure that any elements
 watching the array are properly notified.
 
+### Improve performance for large lists {#large-list-perf}
+
+By default, `dom-repeat` tries to render all of the list items at once. If 
+you try to use `dom-repeat` to render a very large list of items, the UI may 
+freeze while it's rendering the list. If you encounter this problem, enable 
+"chunked" rendering by setting 
+[`initialCount`](/1.0/api/#dom-repeat:property-initialCount). In chunked mode,
```
arthurevans: These links aren't working correctly. Current workaround is to add a `target` attribute to the link (you can do that in markdown by adding `{:target="_blank"}` after the link, like the render link above (L288).

The problem here is that for normal page transitions, we usually try to inject the page contents into the current page rather than loading a whole new page ... But that doesn't work for the API docs, which are basically a separate app.

For consistency, "1.0" in the URL should probably be replaced with `{% polymer_version_dir %}` as in the previous link.

1.0/docs/devguide/templates.md
```diff
@@ -303,6 +303,28 @@ If you or a third-party library mutate the array without Polymer's methods,
 you need to call [`notifySplices`](#notifysplices) to ensure that any elements
 watching the array are properly notified.
 
+### Improve performance for large lists {#large-list-perf}
```
arthurevans: Extra credit: the #notifySplices link 2 lines up is broken, and should probably be another API docs link... :D

arthurevans:
Links need tweaking + one small typo. Otherwise, LGTM!

1.0/docs/devguide/templates.md
kaycebasques: Was supposed to link to a section in `properties.html`...

PR #1570, opened 2016-03-02
Fixes #1569. Use consistent target names.

### LINKED ISSUE #1569: API doc links should open in named tab
Currently each click on an API doc or catalog link opens a new tab. We should change these to `target="api"` or `target="catalog"` so we don't open a jillion tabs (plus, much less time spent loading websites).

arthurevans:


arthurevans:
I think this is straightforward but tagging @robdodson or @ebidel or @kaycebasques for sanity check plz.

robdodson:
Have you tested these in various different browsers? Just wondering if there are cross browser quirks to this approach

1.0/docs/devguide/styling.md
```diff
@@ -344,17 +344,17 @@ Example usage of `my-toolbar`:
...
 at element creation time.  In order to have an element (and its subtree) re-
 evaluate custom property values due to dynamic changes such as application of
-CSS classes, etc., call the [`updateStyles`](../../api/#Polymer.Base:method-updateStyles){:target="_blank"}
+CSS classes, etc., call the [`updateStyles`](/{% polymer_version_dir %}/api/#Polymer.Base:method-updateStyles){:target="api"}
 method on the element. To update all elements on the page, you can also call
 `Polymer.updateStyles`.
 
 You can  directly modify a {{site.project_title}} element's custom property by setting
-key-value pairs in [`customStyle`](../../api/#Polymer.Base:property-customStyle) on the
+key-value pairs in [`customStyle`](/{% polymer_version_dir %}/api/#Polymer.Base:property-customStyle) on the
```
kaycebasques: this one doesn't work

arthurevans: Oop. Yeah, missing the target attr...

kaycebasques:
@robdodson @arthurevans 

Firefox Developer Edition:

1. Click link to API doc. Tab opens as expected.
2. Click another link to different section in API. The API docs tab correctly jumps to the new place, but focus does not change to the API docs tab. It stays on pp.org. 

In Chrome and Safari, when you click a link from pp.org, the focus changes to the API docs tab in addition to jumping to new location (most desirable behavior IMO).

kaycebasques:
LGTM other than broken link and the difference in FF behavior

arthurevans:
OK. Going to go ahead and merge this. After it's deployed, we can run a quick poll for users on polymer-dev and the g+ page to see whether it's considered an improvement.

PR #1571, opened 2016-03-02
Update flex layout migration docs.

arthurevans:
Apologies for all the whitespace fixes.

Actual content changes are between L661 and L801.

@notwaldorf @kaycebasques PTAL

### DIFF
--- a/1.0/docs/migration.md
+++ b/1.0/docs/migration.md
@@ -8,15 +8,15 @@ subtitle: About this release
 
 {% include toc.html %}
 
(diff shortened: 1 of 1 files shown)


arthurevans:
OK, y'all are in another time zone and what we have there now is out of date, so I'm going to pull this in today and patch after the fact if necessary. Regardless of the fact that it's being merged, if you guys could take a quick gander over this when you come in and let me know if it needs any updates, I'd appreciate it.

PR #1586, opened 2016-03-18
Missing 'Function' property type

MeTaNoV:
adding missing `Function` property type


arthurevans:
`Function` isn't a useful type here. You _can_ put any value you like here, but Polymer only uses the listed ones for deserializing attributes.

MeTaNoV:
Thanks for the precision I didn't catch! :)

PR #1594, opened 2016-03-25
Add article on routing.

rictic:
Once this gets the LGTM we'll publish carbon-route.

(important note, we should publish carbon-route before this is merged so that bower install is all good.)

### DIFF
--- a/1.0/articles/routing.md
+++ b/1.0/articles/routing.md
@@ -1,0 +1,164 @@
+---
+layout: default
+type: article
+shortname: Articles
+title: "Encapsulated Routing with Elements"
+
+article: true
+description: An overview to doing distributed routing in an application, as well as an introduction to &lt;carbon-route> and &lt;carbon-location>.
+published: 2016-03-25
+author: rictic
+polymer_version: "1.4.0"
+
+tags:
+- routing
+- urls
+---
+
+<style>
+  iframe {
+    border: 1px solid #D7D7D7;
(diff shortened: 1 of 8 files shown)


1.0/articles/routing.md
```diff
@@ -0,0 +1,165 @@
...
+
+Note also, how little actual routing there is here. Almost all of the work is being done by other elements that are purpose built for their roles in the UI, and they're just data bound to &lt;carbon-route>. This is right and proper.
+
+You can see the core of the idea here, but we’ve jumped ahead a bit. I mean, where does the route in `route="{{route}}"` come from? What’s that `pattern` property doing, and how is it related to `data`?
+
+{%endraw%}
+
+## &lt;carbon-route>
+
+Let’s walk back a bit and consider the consider the routing problem, one piece at a time, beginning with just a &lt;carbon-route>. What does it do, and how does it work?
```
motss: `Let's walk back a bit and consider the consider the` should be changed to 
`Let's walk back a bit and consider the`

rictic: Good catch, thanks! Fixed

rictic:
Converted element references from &lt;element-name> to `<element-name>` for consistency with the rest of the site.

rictic:
Got the LGTM from Arthur over chat!

arthurevans:
Here's one over github: LGTM!
:ship: :it:

PR #1595, opened 2016-03-26
Add route="{{route}}" to initial snippet, as we reference it below.

rictic:


rictic:
Thanks to @aomarks for pointing this out!

arthurevans:
LGTM.

PR #1596, opened 2016-03-28
Fix Polymer.Collection example formatting

jordanbtucker:
var collection = Polymer.Collection.get(changeRecord.base); var index = changeRecord.base.indexOf(collection.getItem(key));

becomes

    var collection = Polymer.Collection.get(changeRecord.base);
    var index = changeRecord.base.indexOf(collection.getItem(key));


arthurevans:
LGTM. Thanks!

PR #1597, opened 2016-03-29
Get routing demos working cross-browser.

rictic:
There may still be a bug in carbon-route with IE 10 but this should support IE 11, Safari, and Firefox.


rictic:
Fixes part of https://github.com/PolymerElements/carbon-route/issues/62

arthurevans:
Tested on Safari, LGTM.

PR #1581, opened 2016-03-07
Fix button styles in footer.

ashi009:
- Align footer with content
- Remove deprecated button icons (Removed in https://github.com/Polymer/core-icons/commit/1e94a1175f7b97cb58ef54ecd5ca07c04053bc92)
- Move margin to anchor to make action area consitent with the visual appearance
- Add margin-bottom to paper-button on mobile device

PR #1598, opened 2016-03-30
Fix footer on doc pages.

ashi009:
This is a fix after #1581, which restores footer on doc pages to normal.


arthurevans:
Can you explain what issue this PR is solving? I don't know what to look for in testing it.

ashi009:
This fixes the footer on https://www.polymer-project.org/1.0/docs/devguide/feature-overview.html and all other doc pages.

robdodson:
I think it just gives more padding so the social buttons are better vertically centered

ashi009:
Indeed.  I've tested on different screen sizes, all looks good to me.

arthurevans:
Okey doke.

PR #1609, opened 2016-04-20
Add an example that shows an unintentional bug

rianby64:
As I was reading the documentation, I wanted to experiment the possibilities that offers the observer feature.
And I noticed that I can change the property's value that is being observed. This leads to an infinity loop.
This is for me an obvious anti-pattern. Maybe will be useful for other developers to see this warning?

### DIFF
--- a/1.0/docs/devguide/properties.md
+++ b/1.0/docs/devguide/properties.md
@@ -650,6 +650,22 @@ can be called while the other dependencies are still undefined. For example:
 Note that {{site.project_title}} doesn't guarantee that properties are
 initialized in any particular order.
 
+Also keep in mind that the observer should never change the value it's observing.
+The following code will lead to an infinity loop:
+
+    properties: {
+      firstName: {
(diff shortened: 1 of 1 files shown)


arthurevans:

rianby64:
Thanks for commenting. I've just one question. Does this PR worth? Should we show how to avoid such infinity loops?

arthurevans:
Yes, I think it's a good caution. Instead of "never" I'd be inclined to say, "Be very careful when changing the observed value to avoid an infinite loop."

Two examples of avoiding it would be:

*   Input validation (if it's not capitalized, capitalize it)—because it's conditional, it doesn't loop infinitely.
*   Explicitly suppressing the observer:

        nameChanged: function(newFirstName) {
          if (! this.updatingName) { 
              this.updatingName = true;
              this.firstName = 'Sr. ' + newFirstName;
              this.updatingName = false;
          }
        }

Issue #1610, opened 2016-04-24
undocumented: `fullbleed` requires `iron-positioning` style module


tony19:
The [docs for **Full Bleed**](https://elements.polymer-project.org/guides/flex-layout#full-bleed-lt-body-gt-) do not state that the `iron-positioning` style module is required, making it difficult to determine how to use it. I'd submit a pull request, but I can't find the source for that page. I was able to find one incorrect example in the [Migration Guide](https://github.com/Polymer/docs/commit/f73b1a88a28491f6c64304fd3c9a823356cf1d9e?diff=unified#diff-40fb74a27ef06c86d2f8816e0f9a0a6bR799), where `iron-flex-alignment` is specified instead of `iron-positioning`.

tony19:
@arthurevans Where can I find the source for the [layout guide](https://elements.polymer-project.org/guides/flex-layout#full-bleed-lt-body-gt-)?

arthurevans:
@tony19 you can find it at https://github.com/Polymer/polymer-element-catalog/blob/master/guides/flex-layout.md

Thanks!

tony19:
Closed as part of #1611. The layout guide will be fixed in https://github.com/Polymer/polymer-element-catalog/issues/273.

PR #1611, opened 2016-04-24
Fix fullbleed example in 1.0 migration guide #1610

tony19:
The [migration guide's example](https://github.com/Polymer/docs/commit/f73b1a88a28491f6c64304fd3c9a823356cf1d9e?diff=unified#diff-40fb74a27ef06c86d2f8816e0f9a0a6bR799) used the `fullbleed` class but did not include the `iron-positioning` style module, where the class is defined. The example included `iron-flex-alignment` possibly for the `center-center` class, which was already defined in `iron-flex` and thus redundant in this case.

This patch replaces `iron-flex-alignment` with `iron-positioning` in the example.

### DIFF
--- a/1.0/docs/migration.md
+++ b/1.0/docs/migration.md
@@ -796,7 +796,7 @@ Example of using the layout classes in the main page:
       <!-- import module -->
       <link rel="import" href="/bower_components/iron-flex-layout/iron-flex-layout-classes.html">
       <!-- include the necessary style modules -->
-      <style is="custom-style" include="iron-flex iron-flex-alignment"></style>
+      <style is="custom-style" include="iron-flex iron-positioning"></style>
     </head>
     <body class="fullbleed layout horizontal center-center">
      ...


1.0/docs/migration.md
```diff
@@ -796,7 +796,7 @@ Example of using the layout classes in the main page:
       <!-- import module -->
       <link rel="import" href="/bower_components/iron-flex-layout/iron-flex-layout-classes.html">
       <!-- include the necessary style modules -->
-      <style is="custom-style" include="iron-flex iron-flex-alignment"></style>
+      <style is="custom-style" include="iron-flex iron-positioning"></style>
```
arthurevans: I believe `center-center` is `iron-flex-alignment`, so it needs all three modules, `iron-flex iron-flex-alignment iron-flex-positioning`

tony19: `center-center` is actually defined in both `iron-flex` ([iron-flex-layout-classes.html:70](https://github.com/PolymerElements/iron-flex-layout/blob/e4e1058/iron-flex-layout-classes.html#L70),[77](https://github.com/PolymerElements/iron-flex-layout/blob/e4e1058/iron-flex-layout-classes.html#77)) and `iron-flex-alignment` ([iron-flex-layout-classes.html:152](https://github.com/PolymerElements/iron-flex-layout/blob/e4e1058/iron-flex-layout-classes.html#L152)).

arthurevans:
That is ... weird. But you're totally right. Thanks for pointing that out!

PR #1615, opened 2016-05-01
Better documentation for using ES6/ES2015 classes

lastmjs:
I had to search around for this solution, I think adding some documentation would be helpful.


lastmjs:
I signed it!

PR #1617, opened 2016-05-09
Minor doc update

TheIronDeveloper:
This commit addresses https://github.com/Polymer/polymer/issues/3613

**Larger explanation**: Chrome provides a global event during event
listeners, so the docs example would still work for Chrome... but
Firefox would end up breaking (since no global event is provided).


arthurevans:

TheIronDeveloper:
That is totally alright, I'll try to make an updated PR in the near future.  Thanks for the update! 👍

PR #1619, opened 2016-05-16
Rename carbon-route and friends to app-*

rictic:
Part of https://github.com/PolymerElements/app-route/issues/100


rictic:
Should we include a note in the article that it used to be called carbon-route?

arthurevans:
Yes, we should definitely include a note up front, especially since it was published with the old names. I would say,

`<carbon-route>` and `<carbon-location>` have been renamed to `<app-route>` and `<app-location>` with release 0.9.0. This article has been updated with the new names. Or words to that effect.

rictic:
Added a note about the renaming

arthurevans:
Already fixed in the new site.

PR #1634, opened 2016-06-02
Add an example how to avoid an infinity loop

rianby64:
As I was reading the documentation, I wanted to experiment the possibilities that offers the observer feature. And I noticed that I can change the property's value that is being observed. This leads to an infinity loop.

### DIFF
--- a/app/1.0/docs/devguide/properties.md
+++ b/app/1.0/docs/devguide/properties.md
@@ -732,6 +732,46 @@ nameChanged: function(newFirstName, oldFirstName) {
 Note that Polymer doesn't guarantee that properties are
 initialized in any particular order.
 
+Be very careful when changing the observed value to avoid an infinite loop.
+The following code will lead to an infinity loop:
+
+```
+    properties: {
+      firstName: {
+        type: String,
+        observer: 'nameChanged'
+      }
+    },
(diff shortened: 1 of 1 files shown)


rianby64:
This idea comes from #1609 
Thanks to @arthurevans

katejeffreys:
@arthurevans if you can provide a technical review, I will edit for style.

app/1.0/docs/devguide/properties.md
```diff
@@ -732,6 +732,45 @@ nameChanged: function(newFirstName, oldFirstName) {
 Note that Polymer doesn't guarantee that properties are
 initialized in any particular order.
 
+Be very careful when changing the observed value to avoid an infinite loop.
```
arthurevans: This addition should be extracted into a separate section, since it is not strictly related to this section ("Always include dependencies as observer arguments"). Could be "Modifying the observed value in an observer" or "Avoid infinite loops."

katejeffreys: Excellent, thanks. I will do that.

katejeffreys:
Closing this because the docs have been re-organized and the changes need to be made to a different file. I have made a new PR with the changes to the correct file.

rianby64:
Ok. Tell me if I can help in some way. Thanks.

PR #1641, opened 2016-06-02
Add a service worker

notwaldorf:
I thiiiink this should work? Watch it live: https://service-worker-dot-polymer-project.appspot.com

- precaching the elements, the css/js and the main page. Total precache size is about 869 kB for 26 resources
- everything else should be runtime caching.

screen shot 2017-03-15 at 2 40 13 pm

R: @ebidel, @jeffposnick as an fyi since he's a 💎 and helped me with most of this anyway.


gulpfile.js
```diff
@@ -82,6 +83,58 @@ function createReloadServer() {
...
 }
 
+function writeServiceWorkerFile() {
+  var path = require('path');
+  var rootDir = 'dist';
+
+  var config = {
+    cacheId: 'polymerproject',
+    staticFileGlobs: [
+      rootDir + '/1.0/index.html',
```
arthurevans: Since these two pages are assembled server-side, I think the should be in the `dynamicUrlToDependencies` object so the service worker cache will get updated if  the templates change. I think that would look like this:

```
dynamicUrlToDependencies: {
  '/1.0/': [ '/1.0/index.html', '/templates/head-meta.html', '/templates/site-nav.html' ],
 '/1.0/about':  [ '/1.0/about.html', '/templates/head-meta.html', '/templates/site-nav.html' ]
}
```

arthurevans:
In testing, offline seems to work well. I have a few overall questions/comments though:

1) We need a loading/loading failed indication I think there's a separate ticket to show loading progress, but we also need to detect a failed load. Offline exacerbates this problem since we'd expect only some of the pages to be cached.

2) Not a service worker question but related to caching: we don't currently have any kind of cache busting scheme, which I recall as being an issue on the old site because the browser was super duper enthused about holding on to HTML imports. (Also, cache times are set to the default 10 minutes which is probably overkill and forcing lots of extra traffic for non-SW browsers.) 

#2 is more a question for @ebidel who may remember the history of the old site better...

arthurevans:
BTW: staged for testing at https://service-worker-dot-polymer-project.appspot.com/

gulpfile.js
notwaldorf: I don't think this is needed. We are caching things from '/dist' (that's what rootdir points to), which is after the page has been assembled, so it doesn't really matter what it was assembled from. My understanding is that dynamicURlToDependencies is needed when you do client side templating, which we don't

arthurevans: OK I defer to @jeffposnick on this one. I thought it was for server-side templating, which we do do. My understanding was that sw-precache creates a hash for each each resource to determine when it changes.

jeffposnick: Yup, @arthurevans is right. Good catch! `dynamicUrlToDependencies` lets you keep track of which dependencies uniquely define a given URL, to ensure that if any of those underlying dependencies change, the cache entry for that URL will be updated.

For some setups, it could be easier to use globbing + a few constants to generate this mapping instead of trying to maintain it by hand. See, for example, https://github.com/GoogleChrome/ioweb2015/blob/master/gulp_scripts/service-worker.js#L24

notwaldorf: But since we regenerated this file on every build, if it changes doesn't the cache entry get update as well?

app/js/app.js
```diff
@@ -67,4 +67,11 @@ if (console) {
   console.log("%cWelcome to Polymer!\n%cweb components are the <bees-knees>",
               "font-size:1.5em;color:#4558c9;", "color:#d61a7f;font-size:1em;");
 }
+
+// Register service worker if supported.
+if ('serviceWorker' in navigator) {
+  console.log('Registering service worker');
```
ebidel: 🔫 it

gulpfile.js
```diff
@@ -15,6 +15,7 @@ let $ = require('gulp-load-plugins')();
 let matter = require('gulp-gray-matter');
 let styleMod = require('gulp-style-modules');
 let cssslam = require('css-slam');
+var swPrecache = require('sw-precache');
```
ebidel: `varToLet.then(isHappy => { ... });`

gulpfile.js
```diff
@@ -82,6 +83,58 @@ function createReloadServer() {
...
+  var rootDir = 'dist';
+
+  var config = {
+    cacheId: 'polymerproject',
+    staticFileGlobs: [
+      rootDir + '/1.0/index.html',
+      rootDir + '/1.0/about.html',
+      rootDir + '/elements/**',
+      rootDir + '/js/*.js',
+      rootDir + '/css/*.css',
```
ebidel: can you convert to template strings?

gulpfile.js
```diff
@@ -320,10 +375,15 @@ gulp.task('clean', 'Remove dist/ and other built files', function() {
...
+  writeServiceWorkerFile();
+});
+
+
 // Default task. Build the dest dir.
 gulp.task('default', 'Build site', ['clean', 'jshint'], function(done) {
   runSequence(
-    ['style', 'images', 'vulcanize', 'vulcanize-demos', 'js'],
-    'copy', 'md:docs', 'md:blog',
+    ['style', 'images', 'vulcanize', 'vulcanize-demos', 'js', ],
```
ebidel: Remove the trailing `,`. Creeps me out.

ebidel:
Reviewed

notwaldorf:
@ebidel fixeded!

gulpfile.js
notwaldorf: @jeffposnick ^^^

jeffposnick:
I'm going to take this discussion out of some hidden/outdated comment threads and into the main PR. Apologies if any of my previous comments confused matters, as I forgot some of the context myself. Re: how/whether to use `dynamicUrlToDependencies` for, there's basically two different approaches:

1. List (some or) all the possible navigation URLs as keys in `dynamicUrlToDependencies`, mapped to the individual templates that uniquely define the content used for each URL. In this approach, the browser would request every URL when the SW installs, and store a complete, server-rendered copy of the URL in the cache, and use that cached copy for subsequent navigations. Every time any of the partial templates is updated, all those cached pages that depend on a template would be automatically invalidated and refetched.
2. List (almost) nothing in `dynamicUrlToDependencies`, and instead only cache the "App Shell" HTML structure, and then rely on the `navigateFallback` option to route to the App Shell, and client-side templating to populate the shell via dynamic content. This separation between shell and content means, among other things, that you don't end up invalidating a bunch of cached content just because a single HTML element is tweaked in a template.

One is the approach we took with IOWA2015: https://github.com/GoogleChrome/ioweb2015/blob/master/gulp_scripts/service-worker.js#L24
Two is the approach we took with IOWA2016: https://github.com/GoogleChrome/ioweb2016/blob/master/gulp_scripts/service-worker.js#L22

There's no single "right" way of doing this, but given what I had previously discussed with @notwaldorf, I know that the goal was an App Shell + dynamic content approach, and that tends to be an approach that works well with sites that have lots of pages, only a subset of which might be visited.

Given that, the current PR looks like it has things properly implemented. I'd love to give this another once-over via looking in the Network tab and checking all the responses before this actually gets deployed to production, just to confirm that I haven't been giving bad advice.

gulpfile.js
arthurevans: @notwaldorf By "this file" do you mean the service worker file, or the `dist/1.0/index.html`?

My point was that if we update `templates/head-meta.html`, that won't change the contents of `dist/1.0/index.html`

notwaldorf: Why not? When `templates/head-meta.html` changes, on the next rebuild (because you need it to deploy) this will regenerate `dist/1.0/index.html`, because we regenerate everything on every build. And this new, regenerated file will be different.

notwaldorf:
@jeffposnick I still don't understand. We regenerate the `/dist` files on every build. If one of the templates changes, the resulting `/dist/1.0/index.html` changes as well. I made the SW look at the `/dist` file, not the source one, so the generated file gets pre-cached, and it should be updated. I don't see how `dynamicUrlToDependencies` would help in this case? Halp. 😭

jeffposnick:
Right, @notwaldorf! I don't think you'll need `dynamicUrlToDependencies` with the App Shell approach you're using.

> Given that, the current PR looks like it has things properly implemented.

gulpfile.js
arthurevans: What I've been trying to say is, it _won't_. The jinja templating is server side, not build-time. If you look at `dist/1.0/index.html`, you'll see:

```
```

The `head-meta` template gets interpolated on the server at request time by `server.py`:

https://github.com/Polymer/docs/blob/master/server.py#L59

arthurevans:
BTW @jeffposnick, you can check this branch out live at https://service-worker-dot-polymer-project.appspot.com/

gulpfile.js
jeffposnick: If that's the case, then a single entry in `dynamicUrlToDependencies` does seem required:

```js
dynamicUrlToDependencies: {
  'dist/1.0/': ['path/to/templates/head-meta.html', 'path/to/dist/1.0/index.html']
}
```

(This will take care of both requests for 'dist/1.0/' and 'dist/1.0/index.html' due to [`directoytIndex`](https://github.com/GoogleChrome/sw-precache#directoryindex-string)'s default behavior.

arthurevans: Oh, I had totally missed `directoryIndex`. That explains a bit that I was confused about.

`dist` is the build directory, so I think:

```
dynamicUrlToDependencies: {
  '1.0/': [ 'dist/1.0/index.html', 'dist/templates/head-meta.html', 'dist/templates/site-nav.html' ]
}
```

notwaldorf: But that's bad news bears -- site-nav is included in most of the other
pages on the site, that could be runtime cached. Don't we need to list them
too?

notwaldorf: (And head-nav is included in literally every page on the site)

jeffposnick: This gets into the two scenarios outlined at https://github.com/Polymer/docs/pull/1641#issuecomment-224081435

In scenario 2), the App Shell (`index.html`, for this site) is the only thing that needs to be kept up to date, and it then pulls in dynamic content at runtime and populates the shell with the body of the specific page that's actually requested. That implies that everything else on the page stays the same (header/side nav/etc.) and just the dynamic content gets swapped in/out depending on the URL.

If we're in some sort of hybrid of the two scenarios, then things can be messy. Should we all discuss this off thread to avoid the back and forth and get some certainty?

arthurevans: If you're runtime caching, I think the SW is just caching the contents of the stuff it pulls down over the network, which is no problem.

The problem with static files is that, at build time, sw-precache caclulates a hash from the contents of the static file in your local build, and bakes that into the service-worker.js code it generates. 

So it's actually caching something like:

```
index.html?sw-precache=<some-hash-string>
```

If it finds that file in its cache, it assumes it's up to date if <some-hash-string> matches the hash string it expects.

I think. I actually broke my brain a little bit on this.

app/1.0/shell.html
```diff
@@ -0,0 +1,24 @@
...
+-->
+<!DOCTYPE html>
+<html itemscope itemtype="http://schema.org/Organization" lang="en">
+<head>
+<title>{% block title %}{% endblock %} - Polymer Project</title>
+{% include 'templates/head-meta.html' %}
+</head>
+<body class="fullbleed" unresolved>
+  <pw-shell>
+    {% include 'templates/site-nav.html' %}
```
arthurevans: Since every page load replaces the contents of `<pw-shell>`, it shouldn't need to have any children in the app shell file. (site nav and devguide-shell).

gulpfile.js
```diff
@@ -91,37 +91,47 @@ function writeServiceWorkerFile() {
     cacheId: 'polymerproject',
     staticFileGlobs: [
       `${rootDir}/1.0/index.html`,
```
arthurevans: These three (index, shell, about) shouldn't be in the static file list, since they're assembled on the server.

gulpfile.js
```diff
@@ -91,37 +91,47 @@ function writeServiceWorkerFile() {
...
+      `${rootDir}/1.0/shell.html`,
       `${rootDir}/1.0/about.html`,
       `${rootDir}/elements/**`,
       `${rootDir}/js/*.js`,
       `${rootDir}/css/*.css`,
       `${rootDir}/bower_components/**/webcomponents-lite.min.js`,
     ],
-    navigateFallback: '/1.0/index.html',
+    dynamicUrlToDependencies: {
+      '/1.0/shell.html': [
```
arthurevans: I think this should be the actual URL the service worker will load over the network—`/1.0/shell`—rather than the file on disk (`/1.0/shell.html`). If we request `/1.0/shell.html` the server should just reply with a redirect. I'm having a lot of trouble figuring out what's going on from the network panel :/

Also, the dependencies only need to include the stuff that the shell actually includes—which is basically `shell.html` and `head-meta.html`.

arthurevans:
Updated https://service-worker-dot-polymer-project.appspot.com/ ... Looking at the network log, we're getting two requests for shell.html with different sw-precache params—I'm pretty sure that one's from the static file globs and one's from the dynamic URL object. Can also see we're getting redirects for these paths.

gulpfile.js
arthurevans: Maybe add the 404 page as a dynamicUrl (since we'll probably always want to have that available?) and the rest (index, about) to the runtime caching patterns?

notwaldorf:
@arthurevans @ebidel @jeffposnick A new day, a new commit. I've added the top level app-shell, added all its templates to it, used `fastest` instead of `networkFirst` for everything after all, and stole the toast code from IOWA (and added it here).

Because the `navigateFallback` is the empty shell and not `1.0/index.html`, this means that `http://localhost:8080/` does not redirect to `http://localhost:8080/1.0`, so you don't actually see any content. I think this is bad, maybe, but I'm not 100% sure I understand why using `1.0/index.html` as the `navigateFallback` was bad

gulpfile.js
```diff
@@ -91,37 +91,47 @@ function writeServiceWorkerFile() {
...
+        path.join(rootDir, 'templates', 'base-devguide.html'),
+        path.join(rootDir, 'templates', 'head-meta.html'),
+        path.join(rootDir, 'templates', 'site-nav.html'),
+        path.join(rootDir, 'templates', 'base-blog.html'),
+        path.join(rootDir, 'templates', 'page-md-styles.html'),
+      ]
+    },
+    navigateFallback: '/1.0/shell.html',
```
notwaldorf: Hmm, why? That file doesn't exist? :(

app/1.0/shell.html
notwaldorf: Sure. I wanted to add all the things in there, but I guess that makes sense

jeffposnick:
I'll take a look at a local server running the current code!

gulpfile.js
arthurevans: It exists in the dist/ folder, but if you change the path in `dynamicUrlToDependencies`, above (which eliminates an excess redirect), the SW will request and cache the file as `/1.0/shell`, so you should change it here, too.

arthurevans:
On `navigateFallback`, I think the drawback is that it's going to load that file first every time you navigate to the site from somewhere else: so it's going to parse everything in the `navigateFallback` page, then it should immediately run the code to AJAX-load in the actual content. So the more content there is in there, the longer it'll take to load.

gulpfile.js
```diff
@@ -82,6 +83,68 @@ function createReloadServer() {
...
+        cache: {
+          maxEntries: 10,
+          name: 'blog-cache'
+        }
+      }
+    }],
+    stripPrefix: rootDir + '/',
+    verbose: false  /* When debugging, you can enable this to true  */
+  };
+  swPrecache.write(path.join(rootDir, 'service-worker.js'), config);
```
jeffposnick: This should be prefixed with a `return` so that the Promise is propagated back to the `gulp.task()`.

gulpfile.js
```diff
@@ -82,6 +83,68 @@ function createReloadServer() {
...
+      `${rootDir}/1.0/shell.html`,
+      `${rootDir}/1.0/about.html`,
+      `${rootDir}/elements/**`,
+      `${rootDir}/js/*.js`,
+      `${rootDir}/css/*.css`,
+      `${rootDir}/bower_components/**/webcomponents-lite.min.js`,
+    ],
+    dynamicUrlToDependencies: {
+      '/1.0/shell.html': [
+        path.join(rootDir, 'templates', 'base-devguide.html'),
```
jeffposnick: I don't see `base-devguide.html` (and `base-blog.html` or `page-md-styles.html`) used to uniquely determine the output of the `shell.html` template, so I think they can be left out.

Apologies if I'm missing some way that they're being used by sub-templates or something like that.

gulpfile.js
```diff
@@ -82,6 +83,68 @@ function createReloadServer() {
...
+  let config = {
+    cacheId: 'polymerproject',
+    staticFileGlobs: [
+      `${rootDir}/1.0/index.html`,
+      `${rootDir}/1.0/shell.html`,
+      `${rootDir}/1.0/about.html`,
+      `${rootDir}/elements/**`,
+      `${rootDir}/js/*.js`,
+      `${rootDir}/css/*.css`,
+      `${rootDir}/bower_components/**/webcomponents-lite.min.js`,
```
jeffposnick: It looks like you need to pick up `bower_components/highlight/highlight.js`, too.

jeffposnick:
I've left a couple of comments, but I need some more time tomorrow to take a closer look at the App Shell behavior.

app/js/app.js
```diff
@@ -68,9 +68,54 @@ if (console) {
...
+  ).then(function(registration) {
+    registration.onupdatefound = function() {
+      // The updatefound event implies that registration.installing is set; see
+      // https://slightlyoff.github.io/ServiceWorker/spec/service_worker/index.html#service-worker-container-updatefound-event
+      const installingWorker = registration.installing;
+      installingWorker.onstatechange = function() {
+        switch (installingWorker.state) {
+          case 'installed':
+            if (!navigator.serviceWorker.controller) {
+              window.showToast('Caching complete! Future visits will work offline.');
```
arthurevans: I'm not sure whether we should do this? Since we have runtime caching, only pages the user has visited will be available offline, so this seems like a bit of an oversell. 

(If we have a toast, maybe "Pages you view are cached for offline use."? Not wedded to this wording, just trying to find a middle ground.)

app/js/app.js
```diff
@@ -68,9 +68,54 @@ if (console) {
...
+    console.error('Service worker registration failed:', e);
+  });
+}
+
+// Check to see if the service worker controlling the page at initial load
+// has become redundant, since this implies there's a new service worker with fresh content.
+if (navigator.serviceWorker && navigator.serviceWorker.controller) {
+  navigator.serviceWorker.controller.onstatechange = function(event) {
+    if (event.target.state === 'redundant') {
+      window.showToast('A new version of this app is available. Refresh the page.');
```
arthurevans: Rather than the app-y language, maybe just, "Site updated. Refresh this page to see the latest content." ? May overpromise since the current page may _have_ no new content. But still... WDYT?

gulpfile.js
notwaldorf: They're being used by subtemplates (for example any of the docs pages extend `base-devguide.html` and use `page-md-styles.html` for styling, any of the blog pages extend `base-blog.html`. If I don't put them here, and they change, then won't the cache not get busted correctly?

app/js/app.js
notwaldorf: I like that sentence!

app/js/app.js
notwaldorf: Yup! I like it. I stole these from IOWA obviously

jeffposnick:
A quick update: I'm poking around a bit with Chrome DevTools right now, and as far as I can tell, initial navigation requests that should be picked up via `sw-precache`'s App Shell fallback are instead being picked up by a runtime caching strategy. I'll report back with more findings and any suggestions.

jeffposnick:
...and I see what's going on. The bad news is that it'll require an update to `sw-precache` to change the behavior—basically pushing down the inlined `sw-toolbox` code to the bottom of the generated service worker.

I will work on that along with further tweaks to the `sw-precache` config that I can recommend.

gulpfile.js
arthurevans: As far as I understand it, those templates are used to generate the various doc pages that we cache using the runtime caching strategy. Theoretically, when the service worker is in charge, we load the shell, then replace the `innerHTML` of the `pw-shell` with the `innerHTML` from the doc page's `pw-shell`. 

When those files change, the _shell_ hasn't changed, just stuff that gets injected inside the `pw-shell`. AFAICT.

An yes, we probably could have designed this better for Service Worker :cry: 

On the plus side, the current site kind-of works with JavaScript turned off. So that's something.

jeffposnick:
I've got a change to `sw-precache`'s behavior at https://github.com/GoogleChrome/sw-precache/issues/128

I don't know how rushed we are to get this SW implementation finished—I think @notwaldorf said she was going to be out of the office soon?—but if we needed something sooner than the next `sw-precache` release, we can either change the `package.json` version to pull in a pre-release, or use a locally modified copy of the [`service-worker.tmpl`](https://github.com/GoogleChrome/sw-precache/blob/master/service-worker.tmpl) file.

arthurevans:
I think we should wait for the release. @notwaldorf is out next week and I'm out M-W as well. I think we're pretty close but I don't want to rush to get this in and then realize we've hamstrung ourselves for i18n or something else critical...

notwaldorf:
Yeah, we can totally wait on it. I'm already worried this is going to break everything, so I'd rather not risk it :)

ebidel:
Shall we revive this?

jeffposnick:
That sounds like a good idea. `sw-precache` 4.0.0 is out now, and that includes https://github.com/GoogleChrome/sw-precache/issues/128 which was needed.

ebidel:
@notwaldorf pong'd back to you.

FluorescentHallucinogen:
Any progress?

notwaldorf:
Argh, I'll take another look this week.

gulpfile.js
notwaldorf: Done! (see below)

notwaldorf:
@ebidel @jeffposnick I did some changes! I would looove if someone who isn't me could test this locally and see if it's looking good -- I'm noticing a bunch of weird things: main page images that are cached (and I see them displayed) actually show up red in the console as GET fails :(

I also can't seem to get a newer version of a file. I did this:
- load the main `index.html` (gets cached)
- go offline. view it offline.
- modify one of the headers in `index.html`, go back online
- i keep seeing the old, cached version, and not the live one.

jeffposnick:
Thanks for continuing with this, and sorry that the process hasn't been smoother, @notwaldorf!

I'm tied up with a few things until the end of the week, but I'll give some feedback then.

gulpfile.js
```diff
@@ -82,6 +83,66 @@ function createReloadServer() {
...
+    dynamicUrlToDependencies: {
+      '/1.0/shell.html': [
+        path.join(rootDir, 'templates', 'head-meta.html'),
+        path.join(rootDir, 'templates', 'site-nav.html'),
+      ]
+    },
+    navigateFallback: '/1.0/shell',
+    runtimeCaching: [
+    {
+      urlPattern: /\/images\/*/,
```
jeffposnick: This picks up images from Google Maps:

image

It looks like it would be a bit hairy to tweak the RegExp to avoid Google Maps images while leaving it flexible enough to pick up any images on this site. So, we can use the *other* way of specifying URL routes, which just uses a string instead of a RegExp, but has the advantage of only matching same-origin URLs by default:

`urlPattern: '/images/(.*)'`

(The `sw-toolbox` [docs](https://googlechrome.github.io/sw-toolbox/docs/master/tutorial-usage.html) explain the different URL routing options, but it's kind of a mess, to be honest. In this particular case, it comes in handy, though.)

gulpfile.js
```diff
@@ -82,6 +83,66 @@ function createReloadServer() {
...
+      `${rootDir}/1.0/shell`,
+      `${rootDir}/1.0/about.html`,
+      `${rootDir}/elements/**`,
+      `${rootDir}/js/*.js`,
+      `${rootDir}/css/*.css`,
+      `${rootDir}/bower_components/**/webcomponents-lite.min.js`,
+      `${rootDir}/bower_components/highlight/highlight.js.js`,
+    ],
+    dynamicUrlToDependencies: {
+      '/1.0/shell.html': [
```
jeffposnick: I think the property name should be `/1.0/shell`, not `/1.0/shell.html`. The property names are meant to be the actual URLs that the server knows about, which doesn't include `.html`.

jeffposnick:
@notwaldorf: I think things are looking good! 🎆  The issues mentioned in https://github.com/Polymer/docs/pull/1641#issuecomment-225253917 seem to be fixed by the move to `sw-precache` 4.0.0.

I left two smallish comments about the `sw-precache` config.

Regarding making direct changes to `index.html`, because it's one of the files that's precache by `sw-precache` during the build step, you need to re-run the `generate-service-worker` task in order to get the service worker to pick up those changes. After regenerating the service worker, the next time you go to `index.html` you'll see a toast indicating that something's been updated. Then, if you follow the toast's directions and reload, or navigate to the page again, you'll see the newest content.

One of the things to think about prior to rolling this out is making sure that there's a strategy in place to roll it back in case anything goes wrong 😄 . To that end, [this Stack Overflow answer](https://stackoverflow.com/questions/33986976/how-can-i-remove-a-buggy-service-worker-or-implement-a-kill-switch/38980776#38980776) covers some things to think about, and a strategy for dealing with reverting a "bad" service worker deployment.

gulpfile.js
notwaldorf: Oh sorry! I thought I fixed it in all the places, but I missed this one, ugh

jeffposnick:
I may have misinterpreted what you were saying about changing one of the headers in `index.html` and then trying to trigger the refreshed content.

This dates back to the discussion in this comment: https://github.com/Polymer/docs/pull/1641#issuecomment-224124136

My question is whether the partial templates are inlined when `index.html` (and other `.html` files) are written out to disk at `dist/1.0/`. That earlier comment implies that they are inlined, and changing the partial template would change what's written out to disk. But when I just ran the standard `gulp` build process, I noticed that `dist/1.0/index.html` file still contains a `{% include 'templates/head-meta.html' %}` tag, not the inlined content.

(Has something changed with regard to inlined `{% include %}` in the past few months since this PR was started?)

Given what I see written out to `dist/1.0/*.html`, I think option 1 that I described in this comment would be needed: https://github.com/Polymer/docs/pull/1641#issuecomment-224081435

It would involve basically doing what you're already doing for `/1.0/shell` in dynamicUrlToDependencies`, but doing it for `/1.0/`, `/1.0/about`, and all the other `dist/1.0/*.html` files, too.

notwaldorf:
Oh god, but if I have to list _all_ the files, then adding new files will be a disaster, and there's no point in using any of the tools like `sw-cache` anyway, no?

jeffposnick:
If you can safely assume that everything that matches `dist/1.0/*.html` will always have `{% include 'templates/head-meta.html' %}` and `{% include 'templates/site-nav.html' %}` then it's possible to automate that as part of the build. We did it for IOWA2015. I'll share some adapted code that you could reuse... stay tuned.

I know that there's templating being used elsewhere, like under `dist/1.0/docs/`, but you're using runtime caching strategies for those files, so they don't need to be versioned at build time.

gulpfile.js
```diff
@@ -82,6 +83,66 @@ function createReloadServer() {
...
+    cacheId: 'polymerproject',
+    staticFileGlobs: [
+      `${rootDir}/1.0/index.html`,
+      `${rootDir}/1.0/shell`,
+      `${rootDir}/1.0/about.html`,
+      `${rootDir}/elements/**`,
+      `${rootDir}/js/*.js`,
+      `${rootDir}/css/*.css`,
+      `${rootDir}/bower_components/**/webcomponents-lite.min.js`,
+      `${rootDir}/bower_components/highlight/highlight.js.js`,
```
jeffposnick: `highlight.js.js` ➡️  `highlight.js`

gulpfile.js
```diff
@@ -82,6 +83,66 @@ function createReloadServer() {
...
 
+function writeServiceWorkerFile() {
+  let path = require('path');
+  let rootDir = 'dist';
+
+  let config = {
+    cacheId: 'polymerproject',
+    staticFileGlobs: [
+      `${rootDir}/1.0/index.html`,
+      `${rootDir}/1.0/shell`,
```
arthurevans: This shouldn't be listed in both `staticFileGlobs` and `dynamicUrlToDependencies`, I think.

jeffposnick:
Here's the code to generate `dynamicUrlToDependencies`, dynamically:

```js
let path = require('path');
let glob = require('glob');

let rootDir = 'dist';
let dynamicUrlToDependencies = {};
let partialTemplateFiles = ['head-meta.html', 'site-nav.html']
  .map(file => path.join(rootDir, 'templates', file));
glob.sync(`${rootDir}/1.0/*.html`).forEach(htmlFile => {
  let dynamicUrl = '/1.0/' + path.basename(htmlFile).replace('.html', '');
  let dependencies = partialTemplateFiles.concat(htmlFile);
  dynamicUrlToDependencies[dynamicUrl] = dependencies;
  // Treat /1.0/ like /1.0/index
  if (dynamicUrl === '/1.0/index') {
    dynamicUrlToDependencies['/1.0/'] = dependencies;
  }
});

let config = {
  cacheId: 'polymerproject',
  staticFileGlobs: [
    `${rootDir}/elements/**`,
    `${rootDir}/js/*.js`,
    `${rootDir}/css/*.css`,
    `${rootDir}/bower_components/**/webcomponents-lite.min.js`,
    `${rootDir}/bower_components/highlight/highlight.js`,
  ],
  dynamicUrlToDependencies: dynamicUrlToDependencies,
  // ... etc ...
```

Unfortunately, with that resolved, I'm testing a little bit more now and I'm seeing that `navigateFallback` isn't playing nicely with the `<devguide-shell>` element. Initial navigations to, e.g., http://localhost:8080/1.0/docs/devguide/feature-overview will result in an empty page (basically, what you get at `/1.0/shell`), as if the logic to dynamically insert the underlying content into `<devguide-shell>` isn't run. Once you click around to other pages the `<devguide-shell>` content appears properly, so it's only affecting initial navigations, not subsequent SPA-style fake navigations.

I'm poking around to see if it's an issue that could be fixed by tweaking `<devguide-shell>`, or bad behavior in the service worker.

jeffposnick:
Okay! I see what was going on with the initial navigations. There was actually some code in `<pw-shell>` that preventing the content replacement from happening if the current URL matches the URL of the dynamic content:

https://github.com/Polymer/docs/blob/service-worker/app/elements/pw-shell.html#L277

Removing that `if()` gets initial navigations working. I am not familiar with that code and whether there's any serious downside from removing the `if()`...

arthurevans:
I think that's designed to avoid having the contents load twice when you initially load a page (non-SW-controlled). Ideally, I think we'd avoid _making_ the XHR request in this case, since the page has already been loaded with pre-rendered content. But offhand, I don't know a good way of determining which scenario we're in.

gulpfile.js
jeffposnick: Right, the code snippet I provided at https://github.com/Polymer/docs/pull/1641#issuecomment-241100911 removes a few unneeded files from `staticFileGlobs`.

FluorescentHallucinogen:
Any progress?

FluorescentHallucinogen:
Any progress?

FluorescentHallucinogen:
Any progress?

notwaldorf:
@arthurevans @jeffposnick @keanulee I know you never thought this day would come, but this is ready for another review. Live here: https://service-worker-dot-polymer-project.appspot.com

keanulee:
Looks great! One thing I noticed is that the homepage alone has more than 10 images, so it may be a good idea to bump the number of entries for the images cache (though there is a tradeoff to be made between that and potentially bloating the cache).

jeffposnick:
👍👍👍 

My one concern is that there might be more pages buried somewhere deep in the docs that use `<iframe>`s in the same way they're used on the homepage, and would require adjusting the negated "whitelist" for App Shell navigations.

Tweaking options like the size of the image cache is definitely a good idea, but there isn't a one-size-fits-all answer there.

arthurevans:
Good catch @jeffposnick. This is breaking the demos on the quick tour pages.

I think we can probably at up a convention of using /1.0/samples/ and /2.0/samples for these, so we don't end up adding more exceptions willy-nilly. Pretty sure we already serve them statically.

arthurevans:
(in the future, we hope to add a whole lot more live demos, so that means more iframes. But the source for those iframes can all live under samples.)

notwaldorf:
Ok so a bunch of things:
- @arthurevans @jeffposnick moved the `homepage` demos under `samples/`. I didn't add a redirect because we don't link directly to it, so i doubt anyone else does
- arthur: also deleted the 2.0/homepage demo altogether since we have to use the 1.0 for now, since the site is in 1.0. lmk if you disagree)
- @keanulee bumped the cache size.
- only the main index page gets 100/100 on lighthouse. this is because the other pages have a "display nothing when JS is disabled" problem, which happens from when I tried to optimize some FOUC away 😩. i'll fix it in a separate PR.

🕶 https://service-worker-dot-polymer-project.appspot.com

templates/head-meta.html
```diff
@@ -10,7 +10,7 @@
 
 <link rel="publisher" href="https://plus.google.com/107187849809354688692">
 <link rel="shortcut icon" href="/images/logos/p-logo-32.png">
-<link rel="stylesheet" href="//fonts.googleapis.com/css?family=Source+Code+Pro:400,700|Roboto+Slab:400,700">
+<link rel="preload" href="//fonts.googleapis.com/css?family=Source+Code+Pro:400,700|Roboto+Slab:400,700">
```
keanulee: Seems like you don't actually load this resource anywhere else. Remove? (Some of the code is in Source Code Pro, but that seems to be already provided by the system)

notwaldorf: We use the slabs for all the titles, and Source Code Pro isn't available on _all_ systems

keanulee: I thought `link[rel=preload]` doesn't actually include the resource in the page; it only initiates the fetch (yes, it shows up in dev tools timeline, but that's only the preload). Maybe test this in a browser that doesn't have Slabs/Source Code Pro and see if it's the expected result?

jeffposnick: I saw an error message in the console about those font resources being loaded via `preload` but not actually used. My understanding is similar to what @keanulee said, in that you need to explicitly use `<link rel="stylesheet">` to put the fonts to use. `preload`'s benefit is to kick off the request to start downloading the resource earlier when it might not normally be loaded right away: https://developers.google.com/web/updates/2016/03/link-rel-preload

notwaldorf: Yeah it got a commit that deletes it (will deal without preload in a separate PR) but I'm stuck fixing  bad FOUC the shell.html introduces so that's why it's not up yet.

notwaldorf:
@keanulee removed `link preload` changed, fixed FOUC, redeployed. One last look?

gulpfile.js
```diff
@@ -85,6 +86,84 @@ function createReloadServer() {
...
+    staticFileGlobs: [
+      `${rootDir}/images/logos/p-logo.png`,
+      `${rootDir}/elements/**`,
+      `${rootDir}/js/*.js`,
+      `${rootDir}/css/*.css`,
+      `${rootDir}/bower_components/**/webcomponents-lite.min.js`,
+      `${rootDir}/bower_components/highlight/highlight.js`,
+    ],
+    dynamicUrlToDependencies: dynamicUrlToDependencies,
+    navigateFallback: '/shell',
```
keanulee: @jeffposnick I was under the impression that `navigateFallback` would only be used if the URL is not cached. (e.g. If I've been to `/community/` and I reload `/community/`, the response should be the community page). What I'm seeing is `/shell` being the response instead, which later fetches `/community/` and inserts that, but I was hoping that a cached page would be returned if it exists, and `/shell` is returned only on cache miss. That way `/shell` can be more of a "this page isn't available, but you can click around to another page you might have cached" message and `<pw-shell>` won't need to handle the SW case separately.

app/elements/pw-shell.html
```diff
@@ -365,9 +366,27 @@
...
-      if (this._currentPageURL !== this._sectionRoute.path) {
+      // We only want to switch out the app-shell content if:
+      // - we are not refreshing the navigate fallback page (the server is already
+      // serving the correct page, and we don't need to load it twice)
+      // - this is a page to page transition (i.e. not an initial navigation) or
+      // - this is an initial navigation _and_ it was served by a Service Worker,
+      // since that is going to just give us the shell.html and we need to
+      // actually swap its content out.
+      var hasServiceWorker = navigator.serviceWorker && navigator.serviceWorker.controller;
+      var isNavigateFallback = this._isInitialNavigation && this._sectionRoute.path === '/';
```
keanulee: After I have the SW installed, if I reload on a URL that isn't cached (say https://service-worker-dot-polymer-project.appspot.com/2.0/docs/devguide/properties, but really any page that's not `/`). I get `/shell` and `<pw-shell>._isInitialNavigation` is true by the time the page is done loading (`_handleResponse` is never called since you're offline and won't get a response for an uncached page). Now if I click on the logo/homepage link, the navigation fails because`isNavigateFallback` (this line) will be true.

One way to fix this is to do `this._isInitialNavigation = false;` in `_handleError` as well. The more robust way IMO is to hopefully get what I mentioned to @jeffposnick in the gulpfile comment, which is don't serve `/shell` if the content is cached so that If we do get `/shell` then we know that there's no content to swap, and future navigation wouldn't need to go through this check (in fact, pw-shell would behave the same with or without a service worker, which from a web dev point a view should be the case since the SW only acts as a network proxy).

gulpfile.js
jeffposnick: The service worker will prefer the cached copy of any navigation targets if it exists in the cache managed by `sw-precache`, but not for cached copies that are cached in one of the runtime caches. So you won't get the behavior you hope for.

(The reason for this is that `sw-precache` needs to decide synchronously whether to respond to a `fetch` or not, and it does that by comparing the request URL to a static list of URLs in its manifest. Determining whether or not an entry exists in the runtime cache can only be done asynchronously, via a `caches.match()`.)

keanulee: Is there a way to specify our own custom handler in `runtimeCaching` config and remove `navigateFallback` to get this desired behavior? Also, is this something that sw-helpers will be able to handle? Seems like a pretty good use case for non-shell type (i.e. server-generated) sites (i.e. show me the page if it has been runtime cached, or show me a fallback if it hasn't).

jeffposnick: It is possible to disable `navigateFallback` and then write your own custom logic (pulled in via `importScripts`) that responds to navigation requests using whatever logic makes the most sense.

It really sounds like you don't want to use the App Shell model at all, though...

jeffposnick: ...and just to give some more context about why the App Shell model is something we normally recommend: If you start using individual, full HTML pages cached at runtime to fulfill navigation requests, you run into problems if you ever make changes to your page layout or the list of shared subresources that are referenced in the "boilerplate" HTML.

It's easy to imagine scenarios in which you update your layout, for instance, and then all of your individually cached pages don't get those updates right away. Assuming a stale-while-revalidate (i.e. `fastest`) update strategy, they'll pick up the updates eventually, but only on a URL-by-URL basis. So if you end up with an initial navigation to a URL that was in your runtime cache but was last accessed a year ago, you'll get the layout/subresources from last year, even if you've visited other pages more recently, and that will persist throughout the session due to SPA navigations.

With the App Shell model, there's a single, canonical HTML layout + subresources that is used for all navigations, and as soon as it's updated, all subsequent navigations (both SPA and "real" navigations) will continue to use those updates.

I wrote about this in more detail at https://jeffy.info/2017/01/24/offline-first-for-your-templated-site-part-2.html#messy-updates

keanulee: @jeffposnick Just had the chance to catch up on your earlier comments in this PR (before I was mentioned in this PR). One of the reasons why the docs site doesn't follow the app shell structure is that most of the changing content on the site are the docs themselves and not the layout/structure/code. The docs are rich content HTML (e.g. headings, paragraphs, code snippets, inline iframe demos), and the structure can vary from page to page. So the best way to encode docs is in HTML and not HTML embedded in JSON (or something) that is fetched and processed by the shell. The structure of a page is primarily defined by the definition of `<pw-shell>`, so if we were to change the layout of the site, only the precached file of `<pw-shell>` would change and the content pages wouldn't need to be updated.

Another reason why the app shell pattern may not be appropriate is that we need all the content of the site to be crawlable. At time of writing, Googlebot runs JS, but many other search engines do not so if content isn't in HTML, it won't be detected (e.g. searching for "polymer square brackets data binding" - there's no page titled that or has that in meta description, but it's in the content).

I like the idea of disabling `navigateFallback` and a custom runtime caching for the docs pages (e.g. /2.0/toolbox/, /blog/, etc.), something akin to network-first with a fallback on network-miss and cache-miss. (I prefer network-first over stale-and-revalidate because we shouldn't show stale docs content if we don't have to). I'm gonna try to see if I can make this work by adding a new SW JS file and using the `importScripts` config.

keanulee:
@notwaldorf Can you deploy my latest commit to the test server so we can try it out?

notwaldorf:
@keanulee done!

arthurevans:
We can use 2.0 samples since they're in an iframe (correct me if I'm misunderstanding the question). But that doesn't have to be part of this PR.

I want to add local 2.0 components for the live demos on the quick tour instead of relying on Polygit. I think it makes sense to make that change in a separate PR after we get the SW shipped.

Yeah, I wouldn't expect any direct links to the homepage demos, and I think it makes sense to put them in /samples/ to reduce the number of special-case rules we have to deal with.

notwaldorf:
oshit what, when did this become green and merge ready? 🎉

katejeffreys:
This is approved! What else needs to be done before it can be merged (apart from resolving the merge conflict)?

notwaldorf:
Wasn't there a flash of content or something, @keanulee? I've stopped paying attention to this PR since people started disagreeing on its intent :)

notwaldorf:
shhhhh googlebot shhhh. nobody talked to you.

FluorescentHallucinogen:
@notwaldorf 
> I've stopped paying attention to this PR since people started disagreeing on its intent

I'm very interested in docs offline work. Hope it will be finally merged soon. 🙏

gulpfile.js
```diff
@@ -330,7 +330,7 @@ gulp.task('copy', 'Copy site files (polyfills, templates, etc.) to dist/', funct
       '*',
       'app/manifest.json',
       'app/service-worker.js',
-      '!{README.md,package.json,gulpfile.js,test_runner.py}',
```
arthurevans: We actually need the test_runner in here for `npm run test` to work correctly.

keanulee: This was removed in `master` - should I re-add it here?

arthurevans:
Restaged at https://service-worker-dot-polymer-project.appspot.com/. Improves the lighthouse score of the current site from 61 to 91. First-load performance remains not great, but this version doesn't change that either way.

gulpfile.js
arthurevans: D'oh. Never mind—test_runner.py was _removed_ from the  exclude list, which is correct... For some reason, this change showed up as the only relevant change in the merge commit (thx github) so I thought something had gotten reversed by accident. Plz ignore.

arthurevans:
Did a few very unscientific tests with updating content and adding nav links, and updates appear to work OK.

arthurevans:
🎆 🎆 🎆 

(and please don't let it blow up in my face)

khushalpratap:
ere

PR #1642, opened 2016-06-02
Fix some links.

arthurevans:


notwaldorf:
Actually I've noticed that I can't actually load this page, possibly because of its name and too many trailing dots? https://www.polymer-project.org/1.0/blog/2015-11-02-release-1.2.0.html is the only one that doesn't load for me

arthurevans:
This PR should fix that. That blog post has a bunch of `{% highlight html %}` tags in there that are killing Jinja.

notwaldorf:
ohhhh got it. 👍

PR #1646, opened 2016-06-04
Fix blog post titles

### LINKED ISSUE #1632: [new site] All the blog posts has the same title
The News section always has the same title metatag: **News - Polymer Project**.

screen shot 2016-05-28 at 01 38 38

I think it will have the post title.

arthurevans:
Fixes #1632

/cc: @notwaldorf 

I _think_ this is all we need to do here?

### DIFF
--- a/templates/base-blog.html
+++ b/templates/base-blog.html
@@ -10,7 +10,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 <!DOCTYPE html>
 <html itemscope itemtype="http://schema.org/Organization" lang="en">
 <head>
-<title>News - Polymer Project</title>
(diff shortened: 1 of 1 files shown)


notwaldorf:
Hmmmmmmmm. I think this will make the blog posts work, but maybe not the `blog/index.html` page, since it doesn't have a title.

arthurevans:
It seems like the title for `blog/index` is set here:

https://github.com/Polymer/docs/blob/98d192071c411d9eeceaaa0902e1c0360051910c/app/1.0/blog/index.html#L2

In local testing, it appears to show up correctly as "News — Polymer Project".

notwaldorf:
Oh in that case LGTM :shipit:

PR #1647, opened 2016-06-06
Step 2 after 1

marcus7777:


arthurevans:
Thanks for the PR. This isn't actually step 2, it's an alternative version of step 1. It's either/or. 

I think I can make this more clear by removing "Step 1" from the headings and introducing the two branches in the intro text. So I'm going to close this PR and do that instead.

marcus7777:
I think the no install version
is also a little confusing with **install** which came in so many different things bower, node, npm, applications, extensions, apps ... 
maybe 'a online way of getting set up' ?

PR #1652, opened 2016-06-09
fixing incorrect setup instructions

marcopashkov:


marcopashkov:
I signed it!

arthurevans:
The original command is correct; if you run `dev_appserver.py .` in the project directory you'll get a broken site; you need to run it against the built site in `dist/`.

PR #1653, opened 2016-06-09
Fix bad link syntax

alanorozco:


arthurevans:
Thanks for the heads-up. This is fixed in #1669 so closing here.

PR #1658, opened 2016-06-18
Add UNIX variable for auto-cd'ing into previously-made directory

kevinSuttle:
http://unix.stackexchange.com/a/219627/55531


arthurevans:
Thanks for the PR. I'm not going to merge this, but I thought I should explain why.

While this would save some keystrokes (and potential cut & paste errors), using `$_` would require explanation for those not familiar with it, and it assumes the user is using bash or one of its relatives. As written, the instructions currently work for every UNIX shell I'm aware of as well as the Windows command prompt.

PR #1665, opened 2016-06-23
Add note about bower and gulp dependency

ThomasWo:
Installing from a clean slate, `polymer init <blah>` fails silently without bower (and gulp is needed in there somewhere).  Thus I suggest adding a note about requiring the use of bower and gulp.

### DIFF
--- a/app/1.0/start/toolbox/set-up.md
+++ b/app/1.0/start/toolbox/set-up.md
@@ -16,7 +16,7 @@ App Toolbox template in less than 15 minutes.
 1.  Install the LTS version (4.x) of Node.js. The current version (6.x) should work, but is not
     officially supported. Versions below LTS are not supported.
 
-1. If you don't have them installed, install bower and gulp
+1. If you don't have bower installed, install it
 
         npm install -g bower
 
--- a/app/1.0/start/toolbox/set-up.md
+++ b/app/1.0/start/toolbox/set-up.md
@@ -16,6 +16,10 @@ App Toolbox template in less than 15 minutes.
 1.  Install the LTS version (4.x) of Node.js. The current version (6.x) should work, but is not
     officially supported. Versions below LTS are not supported.
 
+1. If you don't have them installed, install bower and gulp
+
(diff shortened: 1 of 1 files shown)


ThomasWo:
I signed it!

app/1.0/start/toolbox/set-up.md
```diff
@@ -16,6 +16,10 @@ App Toolbox template in less than 15 minutes.
 1.  Install the LTS version (4.x) of Node.js. The current version (6.x) should work, but is not
     officially supported. Versions below LTS are not supported.
 
+1. If you don't have them installed, install bower and gulp
+
+        npm install -g bower gulp
```
addyosmani: @justinfagnani Do we currently make an attempt to resolve/install these dependencies or warn the user about them not being available or is this line worth adding to the setup docs for toolbox?

FredKSchott: Gulp is no longer required by polymer-cli (it used to be). No need to mention here anymore.

ThomasWo:
@FredKSchott I've removed the gulp reference in the file, committed, and squashed.  PTAL

app/1.0/start/toolbox/set-up.md
```diff
@@ -16,6 +16,10 @@ App Toolbox template in less than 15 minutes.
 1.  Install the LTS version (4.x) of Node.js. The current version (6.x) should work, but is not
     officially supported. Versions below LTS are not supported.
 
+1. If you don't have them installed, install bower and gulp
```
justinfagnani: If you don't have bower installed, install it:

justinfagnani:
@ThomasWo are you using two different email addresses for the CLA and for commits?

ThomasWo:
@justinfagnani Looks like the git config on one of my machines was set to the wrong account for some reason, so yes, but it shouldn't be a problem as both accounts have signed the CLA.

justinfagnani:
Thanks @ThomasWo!

PR #1670, opened 2016-06-29
Remove seemingly errant `<var>` in mixin example

kevinpschaaf:


kevinpschaaf:
Noticed this when cross-referencing docs for release notes... not sure if that `<var>` is some sort of styling/metadata that's not being applied or what...

arthurevans:
Oops. Yeah, it's just the standard HTML `<var>` tag. We use it for placeholder text. But it doesn't work inside a Markdown code block, so we need to use `<pre><code>` to wrap 'em instead.

arthurevans:
Fixed this in #1717. Thanks for the heads-up!

PR #1673, opened 2016-07-01
add 1.6 release notes

kaycebasques:


arthurevans:
@kevinpschaaf @sorvell or @azakus can one of you take a look at the descriptions here?

kevinpschaaf:
Yep, LGTM!

app/1.0/docs/release-notes.md
```diff
@@ -11,6 +11,107 @@ title: Release notes
...
+
+    The following use of CSS custom properties inside `@media` rules is
+    now supported, allowing for dynamic change of custom property values
+    based on screen size, etc.:
+
+    ```html
+    <style>
+      @media (max-width: 400px) {
+        :host {
+          --gutter-width: 16px;
```
robdodson: @kevinpschaaf does this work even in browsers which don't support custom properties?

kevinpschaaf: @robdodson Ah good point.  We need to add another caveat there.  Users will need to call `Polymer.updateStyles()` to re-evaluate the media queries upon window resize.

We didn't include a window resize listener in the core to allow the user to choose an appropriate strategy for calling `updateStyles`, since there is a performance/responsiveness tradeoff depending on whether it is called on every resize or throttled somehow.

Here's a jsBin that shows the basic code to add to an app to ensure styles are updated on window resize, throttled to one update per 100ms:
http://jsbin.com/bicazocaji/edit?html,output

kevinpschaaf: Essentially,

```js
    (function() {
      var resizeDebouncer;
      addEventListener('resize', function() {
        resizeDebouncer = Polymer.Debounce(resizeDebouncer, Polymer.updateStyles, 100);
      });
    })();
```

robdodson: Could this be something I opt into as a top level setting? Personally I may be happy to sacrifice the perf for the increased convenience

kaycebasques:
@arthurevans @robdodson @kevinpschaaf added a thingy about browsers that don't support custom properties

arthurevans:
LGTM, thanks!

PR #1674, opened 2016-07-04
Fix missing quotation mark

jart:


arthurevans:
LGTM. Thanks for the fix!

PR #1677, opened 2016-07-11
Unique memcache ID per app version. Fixes #1676.

### LINKED ISSUE #1676: Need to separate memcached files for non-production sites
_From @arthurevans on May 26, 2016 22:9_

Currently, if we flush memcache, whoever runs next updates memcache. If that's a non-prod site, the non-prod files get cached for all sites—including prod.

_Copied from original issue: Polymer/docs2#296_

arthurevans:
R: @ebidel 

(There was another bug about flushing memcache on deploy, but it turns out that alone doesn't work, since the staged version may have different values in memcache than the live version. This fixes that, and as long as you always bump the app version with any major changes, you don't even need to flush memcache.)

### DIFF
--- a/server.py
+++ b/server.py
@@ -50,8 +50,18 @@ env = jinja2.Environment(
 env.globals['include_file'] = include_file
 env.globals['include_file_raw'] = include_file_raw
 
+# memcache logic: maintain a separate cache for each explicit
+# app version, so staged versions of the docs can have new nav
+# structures, redirects without affecting other deployed docs.
+
+# CURRENT_VERSION_ID format is version.hash, where version is the
+# app version passed to the deploy script.
+if 'CURRENT_VERSION_ID' in os.environ:
+  MEMCACHE_PREFIX = os.environ.get('CURRENT_VERSION_ID').split('.')[0] + '/'
+else:
+  MEMCACHE_PREFIX = 'no_version/'
 REDIRECTS_FILE = 'redirects.yaml'
-NAV_FILE = '%s/nav.yaml'
+NAV_FILE =  '%s/nav.yaml'
 ARTICLES_FILE = '%s/blog.yaml'
 AUTHORS_FILE = '%s/authors.yaml'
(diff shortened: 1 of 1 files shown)


arthurevans:
Staged (with other content changes) at:

 https://data-system-dot-polymer-project.appspot.com/1.0/docs/devguide/feature-overview 

Live docs have different nav structure under "Data system," which wasn't possible before:

https://www.polymer-project.org/1.0/docs/devguide/feature-overview

server.py
```diff
@@ -50,8 +50,18 @@ def include_file_raw(name):
...
+# memcache logic: maintain a separate cache for each explicit
+# app version, so staged versions of the docs can have new nav
+# structures, redirects without affecting other deployed docs.
+
+# CURRENT_VERSION_ID format is version.hash, where version is the
+# app version passed to the deploy script.
+if 'CURRENT_VERSION_ID' in os.environ:
+  MEMCACHE_PREFIX = os.environ.get('CURRENT_VERSION_ID').split('.')[0] + '/'
+else:
+  MEMCACHE_PREFIX = 'no_version/'
```
ebidel: this should really be above the `if` (as the default).

server.py
```diff
@@ -50,8 +50,18 @@ def include_file_raw(name):
...
 REDIRECTS_FILE = 'redirects.yaml'
-NAV_FILE = '%s/nav.yaml'
+NAV_FILE =  '%s/nav.yaml'
```
ebidel: remove extra space

ebidel:
LGTM. Just the nits.

server.py
arthurevans: You mean like: 

```
MEMCACHE_PREFIX = 'no_version/'
if 'CURRENT_VERSION_ID' in os.environ:
  MEMCACHE_PREFIX = os.environ.get('CURRENT_VERSION_ID').split('.')[0] + '/'
```

?

ebidel: Yep

PR #1679, opened 2016-07-17
fix a typo

emmiter:
It seems that ***not*** may leave out in line 13.

### DIFF
--- a/app/1.0/docs/devguide/local-dom.md
+++ b/app/1.0/docs/devguide/local-dom.md
@@ -10,7 +10,7 @@ its _light DOM_ for clarity.
 
 Polymer supports multiple local DOM implementations. On browsers
 that support shadow DOM, shadow DOM may be used to create local DOM. On other
-not supported browsers, Polymer provides local DOM via a custom
+browsers, Polymer provides local DOM via a custom
 implementation called _shady DOM_ which is inspired by shadow DOM.
 
 Shady DOM requires you to use the [Polymer DOM API](#dom-api)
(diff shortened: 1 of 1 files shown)


emmiter:
I signed it!

app/1.0/docs/devguide/local-dom.md
```diff
@@ -10,7 +10,7 @@ its _light DOM_ for clarity.
 
 Polymer supports multiple local DOM implementations. On browsers
 that support shadow DOM, shadow DOM may be used to create local DOM. On other
-supported browsers, Polymer provides local DOM via a custom
+not supported browsers, Polymer provides local DOM via a custom
```
arthurevans: "Supported" could mean "supported by Polymer" or "that support shadow DOM". I think the original intention was probably the first one, but it's quite confusing. I think a better fix would be to remove the word "Supported" entirely, "On other browsers ..."

emmiter: @arthurevans

emmiter: @arthurevans In short,  I agree with the solution "On other browsers..." because this expression could be understanded clearly both in english and in chinese. 

However,  between "supported by Polymer" and "that support shadow DOM" , considering the context I tend to think it means the second one both in the original intention, and in the later better fix "On other browsers..." , um... at least look at it in chinese. 

If it refres to "supported by Polymer",then the sentence becomes "On other supported browsers *by Polymer* ...". With the subsequent lines, I really think it doesn't make sense.

Really appreciate for your reading : )

arthurevans:
LGTM. Thanks for the fix!

PR #1686, opened 2016-07-25
Uwaterloo 210J pfryerda edits

pfryerda:
Now against data-systems rather than master.

### DIFF
--- a/app/1.0/docs/devguide/data-system.md
+++ b/app/1.0/docs/devguide/data-system.md
@@ -326,16 +326,16 @@ Related tasks:
 Polymer implements the mediator pattern, where a host element manages data flow between itself and
 its local DOM nodes.
 
-When two elements are connected with a data binding, data changes can flow *downward*, that is, from
-host to target, *upward*, from target to host, or both ways.
+When two elements are connected with a data binding, data changes can flow **downward** — from
+host to target — **, upward** — from target to host — , or both.
 
-When two elements in the local DOM are bound to the same property, data appears to flow from one
-element to the other, but this flow is *mediated* by the host. A change made by one element
-propagates *up* to the host, then the host propagates the change *down* to the second element.
+When two elements in the local DOM are bound to the same property data appears to flow from one
+element to the other, but this flow is **mediated** by the host. A change made by one element
+propagates **up** to the host, then the host propagates the change **down** to the second element.
 
 ### Data flow is synchronous
 
-Data flow is **synchronous**. So when your code makes an [observable change](#observable-changes),
+Data flow is **synchronous**. When your code makes an [observable change](#observable-changes),
 all of the data flow and property effects from that change occur before the next line of your
 JavaScript is executed, unless an element explicitly defers action (for example, by calling an
 asynchronous method).
@@ -353,8 +353,8 @@ The two types of data binding annotations are:
 
 
 
-*   Automatic, which allows upward (target to host) and downwards (host to target) data flow.
-    Automatic bindings use curly brackets (`{{ }}`):
+*   **Automatic**, which allows upward (target to host) and downwards (host to target) data flow.
+    Automatic bindings use double curly brackets (`{{ }}`):
 
 
 ```
@@ -363,8 +363,8 @@ The two types of data binding annotations are:
 ```
 
 
-*   One-way, which only allows downwards data flow. Upward data flow is disabled. One-way bindings
-    use square brackets (`[[ ]]`).
+*   **One-way**, which only allows downwards data flow. Upward data flow is disabled. One-way bindings
+    use double square brackets (`[[ ]]`).
 
 
 ```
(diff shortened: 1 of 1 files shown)


app/1.0/docs/devguide/data-system.md
```diff
@@ -326,16 +326,16 @@ Related tasks:
 Polymer implements the mediator pattern, where a host element manages data flow between itself and
 its local DOM nodes.
 
-When two elements are connected with a data binding, data changes can flow *downward*, that is, from
-host to target, *upward*, from target to host, or both ways.
+When two elements are connected with a data binding, data changes can flow **downward** — from
```
arthurevans: We use italics typically when introducing a new term or defining one, so I think the italics are correct here.

We don't use space around an em dash. In other words—just the dash.

app/1.0/docs/devguide/data-system.md
```diff
@@ -326,16 +326,16 @@ Related tasks:
...
+host to target — **, upward** — from target to host — , or both.
```
arthurevans: Don't combine both the dash and comma to set off a clause. Choose one or the other.

"or both" => both what? There are several possible referrents in this sentence, including "two elements" and "data changes" ... "both ways" is an extra word, but I think it removes ambiguity here.

app/1.0/docs/devguide/data-system.md
```diff
@@ -326,16 +326,16 @@ Related tasks:
...
 
-When two elements in the local DOM are bound to the same property, data appears to flow from one
-element to the other, but this flow is *mediated* by the host. A change made by one element
-propagates *up* to the host, then the host propagates the change *down* to the second element.
+When two elements in the local DOM are bound to the same property data appears to flow from one
+element to the other, but this flow is **mediated** by the host. A change made by one element
```
arthurevans: Technically, we're introducing a new term here, so mediated should be italic. In practice, we don't yet have a good definition in place, and the reader is left to infer the definition from the following text. Leave it italics or remove the emphasis, but don't make it bold and call extra attention to my failure here :D

app/1.0/docs/devguide/data-system.md
```diff
@@ -526,7 +526,7 @@ A change notification event is fired when one of the following
 
 *   An array mutation.
 
-The event type follows a naming convention of <code><var>property</var>-changed</code>, where
+Event types follow a naming convention of <code><var>property</var>-changed</code>, where
```
arthurevans: So this is an interesting one. The JavaScript `Event` object has a [type property](https://developer.mozilla.org/en-US/docs/Web/API/Event/type), which defines the well... type of event that occurred. "Event types" suggests that we're talking about all events, everywhere, but in this case we're talking Polymer's change notification events, and "The" refers back to "A change notification event" in the previous sentence.

This could be expanded to something like: "The event's type property indicates which property changed: it follows the naming convention of blahditty blah."

app/1.0/docs/devguide/data-system.md
```diff
@@ -539,7 +539,7 @@ The contents of the event vary depending on the change.
 
 Native elements like `<input>` don't provide the change notification events that Polymer uses for
 upward data flow. To support two-way data binding of native input elements, Polymer lets you
-associate a _custom change notification event_ with a data binding. For example, when binding to a
```
arthurevans: Again, we're essentially introducing new terminology here, and the italics are appropriate.

app/1.0/docs/devguide/data-system.md
```diff
@@ -562,8 +562,7 @@ Related tasks:
 
 ### Element initialization
 
-When an element initializes its local DOM, it configures the properties of its local DOM children
-initializes data bindings.
+When an element initializes its local DOM, it configures the properties of its local DOM children's data bindings.
```
arthurevans: Good catch. What's missing here is actually the word "and" between "configures the properties of its local DOM children **and** initializes data bindings."

app/1.0/docs/devguide/data-system.md
```diff
@@ -480,22 +480,22 @@ propagates data downward by setting the target element’s properties or invokin
...
 Polymer elements use events to propagate data upward. The target element fires a non-bubbling event
-when an observable change occurs. (Change events are described in more detail in [Change
-notification events](#link).)
+when an observable change occurs. Change events are described in more detail in [Change
+notification events](#link).
 
-For two-way bindings, the host element listens for these change events and propagates the
-changes—for example, setting a property and invoking any related property effects. The property
+For **two-way bindings**, the host element listens for these change events and propagates the
+changes — for example, setting a property and invoking any related property effects. The property
```
arthurevans: Per previous note, no space around em dashes.

arthurevans:
Thanks @pfryerda—reviewed and left comments.

app/1.0/docs/devguide/data-system.md
pfryerda: Would you rather see `_text_` for italics or `*text*` for italics?

arthurevans: Good question. I'd go with `_text_` because it's easier to distinguish from `**bold text**`. (Unfortunately, the conversion script I used to turn Docs into Markdown had different preferences.)

pfryerda: I've changed to all of the italics in the section I looked at using `_text_`.

pfryerda:
@arthurevans You should see the changes you commented on. Let me know if there's anything more.

arthurevans:
LGTM. Thanks.

PR #1690, opened 2016-07-28
DOMs : Add schema to complete the text explanation

alemarre:
Hi, 
The text explanation was a bit confusing for me at the first read (a lot a "new" terms  : local dom, light dom, shadow dom, shady dom)
I think a schema could help understanding

### DIFF
--- a/app/1.0/docs/devguide/local-dom.md
+++ b/app/1.0/docs/devguide/local-dom.md
@@ -18,6 +18,25 @@ when manipulating DOM from JavaScript. This interface covers most of the
 common DOM methods and properties, and is compatible with both shady DOM and
 native shadow DOM.
 
+```html
+<dom-module>
+  <template>
+  
+    <!-- 
+      Local DOM. Possible implementations : 
+        - Shadow DOM (browser implementation)
+        - Shady DOM (Polymer implementation)
+    -->
+    
+    <content>
+      <!-- 
+        Light DOM
+      -->
+    </content>
+    
+  </template>
+</dom-module>
+```
 
 **Note:** Currently Polymer uses shady DOM by default on all
 browsers. To opt-in to using shadow DOM where available, see [Global


alemarre:
I signed it

app/1.0/docs/devguide/local-dom.md
```diff
@@ -18,6 +18,25 @@ when manipulating DOM from JavaScript. This interface covers most of the
...
+<dom-module>
+  <template>
+  
+    <!-- 
+      Local DOM. Possible implementations : 
+        - Shadow DOM (browser implementation)
+        - Shady DOM (Polymer implementation)
+    -->
+    
+    <content>
```
stramel: `<content>` should be `<slot>` now...

katejeffreys: This is a suggested edit to a 1.0 doc so it should stay `<content>`

app/1.0/docs/devguide/local-dom.md
```diff
@@ -18,6 +18,25 @@ when manipulating DOM from JavaScript. This interface covers most of the
...
+      <!-- 
+        Light DOM
+      -->
+    </content>
+    
+  </template>
+</dom-module>
+```
```
katejeffreys: Something seems not quite right about this. I will try to explain my thoughts.

I read the the comments in the `<template>` section as instructing me to insert my own code here, and that my code must be either: 

* A "Shadow DOM (browser implementation)", or
* A "Shady DOM (Polymer implementation)"

I would not know what either of these things are, or how to write them. 

Thus I find this suggested addition to be more confusing rather than clarifying.

But I will ask @arthurevans to comment here and will trust his opinion on the matter.

app/1.0/docs/devguide/local-dom.md
stramel: so we should keep it `<content>` even though 1.7+ can make use of slots and reduces changes going to 2.0?

katejeffreys: ¯\_(ツ)_/¯ Or note hybrid syntax?

PR #1691, opened 2016-07-28
Moving style block to <dom-module>

delki8:
Following the recommendation at https://www.polymer-project.org/1.0/docs/devguide/styling, moving the <style> block from within <template> to <dom-module>

### DIFF
--- a/app/1.0/docs/devguide/feature-overview.md
+++ b/app/1.0/docs/devguide/feature-overview.md
@@ -17,12 +17,13 @@ A basic Polymer element definition looks like this:
 
 ```
     <dom-module id="element-name">
-
+      
+      <style>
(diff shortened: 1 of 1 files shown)


delki8:
I signed it!

arthurevans:
The current placement, inside the template, is correct. The styling doc you cited says:

Scoped styles should be provided via `<style>` tags placed inside the element's local DOM `<template>`.

You might have been confused by a later note, which describes the previous recommendations:  "**Prior to Polymer 1.1,** the recommendation was to place `<style>` tags inside the `<dom-module>` for an element (but outside the `<template>`). This is still supported, **but is no longer recommended**." (emphasis added.)

delki8:
Ok, no problem ;)

PR #1699, opened 2016-08-03
Data system doc update

arthurevans:


tjsavage:
Love the update!! Awesome new content. Just two little notes:

- White box around text in the images is a little distracting, would it be possible to remove/make it match the background?
screen shot 2016-08-03 at 4 07 48 pm

- I wish we could make the elements inside "helper elements" section a little more surfaced. They're hugely useful, but I feel like buried under the "Helper Elements" header.  I feel like maybe "Array Repeats & Conditionals" might be possible to break out into a separate page with that header, to make it clear what's in there?

app/1.0/docs/devguide/data-binding.md
```diff
@@ -59,18 +91,21 @@ For example:
...
+To bind to a target element's `textContent`, you can simply include the
+annotation or compound binding inside the target element.
 
 ```
 <dom-module id="user-view">
 
     <template>
-      First: {{firstName}}<br>
-      Last: {{lastName}}
+      <div>{{firstName}}</div>
```
ebidel: Should use use `lastName` as well?

app/1.0/docs/devguide/data-binding.md
```diff
@@ -59,18 +91,21 @@ For example:
...
 
-To bind to a child element's `textContent`, you can simply include the
-annotation inside the child element.
+To bind to a target element's `textContent`, you can simply include the
+annotation or compound binding inside the target element.
 
 ```
 <dom-module id="user-view">
 
     <template>
```
ebidel: indentation on all code samples should be 2 spaces

app/1.0/docs/devguide/data-binding.md
```diff
@@ -4,39 +4,71 @@ title: Data binding
...
+[data path](data-system#data-paths), or data generated based on one or more paths.
 
-A binding is created with a binding _annotation_ in the host element's local DOM template:
+You create data bindings by adding annotations to an element's local DOM template.
 
 ```
 <dom-module id="host-element">
     <template>
-      <child-element name="{{myName}}"></child-element>
+      <target-element target-property="{{hostProperty}}"></target-element>
```
ebidel: would it be nice to mention early on that `{{}}` are two-way, if the target-element opts-in?

app/1.0/docs/devguide/data-binding.md
```diff
@@ -88,412 +123,257 @@ annotation inside the child element.
...
-    }
-  });
-</script>
-...
-<!-- changes to "value" are ignored by child due to readOnly:true -->
-<!-- changes to "someProp" are ignored by host due to square-bracket syntax -->
-<!-- binding serves no purpose -->
-<custom-element some-prop="[[value]]"></custom-element>
-```
+Attribute bindings are always one-way, host-to-target. Values are serialized according to the
```
ebidel: is this true? If the host updates a property (say boolean), won't it add the attribute?

app/1.0/docs/devguide/data-binding.md
```diff
@@ -88,412 +123,257 @@ annotation inside the child element.
...
+<div class$="{{foo}}"></div>
 
-## Binding to structured data {#path-binding}
+<!-- style -->
+<div style$="{{background}}"></div>
 
-You can bind sub-properties of the host by specifying a _path_ inside
-the binding annotation.
+<!-- href -->
+<a href$="{{url}}">
```
ebidel: all of these native ones should probably use `[[]]`

app/1.0/docs/devguide/data-binding.md
```diff
@@ -583,17 +466,56 @@ Finally, if a computed binding has no dependent properties, it is only evaluated
...
 ```
 
-## Binding to array items {#array-binding}
+## Compound bindings {#compound-bindings}
+
+You can combine string literals and bindings in a single property binding or
+text content binding. For example:
+
+```
+<img src$="https://www.example.com/profiles/{{userId}}.jpg">
```
ebidel: these should probably be `[[]]`

app/1.0/docs/devguide/properties.md
```diff
@@ -817,96 +343,16 @@ generated setter of the convention <code>\_set<var>Property</var>(value)</code>.
 ```
 
 For more on read-only properties and data binding, see
-[Property change notification and two-way binding](data-binding#property-notification).
-
-## Computed properties {#computed-properties}
-
-Polymer supports virtual properties whose values are calculated from other
-properties.
+[How data flow is controlled](data-system#data-flow-control).
```
ebidel: these "How data flow is controlled" links all link to this page. Is that correct? They don't go anywhere.

app/1.0/docs/devguide/data-binding.md
```diff
@@ -4,39 +4,71 @@ title: Data binding
...
+(the _host element_) to a property or attribute of an element in its local DOM (the _child_
+or _target element_). The host element data can be a property or sub-property represented by a
+[data path](data-system#data-paths), or data generated based on one or more paths.
 
-A binding is created with a binding _annotation_ in the host element's local DOM template:
+You create data bindings by adding annotations to an element's local DOM template.
 
 ```
 <dom-module id="host-element">
     <template>
```
robdodson: indentation seems off. 4 spaces for <template>, then 2 spaces for <target-element>

app/1.0/docs/devguide/data-binding.md
```diff
@@ -4,39 +4,71 @@ title: Data binding
...
     </template>
 </dom-module>
 ```
 
-## Binding annotations {#property-binding}
+Updating data bindings is a [property effect](data-system#property-effects).
 
-A binding annotation consists of a property name or subproperty name enclosed
-in curly brackets (`{{}}`) or square brackets (`[[]]`).
+## Anatomy of a data binding
```
addyosmani: It might be worth calling out the differences in opt-in approach between one and two way data-binding early on.

app/1.0/docs/devguide/data-binding.md
```diff
@@ -88,412 +123,257 @@ annotation inside the child element.
...
-<!-- Listens for `input` event and sets hostValue to <input>.value -->
-<input value="{{hostValue::input}}">
+There are various reasons that properties can't be bound:
 
-<!-- Listens for `change` event and sets hostChecked to <input>.checked -->
-<input type="checkbox" checked="{{hostChecked::change}}">
-
-<!-- Listens for `timeupdate ` event and sets hostTime to <video>.currentTime -->
-<video url="..." current-time="{{hostTime::timeupdate}}">
+*   Cross-browser isssues with the ability to place binding braces `{{...}}` in some of these
```
addyosmani: Is this a polyfill issue worth linking up to?

app/1.0/docs/devguide/data-binding.md
```diff
@@ -88,412 +123,257 @@ annotation inside the child element.
...
+<!-- label for -->
+<label for$="{{bar}}"></label>
 
-```
-<template>
-  <div>{{user.manager.name}}</div>
-  <user-element user="{{user}}"></user-element>
-</template>
+<!-- dataset -->
+<div data-bar$="{{baz}}"></div>
```
robdodson: could maybe add an aria example after this one?

```
<!-- ARIA -->
<button aria-label$="[[bar]]"></button>
```

app/1.0/docs/devguide/data-binding.md
```diff
@@ -59,18 +91,21 @@ For example:
 syntax. For more information, see [Binding to native element attributes](#native-binding).
 { .alert .alert-info }
 
+If the property being bound is an object or an array, both elements get a reference to the **same
+object**. This means that either element can change the object and a true one-way binding is
+impossible. For more information, see [Data flow for objects and
```
robdodson: I haven't tried it but I wonder if this is true if the object was frozen using [Object.freeze](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze)

app/1.0/docs/devguide/data-binding.md
```diff
@@ -88,412 +123,257 @@ annotation inside the child element.
...
-    <template>
-      <div hidden="{{!enabled}}"></div>
-    </template>
-    ```
+Only a single logical not operator is supported. You can't cast a value to boolean using `!!`. For
+more complicated transformations, use a [computed binding](#annotated-computed).
 
-*   Computed bindings, as described in
-    [Computed bindings](#annotated-computed), below.
+A binding with a logical not operator is **always one-way, host-to-target**.
```
robdodson: Should the example use `[[]]` to reinforce this?

app/1.0/docs/devguide/data-binding.md
```diff
@@ -514,46 +394,49 @@ Example: { .caption }
 ```
 
 In this case, the span's `textContent` property is bound to the return value
-of `computeFullName`, which is recalculated whenever `first` or `last` changes.
+of `_formatName`, which is recalculated whenever `first` or `last` changes.
 
-### Dependent properties in computed bindings {#dependent-properties}
+**Computed bindings are one way.** A computed binding is always one-way, host-to-target.
```
robdodson: Same comment as before, should these examples that are always one way use `[[]]` to reinforce the point?

app/1.0/docs/devguide/data-binding.md
```diff
@@ -643,92 +565,65 @@ _or_ if the array itself is mutated, so the binding uses a wildcard path, `myArr
 </dom-module>
```
robdodson: For some reason I can't do a line comment on line 544 but there's an extra space before the closing ] and an extra carriage return on 541

arthurevans:
@tjsavage With the background color removed, the labels are hard to read. Would love to flip the site's color scheme so that main content is on a white background, which would fix this. I kind of hesitate to bake our weird background color into the diagram; could probably get it to work, though.

Alternately, we could replace these labels with numbers in a colored bubble, which would be better for localization of the diagrams.

app/1.0/docs/devguide/data-binding.md
arthurevans: It's host-to-target, so if the host updates the property, the attribute will be added, but if the target element toggles the attribute, the host property is never updated.

app/1.0/docs/devguide/data-binding.md
arthurevans: Ah, yeah. simplified to just a single name in this example instead of having the extra property.

app/1.0/docs/devguide/data-binding.md
arthurevans: Whoop. Not sure how the weird indentation got in there, but I think I got it all.

app/1.0/docs/devguide/data-binding.md
arthurevans: Fixed.

app/1.0/docs/devguide/properties.md
arthurevans: These should be fixed now.

app/1.0/docs/devguide/data-binding.md
arthurevans: Fixed.

app/1.0/docs/devguide/data-binding.md
arthurevans: I have never been able to track down what exactly the issues are here. So, not sure what we'd say.

app/1.0/docs/devguide/data-binding.md
arthurevans: Done, thanks.

app/1.0/docs/devguide/data-binding.md
arthurevans: Yep, done.

app/1.0/docs/devguide/data-binding.md
arthurevans: Fixed, thanks.

arthurevans:
Hey all, I've staged the latest version: https://data-system-dot-polymer-project.appspot.com ... I'm planning to roll it out tomorrow afternoon if no one spots a showstopper. Sanity checks greatly appreciated.

app/1.0/docs/devguide/data-system.md
```diff
@@ -0,0 +1,788 @@
...
+## Data paths {#paths}
+
+In the data system, a _path_ is a string that identifies a property or subproperty *relative to a
+scope*. In most cases, the scope is a host element. For example, consider the following
+relationships:
+
+![Two elements, user-profile and address card. The user-profile element has a primaryAddress
+property. An arrow labeled 1 connects the property to a JavaScript object. The address-card
+element has an address property. An arrow labeled 2 connects the property to the same JavaScript
+object.](/images/1.0/data-system/data-binding-paths.png)
```
abdonrd: @arthurevans oops!

screen shot 2016-08-20 at 11 15 35

arthurevans:
Thanks @abdonrd ... fixed now.

abdonrd:
Your welcome @arthurevans! :)

PR #1701, opened 2016-08-06
Updated step-2.md

oopg:
iron-icon to be used instead of icon-toggle.


arthurevans:
The doc is correct as it is. The _demo_ file imports and uses `icon-toggle.html`:

https://github.com/googlecodelabs/polymer-first-elements/blob/master/demo/icon-toggle-demo.html#L3

https://github.com/googlecodelabs/polymer-first-elements/blob/master/demo/icon-toggle-demo.html#L16

PR #1703, opened 2016-08-08
Update polymer-cli docs to mention /etc/hosts file

JimPanic:
Regarding https://github.com/Polymer/polymer-cli/issues/312

### DIFF
--- a/app/1.0/docs/tools/polymer-cli.md
+++ b/app/1.0/docs/tools/polymer-cli.md
@@ -271,6 +271,8 @@ Serve from hostname `test` (e.g. app project demo is available at
 
     polymer serve --hostname 'test'
 
+**Note:** Make sure your computer actually knows this hostname by putting it into your [`/etc/hosts` file](https://en.wikipedia.org/wiki/Hosts_(file)).
+
 Open up a page other than the default `index.html` in a specific browser
 (Apple Safari, in this case):
 
--- a/app/1.0/docs/tools/polymer-cli.md
+++ b/app/1.0/docs/tools/polymer-cli.md
@@ -271,7 +271,7 @@ Serve from hostname `test` (e.g. app project demo is available at
 
     polymer serve --hostname 'test'
 
-**Note:** Make sure your computer actually knows this hostname by putting it into your [`/etc/hosts` file](https://en.wikipedia.org/wiki/Hosts_(file)).
+**Note:** Make sure your computer actually knows this hostname by putting it into your [`hosts` file](https://en.wikipedia.org/wiki/Hosts_(file)).
 
 Open up a page other than the default `index.html` in a specific browser
 (Apple Safari, in this case):
--- a/app/1.0/docs/tools/polymer-cli.md
+++ b/app/1.0/docs/tools/polymer-cli.md
@@ -266,8 +266,8 @@ Serve from port 3000:
 
     polymer serve --port 3000
 
-If you have configured a custom hostname on your machine, polymer can serve it with the
-`--hostname` argument (e.g. app project demo is available at `http://test:8080`):
+If you have configured a custom hostname on your machine, Polymer CLI can serve it with the
+`--hostname` argument (for example, app project demo is available at `http://test:8080`):
 
     polymer serve --hostname test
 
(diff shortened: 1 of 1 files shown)


JimPanic:
I signed it!

arthurevans:
@justinfagnani can you take a look at this? Is this correct, actionable, portable? I haven't used this particular feature.

justinfagnani:
I'm not sure sure about this. It's very UNIX-centric, and documents something arguably outside the scope of the CLI: you're local computer needs to have it's networking properly configured to reach a host.

justinfagnani:
I do think the quotes around 'test' need to be removed.

JimPanic:
> I'm not sure sure about this. It's very UNIX-centric, and documents something arguably outside the scope of the CLI: you're local computer needs to have it's networking properly configured to reach a host.

This is very much the scope of the CLI: it is a feature to listen to a hostname other than the IP and a port alone. For that hostname to work it needs to resolve to any local IP. This has also nothing to do with having networking configured or not, this works without any wifi or cable connected as well.

You are right, the `/etc/hosts` was quite UNIX-centric, I changed it now.

> I do think the quotes around 'test' need to be removed.

They might be misinterpreted on some systems to be part of the hostname. You should not need the quotes here anyways.

JimPanic:
Looking at the original issue, they *are* misinterpreted by Windows 10 to be part of the argument apparently.

FredKSchott:
+1 for removing the quotes

Instead of adding an entirely  new NOTE section can we just rephrase the paragraph above to mention that more configuration is needed by the user? Something like:

>If you've configured a custom hostname on your machine, you can serve from it with the `--hostname` argument (e.g. app project demo is available at `http://test:8080`):

>```
>polymer serve --hostname test`
>```

JimPanic:
FYI from the other original issue in polymer-cli:

> I just tested it on a colleague's Windows machine: this is an issue with `cmd.exe`. PowerShell works the same as any UNIX shell in this regard and strips the single and double quotes. `cmd.exe` strips double quotes but not single quotes.

app/1.0/docs/tools/polymer-cli.md
```diff
@@ -266,10 +266,10 @@ Serve from port 3000:
 
     polymer serve --port 3000
 
-Serve from hostname `test` (e.g. app project demo is available at
-`http://test:8080`):
+If you configured a custom hostname on your machine, polymer can serve it with the
```
FredKSchott: Small nit: `If you have configured`

FredKSchott:
+1 from me with the one small Grammar Nazi fix

app/1.0/docs/tools/polymer-cli.md
JimPanic: Ay!

JimPanic:
@FredKSchott Thanks for catching that!

I asked a colleague as well who is a native speaker unlike me and we ended up both being completely confused by if sentences and their rules. lol.

FredKSchott:
Haha no problem, to be honest I'm not even 100% sure if either is "correct" but I think the change reads more formally and less casually which is a good thing imo.

LGTM, @arthurevans would you do the honors of giving a final look-over and merging?

app/1.0/docs/tools/polymer-cli.md
```diff
@@ -266,10 +266,10 @@ Serve from port 3000:
 
     polymer serve --port 3000
 
-Serve from hostname `test` (e.g. app project demo is available at
-`http://test:8080`):
+If you have configured a custom hostname on your machine, polymer can serve it with the
```
arthurevans: polymer => Polymer CLI

app/1.0/docs/tools/polymer-cli.md
```diff
@@ -266,10 +266,10 @@ Serve from port 3000:
...
+`--hostname` argument (e.g. app project demo is available at `http://test:8080`):
```
arthurevans: "e.g." => "for example,"

arthurevans:
Minor nits to match doc style. Otherwise LGTM.

JimPanic:
There we go, incorporated the changes. 👍

arthurevans:
:dart: Awesome. Thanks!

FredKSchott:
@JimPanic woot! Thanks for working to get this merged!

PR #1706, opened 2016-08-11
Missing comma and end bracket for if condition

mohanaravind:
Added comma, added the end bracket


arthurevans:
LGTM. Thanks for the fix!

PR #1708, opened 2016-08-12
Loosen lifecycle guarantees

mgiuffrida:
created/ready are *not* called on local DOM children if those elements aren't registered yet. Sounds obvious,
but this bit me in a Fiddle where the scripts are loaded after the HTML so order matters.


mgiuffrida:
A more concise version, if you prefer: https://github.com/Polymer/docs/compare/master...mgiuffrida:patch-concise

arthurevans:
Thanks! I might play around with the wording of the last paragraph, but LGTM.

PR #1722, opened 2016-08-23
Maybe demonstrate one-way binding when appropriate

dschuyler:
The example shows a div with two-way binding. Is this an opportunity to show when one-way binding would suffice?

P.S. I may be making an assumption that it's better to use one-way binding; and use two-way binding only when necessary. If that's not accurate: if two-way binding should be used unless one-way binding is needed, that would be helpful to know.

### DIFF
--- a/app/1.0/docs/devguide/data-system.md
+++ b/app/1.0/docs/devguide/data-system.md
@@ -38,7 +38,7 @@ Consider a very simple element:
 ```html
 <dom-module id="name-card">
   <template>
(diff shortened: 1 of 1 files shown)


arthurevans:
Good catch. In this case it doesn't matter whether you use `{{}}` or `[[]]`, because the binding is always going to be one-way; but in recent edits to the data binding chapter, we decided to use the explicit one-way annotations for any bindings that are inherently one-way (compound bindings, computed bindings, negated bindings, and attribute bindings). We should do the same thing here.

PR #1723, opened 2016-08-23
Looks like a typo on emphasis

dschuyler:
With a space before the asterisk, md will print both asterisks rather than put emphasis on the text.


arthurevans:
LGTM. Thanks for the fix!

PR #1732, opened 2016-08-30
noted that $$ is just a shorthand

tjsavage:
@arthurevans WDYT? This is to address https://github.com/Polymer/polymer/pull/3186#issuecomment-243296251


arthurevans:
LGTM.

PR #1736, opened 2016-09-08
Remove unused breaking styles

robdodson:
These styles aren't used anymore because we're using google-youtube on this page and they're actually breaking the page in Firefox. @notwaldorf @arthurevans PTAL


notwaldorf:
Makes sense!

PR #1743, opened 2016-09-18
Added a hint for `bower install`

scaratec:
This tutorial is intended for beginners. Without downloading the dependencies via `bower install` the demo won't work and will cause some frustration.


scaratec:
I signed it!

arthurevans:
The zip file comes with the bower components preinstalled, so that shouldn't be an issue. Did you clone the repo off of github instead?

PR #1749, opened 2016-09-20
Fix table formatting

arthurevans:
Currently, the formatting for this table is messed up. See here: 

https://www.polymer-project.org/1.0/docs/devguide/data-binding#native-binding

This PR should make it less messed up.


arthurevans:
Fixed in #1750.

PR #1764, opened 2016-09-28
Updated docs/app/1.0/start/index.html with shorter paragraphs

katejeffreys:
@arthurevans suggested edits to this page to streamline it.

### DIFF
--- a/app/1.0/start/index.html
+++ b/app/1.0/start/index.html
@@ -120,7 +120,7 @@
           </google-youtube>
       </div>
       <div>
-        <p>Join Rob Dodson from the Chrome Developer Relations team for Polycasts, a comprehensive set of videos exploring the ins and outs of Polymer. Start a new Polymer project from scratch, add functionality to your elements with behaviors, or get a crash course on understanding data binding. Create accessible components, and localize them for your target audience.</p>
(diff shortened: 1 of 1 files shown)


app/1.0/start/index.html
```diff
@@ -125,7 +120,7 @@ <h2 class="underline">Build Apps</h2>
           </google-youtube>
       </div>
       <div>
-        <p>Ready to get your hands dirty and revolutionize the modern web? Join Rob Dodson from the Chrome Developer Relations team as he explores the ins and outs of Polymer, a new library that's pushing the boundaries of what's possible in the browser. Learn about the basic building blocks that make up a Polymer application, and see how to compose those elements into buttery smooth mobile experiences. The future of front end development is all here, on the Polycasts playlist!</p>
+        <p>Join Rob Dodson from the Chrome Developer Relations team for Polycasts, a comprehensive set of videos exploring the ins and outs of Polymer. Start a new Polymer project from scratch, add functionality to your elements with behaviors, or get a crash course on understanding data binding. Create accessible components, and localize them for your target audience.</p>
```
arthurevans: Going to suggest we cut the words "with behaviors" because they are going away in 2.0 and that way we can keep the wording the same. Resulting sentence is a little vague, but still OK I think...

PR #1798, opened 2016-10-11
Revise based on Fred's feedback

katejeffreys:


FredKSchott:
It seems like we have two different styles across our section headers, `DESCRIPTION` and `**TOOL NAME**: DESCRIPTION`. Also some are `h2` and some are `h3`. 

Can we change them all to match `**TOOL NAME**: DESCRIPTION` so that the tool name is included? This would be especially helpful for the table of contents.

FredKSchott:
Great! LGTM!

PR #1804, opened 2016-10-17
2.0 preview minimal docs

arthurevans:


app/2.0/docs/tools/advanced.md
```diff
@@ -74,79 +74,8 @@ Source: [github.com/Polymer/web-component-tester](https://github.com/Polymer/web
...
-    npm install -g polybuild
-
-Usage:
-
-    polybuild index.html --maximum-crush
-
-Source: [github.com/PolymerLabs/polybuild](https://github.com/PolymerLabs/polybuild)
+You can use `polymer-build` from the command line if you need to customize your build process.
 
+<a href="https://github.com/Polymer/polymer-build/blog/master/README.md" target="_blank">See the polymer-build README for instructions</a>.
```
addyosmani: This is a typo. I think you meant to link to https://github.com/Polymer/polymer-build/blob/master/README.md. 

Side: I worry that we're losing a lot of insight into what happens behind the scenes when using this module with the new text. If there's time I might mention that it's useful for bundling files, extracting inline resources (CSS/JS), generating Service Workers and so on.

app/2.0/docs/about_20.md
```diff
@@ -0,0 +1,438 @@
...
+
+Could become:
+
+
+```
+<my-anchor>
+  <a>...</a>
+</my-anchor>
+```
+
```
addyosmani: If I'm reading this correctly, we're encouraging use of falling back to the LightDOM for anything that needs to support progressive enhancement while moving away from the idea of type-extensions becoming a thing you can rely on. Is that right?

arthurevans: I think the answer to #2 is "yes" (we're backing off from pushing type-extension elements in the short term). I don't think it has anything to do with progressive enhancement. 

I guess the text here could be a little more nuanced. For something like `<dom-bind>`, the whole point is taking a user-supplied template... Hence, taking a template supplied in light DOM makes sense.

A custom input _could_ just as well use an input element in shadow DOM (which is what `<paper-input>` used to do, and probably still does). However, there are a *lot* of attributes you can tweak on an input element... So rather than trying to proxy all of those through to your custom element API, it's easier to let the user set them directly, which the old `iron-input` did by virtue of being a type-extension element, and the new `iron-input` does by taking an `<input>` in light DOM.

arthurevans:
@katejeffreys Can you test this branch locally to make sure it's working?

I'd like to merge this into master (it should be the same as what we have on the site right now).

Issue #1810, opened 2016-10-25
Tiny markdown issue on docs/app/1.0/blog/routing.md


emazurek:
In **`<app-route>`** section, an `<app-route>` tag is missing its code formatting, rendering it invisible!

katejeffreys:
Thanks @emazurek

PR #1811, opened 2016-10-25
Fix markdown typo

emazurek:
Format `<app-route>` as code in a place where it was missing, making it visible; addresses #1810


emazurek:
I signed it!

PR #1812, opened 2016-10-27
fix hex color code

alirni:
remove #


arthurevans:
LGTM. Thanks for the fix!

PR #1815, opened 2016-11-03
Update firebase.json rewrite example. Fixes #1755

### LINKED ISSUE #1755: firebase.json rewrite property is wrong for polymer
Here https://www.polymer-project.org/1.0/start/toolbox/deploy
in the docs it says the correct way to configure firebase.json to host from firebase:

```
"rewrites": [ {
  "source": "**/!{*.*}",
  "destination": "/index.html"
} ]
```

but it does not work, and you will see 404 for any route other than root. Correct way is:

```
"rewrites": [ {
  "source": "**/!(*.*)",
  "destination": "/index.html"
} ]
```

robdodson:
@arthurevans PTAL

### DIFF
--- a/app/1.0/start/toolbox/deploy.md
+++ b/app/1.0/start/toolbox/deploy.md
@@ -139,7 +139,7 @@ guide](https://www.firebase.com/docs/hosting/quickstart.html).
 
     ```
     "rewrites": [ {
-      "source": "**/!{*.*}",
+      "source": "**/!(*.js|*.html|*.css|*.json|*.svg|*.png|*.jpg|*.jpeg)",
       "destination": "/index.html"
     } ]
     ```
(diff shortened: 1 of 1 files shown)


abdonrd:
@robdodson maybe add the `__` for the Firebase Auth?

robdodson:
good call

robdodson:
ack yeah I'm just not sure if the `__` uses the same syntax...

robdodson:
ok I decided to just do 2 rewrites just to be safe. @mbleigh does this change look ok to you?

robdodson:
@arthurevans I think we might be good to merge

abdonrd:
Hum... We should copy/paste to [Polymer/docs/app/2.0/start/toolbox/deploy.md](https://github.com/Polymer/docs/blob/master/app/2.0/start/toolbox/deploy.md)?

robdodson:
yuuup

PR #1818, opened 2016-11-05
Fix possible small typo in v1 and v2 docs

petrpacas:
This could also be a misunderstanding on my part, so please check

### DIFF
--- a/app/1.0/docs/devguide/quick-tour.md
+++ b/app/1.0/docs/devguide/quick-tour.md
@@ -126,7 +126,7 @@ Learn more: data binding</a></p>
 
 ### Declare a property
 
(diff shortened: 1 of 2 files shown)


petrpacas:
I signed it!

arthurevans:
Thanks for the PR! Our style is not to use possessives with product names, so the decision to avoid "Polymer's" is a conscious one.

petrpacas:
Ok, roger that.
(You are using "Polymer's" troughout the docs tho)

arthurevans:
Yep, you're right, we have more than I thought. O, for an editor!
Will try to clean some of those up!

PR #1820, opened 2016-11-15
Added set method to list of polymer mutator functions

tpanetti:
Looks like the docs forgot the set method. This might seem superfluous but it confused me for a little while.


tpanetti:
Signed

katejeffreys:
Thanks @tpanetti and @stramel, hoping you can explain the PR to me a bit more. I am not sure whether `set` was in fact forgotten, or if it was left out deliberately. Hopefully @arthurevans can also state which way around it is.

The docs state:

"When modifying arrays, a set of array mutation methods are provided on Polymer element prototypes which mimic Array.prototype methods..."

I note that `push`, `pop`, `shift`, `unshift` and `splice` are `Array.prototype` methods, whereas `set` is not ([see docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/prototype)), so it might have been left out deliberately.

tpanetti:
@katejeffreys Looks like I missed the bit about mimicking Array.prototype's methods. That would make this addition inappropriate in the context.

PR #1824, opened 2016-11-21
Fix section link (data-binding)

ifraixedes:
Fix link to compound-bindings section in `data-binding.md` document


arthurevans:
Thanks for the fix!

PR #1826, opened 2016-11-23
Add note about :host-context()

### LINKED ISSUE #1813: Add note about :host-context() in 2.0
See:

https://github.com/Polymer/polymer/tree/2.0-preview#scoped-styling

Appears to be missing from 2.0 preview docs.

katejeffreys:
Fixes #1813


katejeffreys:
@arthurevans Going through old issues, I found this... would you mind sanity-checking if you get the chance?

arthurevans:
Sorry for the delayed review. I think this is a practical detail which would be more appropriate in the upgrade guide, rather than a high-level item for the "About" doc. I also had some concerns about recommending `:dir` because it turns out that it's not supported anywhere, and I don't think we can polyfill it in a particularly satisfying fashion. 

So, I'm not really sure what to say, except that we can't use it.

PR #1833, opened 2016-11-29
fix firebase hosting glob notation

geeksbaek:
I think this glob notation is wrong. It does not work for me. 
Maybe an old version?


geeksbaek:
I signed it!

abdonrd:
Reference to: https://github.com/Polymer/docs/pull/1815

arthurevans:
Thanks for the fix! This was fixed by a prior PR (#1815) that had been hanging around unmerged for way too long. Apologies!

PR #1843, opened 2016-12-13
Data system & properties updates.

arthurevans:
R: @katejeffreys 

Please sanity check.

### DIFF
--- a/app/2.0/docs/devguide/model-data.md
+++ b/app/2.0/docs/devguide/model-data.md
@@ -161,25 +161,11 @@ Example { .caption }
 </dom-module>
 ```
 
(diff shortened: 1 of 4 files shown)


app/2.0/docs/devguide/model-data.md
```diff
@@ -184,7 +164,7 @@ Example: { .caption }
 ### Notify Polymer of array mutations {#notifysplices}
```
katejeffreys: >When it's not convenient to use the Polymer [array mutation methods](#array-mutation).

Is this is meant to be a continuation of the heading? I found it confusing.

arthurevans:
Addressed comments... Going to go ahead and merge.

PR #1852, opened 2016-12-31
[BUGFIX] Use _boundHandler in slotchange example

daKmoR:
Furthermore

"You can use a `slotchange` event listener to react to runtime changes to distribution, but the event
listener doesn't fire for the *initial* distribution."

In Firefox it also fires for the *initial* distribution so it will be called two times using the above example...

### DIFF
--- a/app/2.0/docs/upgrade.md
+++ b/app/2.0/docs/upgrade.md
@@ -704,9 +704,14 @@ listener doesn't fire for the *initial* distribution.
 
 ```
 ready: function() {
+  // super.ready(); // if in a 2.0 only system
   var this._boundHandler = this._processLightChildren.bind(this);
-  setTimeout(this._processLightChildren);
+  setTimeout(this._boundHandler);
   this.$.slot.addEventListener('slotchange', this._processLightChildren);
+}
+
+_processLightChildren: function() {
+  console.log(this.$.slotmobileimage.assignedNodes());
 }
 ```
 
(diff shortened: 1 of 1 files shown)


daKmoR:
I signed it!

app/2.0/docs/upgrade.md
```diff
@@ -704,10 +704,15 @@ listener doesn't fire for the *initial* distribution.
 
 ```
 ready: function() {
+  // super.ready(); // if in a 2.0 only system
```
arthurevans: Should be "for 2.0 class-based elements only." (It's not required for hybrid elements.)

app/2.0/docs/upgrade.md
```diff
@@ -704,10 +704,15 @@ listener doesn't fire for the *initial* distribution.
...
   var this._boundHandler = this._processLightChildren.bind(this);
-  setTimeout(this._processLightChildren);
+  setTimeout(this._boundHandler);
   this.$.slot.addEventListener('slotchange', this._processLightChildren);
```
arthurevans: Need `_boundHandler` here, too.

app/2.0/docs/upgrade.md
```diff
@@ -704,10 +704,15 @@ listener doesn't fire for the *initial* distribution.
...
 }
+
+_processLightChildren: function() {
+  console.log(this.$.slotmobileimage.assignedNodes());
```
arthurevans: I think this should just be `this.$.slot`?

arthurevans:
Thanks!

PR #1857, opened 2017-01-24
Temporarily disable push because of app engine issue.

arthurevans:


arthurevans:
R: @ebidel 
Already deployed to fix #4264. Hoping to roll these push setting back out after we track down (what appears to be) an app engine issue.

arthurevans:
This is deployed, so I'm going to go ahead and merge it so we don't accidentally break ourselves by redeploying push before we're ready.

PR #1859, opened 2017-01-30
Delete bad tools/index.html

FredKSchott:
This file is not linked to anywhere, but contains bad lorem ipsum and is still accessible via URL: https://www.polymer-project.org/1.0/docs/tools/


arthurevans:
LGTM. Did you mean to set this up as a PR against FredKSchott-patch-3?

PR #1861, opened 2017-02-06
Fix lastresponse to last-response

gndplayground:
There is a typo mistake here lastresponse 
https://www.polymer-project.org/1.0/docs/api/dom-bind.

For newbies like me, it was confusing.


gndplayground:
I signed it!

arthurevans:
Thanks for the PR! This page is actually autogenerated doc, so we can't fix it here. The source is here: 

https://github.com/Polymer/polymer/blob/master/src/lib/template/dom-bind.html#L37

Would you be willing to make the PR over there?

Thanks,
Arthur

PR #1864, opened 2017-02-09
Bring 2.0 branch up to date with master

arthurevans:
Also updates navigation.

Do not merge: needs to be reconciled with @notwaldorf's version switcher branch.


arthurevans:
I think barring any objections from @notwaldorf, @katejeffreys should go ahead and test this PR out. Puts the full 2.0 nav back in place so it's easier to test docs in situ.

PR #1866, opened 2017-02-10
1.8.0 doc updates.

arthurevans:
R: @katejeffreys @azakus

### DIFF
--- a/app/1.0/docs/release-notes.md
+++ b/app/1.0/docs/release-notes.md
@@ -106,7 +106,7 @@ Note this feature is implemented as an attribute API only.  There is no correspo
 </template>
 ```
 
-### Template strip-whitespace is recursive
+### Template `strip-whitespace` is recursive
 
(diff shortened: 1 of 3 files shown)


app/1.0/docs/release-notes.md
```diff
@@ -11,6 +11,108 @@ title: Release notes
...
+
+  ...
+
+  <!-- upgrade my-element when loggedIn == true -->
+  <my-element disable-upgrade$="{{!loggedIn}}"></my-element>
+
+</template>
+```
+
+### Template strip-whitespace is recursive
```
katejeffreys: Backticks around `strip-whitespace`?

arthurevans: Good catch... I was going to say that I prefer not to use code font in headings, but in practice I've been wildly inconsistent in this area. There's one other heading in this section that includes code font, so I'll change it here for consistency...

We should loop back to this and add a note to the STYLE.md doc.

PR #1865, opened 2017-02-10
Toolsdocs

katejeffreys:

PR #1871, opened 2017-02-14
Rebased tools docs changes to 2.0

katejeffreys:


abdonrd:
@katejeffreys remember https://github.com/Polymer/docs/pull/1865#pullrequestreview-21201799 :)

app/2.0/docs/tools/glossary.md
```diff
@@ -0,0 +1,162 @@
...
+## crisper
+crisper extracts inline scripts from an HTML file. This is useful if you need to run a transformation on your JavaScript, such as a minifier or a transpiler. You could also use crisper to separate JavaScript from HTML in order to deploy in an environment that uses CSP (Content Security Policy).
+
+# D
+## dependency
+In the context of Polymer apps, a dependency is a resource (a file or set of files) required by your application, such as HTML files or JavaScript libraries. Dependencies should be declared in [`polymer.json`](#polymer-json) to ensure they are included in your build. See also: [dynamic dependency](#dynamic-dependency), [static dependency](#static-dependency), [transitive dependency](#transitive-dependency).
+
+In the context of the Polymer data system, a dependency is a piece of data that an observer, computed property or computed binding depends on.
+
+## dependency variant
```
FredKSchott: - `dependency variants` or just `variants` would be a more accurate title, since we think about these in terms of adding collections of variants. 
- variants are useful for element projects only at the moment. They are useful for testing elements against Polymer 1.x & Polymer 2.x (where bower.json would normally only let you install and test against one of these). I would change any references to your "app" here to reference your "element" instead.

app/2.0/docs/tools/glossary.md
```diff
@@ -0,0 +1,162 @@
...
+# E
+## eager loading
+Eager loading refers to the process of loading dependencies statically, rather than on demand. Compare [lazy loading](#lazy-loading).
+
+## entry point
+The HTML actually loaded by the browser to start the application. Usually `index.html`. In an application built with the [app shell](#app-shell) architecture, the entrypoint is loaded for many different URLs.
+
+# F
+## fragment
+A lazily-loaded HTML file. Fragments are used to define bundles and are often associated with routes or non-critical resources.
```
FredKSchott: - A Fragment is much more about defining a bundle than it is about a lazy-loaded HTML file. I would reorder these two sentances so that we don't bury the lead here.

app/2.0/docs/tools/glossary.md
```diff
@@ -0,0 +1,162 @@
...
+## bower.json
+`bower.json` is a manifest used by bower for configuring packages for use as dependencies. See [the `bower.json` specification](https://github.com/bower/spec/blob/master/json.md) for more information.
+
+# C
+## crisper
+crisper extracts inline scripts from an HTML file. This is useful if you need to run a transformation on your JavaScript, such as a minifier or a transpiler. You could also use crisper to separate JavaScript from HTML in order to deploy in an environment that uses CSP (Content Security Policy).
+
+# D
+## dependency
+In the context of Polymer apps, a dependency is a resource (a file or set of files) required by your application, such as HTML files or JavaScript libraries. Dependencies should be declared in [`polymer.json`](#polymer-json) to ensure they are included in your build. See also: [dynamic dependency](#dynamic-dependency), [static dependency](#static-dependency), [transitive dependency](#transitive-dependency).
```
rictic: > Dependencies should be declared in [`polymer.json`](#polymer-json) to ensure they are included in your build.

This seems either wrong or only true in a very specific context. @justinfagnani @FredKSchott

app/2.0/docs/tools/glossary.md
```diff
@@ -0,0 +1,162 @@
...
+
+## npm
+Package management system for [Node.js](#node-js).
+
+# P
+## package.json
+A metadata file used by [npm](#npm) to manage project dependencies.
+
+## polymer-build
+`polymer-build` is a library for building Polymer projects. `polymer-build` allows you to completely customize your build and combine additional streams and build tasks in any order.
```
FredKSchott: - we get a lot of questions about polymer-build vs the CLI build command. A quick sentence about how they're different here would be helpful.

app/2.0/docs/tools/glossary.md
```diff
@@ -0,0 +1,162 @@
...
+
+## dynamic dependency
+A dynamic dependency is a dependency that is imported at runtime, using a mechanism such as `importHref`. Compare [static dependency](#static-dependency).
+
+# E
+## eager loading
+Eager loading refers to the process of loading dependencies statically, rather than on demand. Compare [lazy loading](#lazy-loading).
+
+## entry point
+The HTML actually loaded by the browser to start the application. Usually `index.html`. In an application built with the [app shell](#app-shell) architecture, the entrypoint is loaded for many different URLs.
```
rictic: Consider "an app shell architecture" rather than "the app shell architecture"

app/2.0/docs/tools/glossary.md
```diff
@@ -0,0 +1,162 @@
...
+## package.json
+A metadata file used by [npm](#npm) to manage project dependencies.
+
+## polymer-build
+`polymer-build` is a library for building Polymer projects. `polymer-build` allows you to completely customize your build and combine additional streams and build tasks in any order.
+
+## polyclean
+A set of basic gulp plugins for cleaning HTML. Includes code to remove comments and whitespace.
+
+## polylint
```
FredKSchott: - if these are 2.0 docs, lets replace polylint references with `polymer-linter` references. 
- @rictic could best help you if the description needs to be changed at all for this

app/2.0/docs/tools/glossary.md
```diff
@@ -0,0 +1,162 @@
...
+# F
+## fragment
+A lazily-loaded HTML file. Fragments are used to define bundles and are often associated with routes or non-critical resources.
+
+# G
+## git
+A distributed version control system. bower depends on git to install components from their GitHub repositories.
+
+## grunt
+A JavaScript task automation tool that can be used to set up a custom build process using the [`polymer-build` library](#polymer-build).
```
rictic: We support grunt with polymer-build? @FredKSchott

app/2.0/docs/tools/glossary.md
```diff
@@ -0,0 +1,162 @@
...
+## polymer-build
+`polymer-build` is a library for building Polymer projects. `polymer-build` allows you to completely customize your build and combine additional streams and build tasks in any order.
+
+## polyclean
+A set of basic gulp plugins for cleaning HTML. Includes code to remove comments and whitespace.
+
+## polylint
+[Linter](#linter) invoked by `polymer lint` command. Lints a Polymer project.
+
+## polymer build
```
FredKSchott: for all of these commands, would they be best organized under the CLI entry? Reading this I get the impression that these are all separate tools / am left wondering "commands for what?".

app/2.0/docs/tools/glossary.md
```diff
@@ -0,0 +1,162 @@
...
+
+When a service worker is installed, apps may be able to be used offline in a meaningful way.
+
+## shell
+In the [app-shell](#app-shell) design pattern, the shell is the main code that boots the app and loads necessary resources. The shell usually includes the common UI for the app, and the [router](#routing), which loads resources needed for the current route.
+
+## static dependency
+A [dependency](#dependency) that is required by a given file, and whose import is hard-coded using `<link rel=”import”>` or `<script src=”...”>`. This dependency will always be loaded when the given file is loaded.
+	
+## `tattoo`
```
FredKSchott: - @usergenic Is tattoo a public tool at this point / worth describing here?

app/2.0/docs/tools/glossary.md
```diff
@@ -0,0 +1,162 @@
...
+This tool has been renamed to [Polymer Analyzer](#polymer-analyzer). It is a static analysis framework for Web Components. 
+
+# L
+## lazy loading
+Lazy loading takes advantage of routing to load elements only when they are required for render. Used correctly, lazy loading can improve performance.
+
+For an example implementation of lazily-loaded elements, see the [Shop app](https://github.com/Polymer/shop). In the Shop app, views are elements which are instantiated only when they are routed to. 
+
+## linter
+A linter is a piece of software that analyzes code for suspicious-looking constructions that can potentially cause errors; for example, division by zero or data that will go out of bounds. The Polymer tools include [`polylint`](#polylint), which does this for Polymer apps and elements.
```
rictic: We can give some web components specific examples. Maybe:

> for example, registering an element without specifying its tag name, or using invalid data-binding syntax.

app/2.0/docs/tools/polymer-json.md
```diff
@@ -0,0 +1,171 @@
...
+  "entrypoint": "index.html",
+  "shell": "src/shop-app.html",
+  "fragments": [
+    "src/shop-list.html",
+    "src/shop-detail.html",
+    "src/shop-cart.html",
+    "src/shop-checkout.html",
+    "src/lazy-resources.html"
+  ],
+  "sourceGlobs": [
```
FredKSchott: "sourceGlobs" -> "sources"

app/2.0/docs/tools/polymer-json.md
```diff
@@ -0,0 +1,171 @@
...
+
+Dependencies that the analyzer component of the Polymer build toolchain can’t discover, and that do not need to be bundled.
+
+### shell
+*Required* 
+Type: `String`
+
+The app shell. Contains the main code that boots the app and loads necessary resources. The shell usually includes the common UI for the app, and the router, which loads resources needed for the current route.
+
+### sourceGlobs
```
FredKSchott: "sourceGlobs" -> "sources"

app/2.0/docs/tools/polymer-json.md
```diff
@@ -0,0 +1,171 @@
...
+  |-service-worker.js
+  |-src
+  |-sw-precache-config.js
+  |-test
+```
+
+The `sourceGlobs` property is set as follows:
+
+```json
+"sourceGlobs": [
```
FredKSchott: "sourceGlobs" -> "sources"

app/2.0/docs/tools/polymer-json.md
```diff
@@ -0,0 +1,171 @@
...
+    "css": {"minify": true},
+    "html": {"minify": true}
+  }]
+```
+
+And here is a configuration to generate two optimized builds: One bundled and one unbundled:
+
+```json
+"builds": [{
+    “name”: “bundled”,
```
FredKSchott: bad quotation mark symbols got copied into here, will break other users who copy-paste this

app/2.0/docs/tools/glossary.md
```diff
@@ -0,0 +1,162 @@
...
+
+# M
+## monolithic
+A monolithic app is one that has been implemented without a modular design pattern.
```
rictic: `modular design pattern` is somewhat vague. What about something like:

> A monolithic app is one that loads all of its code statically and does not use an app shell architecture or dynamic code loading.

app/2.0/docs/tools/polymer-json.md
```diff
@@ -0,0 +1,171 @@
...
+    "bundle": true,
+    "js": {"minify": true},
+    "css": {"minify": true},
+    "html": {"minify": true}
+  },{
+    “name”: “unbundled”,
```
FredKSchott: bad quotation mark symbols got copied into here, will break other users who copy-paste this

app/2.0/docs/tools/glossary.md
```diff
@@ -0,0 +1,162 @@
...
+## polymer-cli
+A command line tool for Polymer projects. `polymer-cli` has a number of tools for working with Polymer and Web Components.
+
+## polymer-linter
+`polymer-linter` is in development. It will be a tool for [linting](#linter) Polymer applications, and will replace [`polylint`](#polylint). For a working linter, use `polylint`.
+
+## `polymer.json`
+`polymer.json` is the Polymer build configuration file. It lets you configure your app’s entrypoint and app shell files, instead of passing these as parameters to polymer build.
+
+Polymer.json also lets you specify the locations of source files, HTML fragments and additional dependencies.
```
rictic: Recommend always writing it as `polymer.json`

rictic: It also lets you configure the new linter!

app/2.0/docs/tools/glossary.md
```diff
@@ -0,0 +1,162 @@
...
+## service worker
+A script that a web browser runs in the background. On first accessing a particular site, the service worker fetches a defined set of resources and caches them. Then, the service worker acts as a client-side proxy. It intercepts network requests for resources and, where possible, supplies the resources from the browser cache. 
+
+When a service worker is installed, apps may be able to be used offline in a meaningful way.
+
+## shell
+In the [app-shell](#app-shell) design pattern, the shell is the main code that boots the app and loads necessary resources. The shell usually includes the common UI for the app, and the [router](#routing), which loads resources needed for the current route.
+
+## static dependency
+A [dependency](#dependency) that is required by a given file, and whose import is hard-coded using `<link rel=”import”>` or `<script src=”...”>`. This dependency will always be loaded when the given file is loaded.
```
rictic: Consider rephrasing to note that these are just two examples of how a static dependency can be written.

e.g. `` is also a static dependency

app/2.0/docs/tools/glossary.md
```diff
@@ -0,0 +1,162 @@
...
+
+## transitive dependencies
+The entire graph of [dependencies](#dependency) for a given file. A file’s transitive dependencies are all of that file’s dependencies, plus their dependencies, and so on. 
+
+# V
+## variant
+See [dependency variant](#dependency-variant).
+
+## vulcanize
+Polymer build tool that follows HTML Imports and `<script>` tags to concatenate resources into a single page for production. Can improve performance by reducing network trips.
```
rictic: This tool is being reimplemented and renamed as `polymer-bundler`.

app/2.0/docs/tools/polymer-json.md
```diff
@@ -0,0 +1,171 @@
...
+  "sourceGlobs": [
+   "src/**/*",
+   "data/**/*",
+   "images/**/*",
+   "bower.json"
+  ],
+  "extraDependencies": [
+    "manifest.json",
+    "bower_components/webcomponentsjs/webcomponents-lite.min.js"
+  ]
```
rictic: Consider adding this to the end of the example polymer.json, as the new linter will need it, and it won't cause problems with the existing linter:

```json
],
"lint": {
  "rules": ["polymer-2-hybrid"]
}
```

app/2.0/docs/tools/polymer-json.md
rictic: These look like smart quotes which will throw off any json parsers.

app/2.0/docs/tools/polymer-json.md
rictic: ditto

app/2.0/docs/tools/glossary.md
katejeffreys: Removing it until further notice

app/2.0/docs/tools/glossary.md
katejeffreys: Pls elaborate?

app/2.0/docs/tools/glossary.md
katejeffreys: Removing until further notice

app/2.0/docs/tools/glossary.md
rictic: Sure, sorry for super vague comment. We talk about dependencies in a lot of ways, even in the context of polymer apps. I think we might want to use a more specific term here than dependency, because you don't have to specify all of your dependencies in your polymer.json. I'm not sure what you do need to specify though, so I'm not sure what a better term would be. Justin, Fred, any thoughts?

katejeffreys:
Thanks! Removing the line in question until further notice so at least it
doesn't say something misleading.

Please let me know if this PR needs more changes.

justinfagnani:
@katejeffreys I'm doing some review now...

app/2.0/docs/tools/glossary.md
```diff
@@ -0,0 +1,160 @@
...
+
+When a service worker is installed, apps may be able to be used offline in a meaningful way.
+
+## shell
+In the [app-shell](#app-shell) design pattern, the shell is the main code that boots the app and loads necessary resources. The shell usually includes the common UI for the app, and the [router](#routing), which loads resources needed for the current route.
+
+## static dependency
+A [dependency](#dependency) that is required by a given file, and whose import is hard-coded using syntax like `<link rel="import">` or `<script src="...">`. This dependency will always be loaded when the given file is loaded.
+
+## transitive dependencies
```
justinfagnani: Maybe make singular and say something like:

A transitive dependency is in indirect dependency, one that is depended on by a direct dependency or other indirect dependency.

The transitive dependencies of a resource are all dependencies, direct and indirect, that a resource depends on.

app/2.0/docs/tools/glossary.md
```diff
@@ -0,0 +1,160 @@
...
+The entire graph of [dependencies](#dependency) for a given file. A file’s transitive dependencies are all of that file’s dependencies, plus their dependencies, and so on. 
+
+# V
+## variant
+See [dependency variant](#dependency-variant).
+
+## vulcanize
+This tool has been re-named [polymer-bundler](#polymer-bundler).
```
justinfagnani: renamed -> replace by

app/2.0/docs/tools/polymer-json.md
```diff
@@ -0,0 +1,174 @@
...
+title: "`polymer.json` specification"
+---
+
+<!-- toc -->
+
+# About `polymer.json`
+
+Creating a `polymer.json` in your project directory allows you to store information about your project structure and desired build configuration(s). It is used by the [Polymer CLI](sdffsdkjhk) as a way to understand the structure of your application.
+
+To make sure your application builds correctly, create a `polymer.json` file at the top level of your project:
```
justinfagnani: mention that this structure is from Shop?

app/2.0/docs/tools/polymer-json.md
```diff
@@ -0,0 +1,174 @@
+---
+title: "`polymer.json` specification"
+---
+
+<!-- toc -->
+
+# About `polymer.json`
+
+Creating a `polymer.json` in your project directory allows you to store information about your project structure and desired build configuration(s). It is used by the [Polymer CLI](sdffsdkjhk) as a way to understand the structure of your application.
```
justinfagnani: `sdffsdkjhk`? :)

app/2.0/docs/tools/polymer-json.md
```diff
@@ -0,0 +1,174 @@
...
+
+If a fragment has static dependencies, provided the fragment is defined in this property, the Polymer build analyzer will find them. You only need to list the file imported by `importHref`.
+
+In a Polymer app, the files listed in the `fragments` property usually contain one or more element definitions that may or may not be required during the user’s interaction with the app, and can thus be lazily loaded.
+
+### extraDependencies
+Optional
+Type: `Array` of `String`
+
+Dependencies that the analyzer component of the Polymer build toolchain can’t discover, and that do not need to be bundled.
```
justinfagnani: can't discover, possibly because they're not statically imported,...

app/2.0/docs/tools/polymer-json.md
```diff
@@ -0,0 +1,174 @@
...
+}
+```
+
+## Properties
+
+### entrypoint
+Optional, Defaults to `index.html`
+Type: `String`
+
+The main entrypoint to your app for all routes, often `index.html`. This file should import the app shell file specified in the shell property. It should be as small as possible since it’s loaded and cached for all routes.
```
justinfagnani: It should be as small as possible because it's _not_ cached very well, since it's served for many different URLs.

Also, because it's served from many URLs, all paths should be absolute.

app/2.0/docs/tools/polymer-json.md
```diff
@@ -0,0 +1,174 @@
...
+### sources
+Optional, Defaults to `["src/**/*"]`
+Type: `Array` of `String`
+
+An optional array of globs matching your application source files. If left out, defaults to all files in your project `src/` directory. You’ll need to set this if you store your source files in other locations.
+
+In the Shop app, source files are stored in `/src`, `/data` and `/images`:
+
+```
+-shop
```
justinfagnani: can you just refer to the Shop layout above?

app/2.0/docs/tools/polymer-json.md
```diff
@@ -0,0 +1,174 @@
...
+* `swPrecacheConfig`: An optional configuration file for the generated service worker.
+* `insertPrefetchLinks`: If `true`, insert prefetch link elements into your fragments so that all dependencies are prefetched immediately.
+* `bundle`: If `true`, bundle your application.
+* `html`:
+** `minify`: If `true`, minify all HTML.
+* `css`:
+** `minify`: If `true`, minify all CSS.
+* `js`:
+** `minify`: If `true`, minify all JS.
+** `compile`: If `true`, use babel to compile all ES6 JS down to ES5.
```
justinfagnani: We should document somewhere the details of the compilation. It's everything in the es2015 preset except modules. It's probably most important that we state that modules aren't supported.

It'd also be good to state the goal of compilation somewhere - to allow code that works in the evergreen browsers to work in older browsers. It's not to compile future language features down to current engines, and we want uncompiled code to work natively in evergreen browsers. This is a bit different than how people usually use compilation.

app/2.0/docs/tools/glossary.md
```diff
@@ -0,0 +1,160 @@
...
+# M
+## monolithic
+A monolithic app is one that loads all of its code statically and does not use an app shell architecture or dynamic code loading.
+
+# N
+## Node.js
+An event-driven JavaScript runtime environment. The [Polymer CLI](#polymer-cli) and [bower](#bower) are both built on Node.
+
+## npm
+Package management system for [Node.js](#node-js).
```
katejeffreys: I think it is okay to have them. I have linked out to nodejs.org for consistency with other entries (e.g the entry on Bower links out to bower.io).

app/2.0/docs/tools/glossary.md
```diff
@@ -0,0 +1,160 @@
...
+
+# P
+## package.json
+A metadata file used by [npm](#npm) to manage project dependencies.
+
+## polyclean
```
katejeffreys: Removed it.

app/2.0/docs/tools/glossary.md
```diff
@@ -0,0 +1,160 @@
...
+## `polyserve`
+`polyserve` is a locally-installable web server for developing [bower](#bower) components. 
+
+# R
+## routing
+Routing is the process an app uses to determine how to respond to a client request for a particular endpoint. A default route is the endpoint the app directs the user to when no route is specified. In a PWA, the [app shell](#app-shell) performs routing.
+
+# S
+## service worker
+A script that a web browser runs in the background. On first accessing a particular site, the service worker fetches a defined set of resources and caches them. Then, the service worker acts as a client-side proxy. It intercepts network requests for resources and, where possible, supplies the resources from the browser cache.
```
katejeffreys: Thx, I've updated to use this definition. I think the stuff about helping to make apps usable offline is important so I have left it in there as an example. Lmk if it doesn't make sense.

app/2.0/docs/tools/glossary.md
```diff
@@ -0,0 +1,160 @@
...
+Routing is the process an app uses to determine how to respond to a client request for a particular endpoint. A default route is the endpoint the app directs the user to when no route is specified. In a PWA, the [app shell](#app-shell) performs routing.
+
+# S
+## service worker
+A script that a web browser runs in the background. On first accessing a particular site, the service worker fetches a defined set of resources and caches them. Then, the service worker acts as a client-side proxy. It intercepts network requests for resources and, where possible, supplies the resources from the browser cache. 
+
+When a service worker is installed, apps may be able to be used offline in a meaningful way.
+
+## shell
+In the [app-shell](#app-shell) design pattern, the shell is the main code that boots the app and loads necessary resources. The shell usually includes the common UI for the app, and the [router](#routing), which loads resources needed for the current route.
```
katejeffreys: Removed it, tx

app/2.0/docs/tools/polymer-json.md
katejeffreys: Your point being?

;)

app/2.0/docs/tools/polymer-json.md
katejeffreys: This sounds like a good idea, but I don't think it belongs in the polymer.json spec. A separate page on compilation might be appropriate? I'm open to your input though.

katejeffreys:
@justinfagnani @FredKSchott @rictic @abdonrd Thank you so much for the reviews.

app/2.0/docs/tools/polymer-json.md
```diff
@@ -0,0 +1,177 @@
+---
+title: "`polymer.json` specification"
+---
+
+<!-- toc -->
+
+# About `polymer.json`
```
arthurevans: Like the other doc, start with H2s. Also, don't really need a header right after the TOC, I think? We usually just start with introductory text.

app/2.0/docs/tools/polymer-json.md
```diff
@@ -0,0 +1,177 @@
...
+    "rules": ["polymer-2-hybrid"]
+  }
+}
+```
+
+## Properties
+
+### entrypoint
+Optional, Defaults to `index.html`
+Type: `String`
```
arthurevans: These two lines are going to run together, so you should probably either end the first one with a period, or format them as a bulleted list or something so the Type: `String` doesn't get wrapped to the end of the first line.

Alternately, just add a `<br>` after the first line...

app/2.0/docs/tools/polymer-json.md
```diff
@@ -0,0 +1,177 @@
...
+
+The main entrypoint to your app for all routes, often `index.html`. This file should import the app 
+shell file specified in the shell property. It should be as small as possible since it’s served for 
+all routes. All paths in the entrypoint should be absolute, because this file is served from many 
+different URLs.
+
+### fragments
+Optional
```
arthurevans: As above (applies to the rest of these lines, below, too.)

arthurevans:
Note that there are a couple of pending comments from me way up the page there ^^

katejeffreys:
@arthurevans I can't see your comments.

app/2.0/docs/tools/glossary.md
```diff
@@ -0,0 +1,162 @@
+---
+title: "Polymer Tools Glossary"
+---
```
katejeffreys: Ok - in that case, should I move it to the /2.0/docs/ level?

PR #1873, opened 2017-02-14
News release blog post

tjsavage:


app/1.0/blog.yaml
```diff
@@ -1,7 +1,15 @@
...
+  author: polymer-team
+  description: "The new Polymer News demo progressive web app helps you start building a cutting-edge PWA using Polymer."
+  tag:
+    - blog
+    - announcements
 - title: "webcomponents.org"
-  path: /1.0/blog/2017-01-09-webcomponents-org
-  published: 2017-01-09
+  path: /1.0/blog/2017-01-10-webcomponents-org
+  published: 2017-01-10
```
arthurevans: Since this is already published with the -09- URL, I don't think we should move it at this point. (We can update the publication date in the YAML, but the path should stay the same.)

app/1.0/blog/2017-02-06-news-pwa.md
```diff
@@ -0,0 +1,21 @@
...
+
+---
+
+We're excited to release a new progressive web app demo and template built with Polymer - [Polymer News](https://news.polymer-project.org/list/top_stories)!
+
+The Polymer News progressive web app is designed to showcase the fast performance and slick user experience possible using modern web platform features, and serve as a jumping-off point for building a more custom and feature-rich app.
+
+It is built with Polymer 1.x (to be upgraded to [Polymer 2.0](https://www.polymer-project.org/2.0/docs/about_20) after 2.0 is released), and patterned after a typical "news" site - with categories and article pages.  It showcases a number of different application patterns:
+
+* It is built using the [Polymer CLI](https://www.polymer-project.org/1.0/docs/tools/polymer-cli) in order to [generate a Service Worker](https://www.polymer-project.org/1.0/toolbox/service-worker) to help it work offline.
```
arthurevans: This sentence reads a little funny to me. Maybe break it in two (* Built with the Polymer CLI, * Uses a service worker.) ... Or skip the CLI part and just mention the service worker part.

app/1.0/blog.yaml
tjsavage: Good point - apologies. This was an old commit hanging around that I never ended up pushing. Will remove.

app/1.0/blog/2017-02-06-news-pwa.md
tjsavage: Will split in two. I kind of want to plug that the CLI generates a SW for you, but could make this sentence better by splitting.

PR #1879, opened 2017-02-16
Add a 1.0/2.0 version switcher

notwaldorf:
This PR is kind of giant but that's mostly because a bunch of files got moved.

TL; DR:
- unversioned the landing page, blog and friends, 404.html/500.html, about.html
- updated the 1.0/2.0 `yaml` pages with these changes
- added a dropdown to the `devguide-shell` (desktop view) and the `pw-shell` (for the mobile drawer)
- for 1.0 docs pages, added a warning that they're deprecated and changed the header colour to a darker grey
- discovered mystical `core-icon` code in `about_10.md` so I nuked it from space
- added a community page!
- change the landing page hero text
- sort of updated the footer links, but I don't really know what should go there, so we should probably update them soon as well

@katejeffreys @arthurevans @frankiefu 
(sorry @frankiefu, you've been looking at docs code this week, suprise review!)


app/elements/pw-shell.html
```diff
@@ -250,13 +286,53 @@
...
+      // maintain this state across the app.
+      this.updateDocsVersion();
+
+      // Update the version when the version selector is used.
+      this.$.versionSelect.addEventListener('iron-select', function(event) {
+        var newSelection = event.detail.item.textContent;
+        if (newSelection === this.activeDocsVersion)
+          return;
+        this.activeDocsVersion = newSelection;
+        window.location = '/' + this.activeDocsVersion + '/start/';
```
frankiefu: I think we can just update the `path` in `<app-location>` to avoid the whole page being reloaded.

notwaldorf:
@keanulee I'm gonna sub you in for @frankiefu now that you're back, if that's ok!

app/elements/pw-shell.html
notwaldorf: Oof, so I kind of rely on the whole page to be reloaded so that the mobile nav bar gets re-pulled from the server. I'm not sure if @keanulee has a better idea :(

notwaldorf: Oh hmm, you mean something like `this._sectionRoute.path = '/' + this.activeDocsVersion + '/start/';`? That seems to do a page-refresh too :/

frankiefu: I mean 
```
<app-location id="appLocation"></app-location>
...
this.$.appLocation.path = '/' + this.activeDocsVersion + '/start/';
```
That shouldn't do a page-refresh.

notwaldorf: Oh yeah that doesn't do a page refresh and as a result the left hand/drawer menu isn't switched out :sob:

frankiefu: I think is fine if it has to do a page refresh when switching version.

arthurevans:
I'm seeing one addition issue with the latest version: when I refresh an API page, the header disappears (dark grey header background is replaced by light grey page background). At this point, the switcher dropdown appears to be broken, too.

notwaldorf:
@arthurevans hmmm i am not seeing that -- i am seeing a weird FOUC flashing, but I recall that only happens locally and goes away when we serve it from app-engine.

notwaldorf:
@arthurevans Ah, figured it out. `iron-doc-viewer` was racy because of the weird way I was lazy loading it, and sometimes it loaded before Polymer and bad things happened.

app/elements/pw-shell.html
keanulee: In the current architecture, there's no data flow from devguide-shell to pw-shell, so setting  this._sectionRoute.path from the dropdown in devguide-shell wouldn't work. Instead, this could be made more similar to how other normal `<a>` links work - iron-location installs a global click handler which prevents default on link clicks, does a `window.history.pushState` with the new path, and fires 'location-changed' on window. So on select changed, we can do a similar thing thru `window`:

```javascript
window.history.pushState({}, '', path);
this.fire('location-changed', {}, {node: window});
```

notwaldorf: Hmm but, I mean, it works in practice, I think. I'm confuzzled.

app/elements/devguide-shell.html
```diff
@@ -216,17 +269,42 @@
...
+      // Figure out which version we're displaying, 1.0 or 2.0. You can't rely on
+      // the parent's version, because if we clicked on the "see 2.0" <a> link inside
+      // this page, then the pw-shell would not have refreshed, so it doesn't
+      // know the version has changed.
+      pwShell.updateDocsVersion();
+      this._setActiveDocsVersion(pwShell.activeDocsVersion);
+      this.$.versionSelect.selected = this.activeDocsVersion == '1.0' ? 0 : 1;
+
+      // Update the version when the version selector is used.
+      this.$.versionSelect.addEventListener('iron-select', function(event) {
```
keanulee: Suggest binding event listener using Polymer-style attributes (to cleanup `attached` callback):

```html
<paper-listbox on-iron-select="_listboxVersionChanged">
```

notwaldorf: 👍

app/elements/pw-shell.html
```diff
@@ -223,6 +251,14 @@
...
         type: Object,
         observer: '_searchParamsChanged'
+      },
+
+      /**
+       * Which version (1.0 or 2.0) of the docs we are using, so that the
+       * main nav and side nav can be kept in sync across page navigations.
+       */
+      activeDocsVersion: {
+        type: String
```
keanulee: Instead of maintaining a separate "version" property and updating it whenever the URL changes, you can just use `this._sectionData.version` which is maintained by `app-route` (or make this a computed property based on that).

keanulee: This way, the URL is considered the "canonical" source of truth of what version is being viewed.

notwaldorf: o shit did i add that there? that's a great property I didn't know I had!!

app/elements/pw-shell.html
```diff
@@ -325,6 +402,11 @@
         Polymer.dom(this).appendChild(document.importNode(fragment, true));
         Polymer.dom.flush();  // Force upgrades.
 
+        // Since the devguide-shell import is loaded, we can remove its
+        // unresolved attribute from the template
+        var devshell = Polymer.dom(this).querySelector('devguide-shell');
+        devshell.removeAttribute('unresolved');
```
keanulee: On a page without `<devguide-shell>`, `devshell` will be null.

app/elements/devguide-shell.html
```diff
@@ -216,17 +269,42 @@
...
           path += 'index';
         }
         this.activePage = path;
       }
+
+      // Figure out which version we're displaying, 1.0 or 2.0. You can't rely on
+      // the parent's version, because if we clicked on the "see 2.0" <a> link inside
+      // this page, then the pw-shell would not have refreshed, so it doesn't
+      // know the version has changed.
+      pwShell.updateDocsVersion();
```
keanulee: If pw-shell gets version from URL, calling this method shouldn't be needed.

Since a new instance of devguide-shell is being created on every page nav (remember, ALL the children of pw-shell is replaced), there's no need to keep `<devguide-shell>.activeDocsVersion` updated with `_setActiveDocsVersion`. Instead, the server template can set the current version through HTML attributes:

base-devguide.html:
```html
<devguide-shell v1-page-link="{{{versions[0]}}}" v2-page-link="{{{versions[1]}}}" active-docs-version="{{{version}}}">
```

notwaldorf: yeah, now that i remember we have this version, it's way easier.

app/elements/pw-shell.html
notwaldorf: Also: as chatted offline: not all URLs have a version, so that's not always true

notwaldorf:
Okay y'all. Did a bunch more changes, all at https://2-0-switcher-dot-polymer-project.appspot.com/

app/elements/devguide-shell.html
```diff
@@ -216,17 +277,39 @@
     },
 
     attached: function() {
+      Polymer.RenderStatus.afterNextRender(this, function() {
+        // Somebody might look at the API docs. Be ready.
+        this.importHref(this.resolveUrl('/elements/iron-doc-viewer.html'));
+      });
+
       // Compute the active page from the parent pw-shell's app route.
       var pwShell = this.parentElement;
```
keanulee: Should use `Polymer.dom(this).parentElement` - this works right now since `pw-shell` has `<content>` at the top level, but that's an implementation detail that shouldn't be relied upon.

app/elements/pw-shell.html
```diff
@@ -250,13 +286,44 @@
...
+      var newSelection = event.detail.item.textContent;
+      if (newSelection === this.activeDocsVersion)
+        return;
+      this.activeDocsVersion = newSelection;
+      this.set('_sectionRoute.path', '/' + this.activeDocsVersion + '/start/');
     },
 
     _toggleDrawer: function() {
+      // Sync the version selector with the active docs version.
+      this.$.versionSelect.selected = this.activeDocsVersion == '1.0' ? 0 : 1;
```
keanulee: The `versionSelect` listbox should be updated whenever `activeDocsVersion` changes, not when the drawer is toggled. Consider using a binding:

```html
<paper-listbox selected="{{activeDocsVersion}}">
```

Or if that doesn't work, an observer:

```javascript
_activeDocsVersionChanged: function(activeDocsVersion) {
  this.$.versionSelect.selected = ...
}
```

app/elements/pw-shell.html
```diff
@@ -250,13 +286,44 @@
...
+
+      var links = this.querySelectorAll('a[data-version]');
```
keanulee: Since the contents of `<pw-shell>` is reloaded on version changed (and replaced with only links to the relevant version), this (expensive) work of toggling `hidden` should not be needed.

notwaldorf: For the mobile UI we have to send _all_ the links, both for v1 or v2 up front (on desktop, the`devguide-shell` gets switched out on every page refresh, but the mobile drawer (in the `pw-shell` is done just once, and if you start off at a _non_versioned page, you're hosed.

app/elements/pw-shell.html
notwaldorf: Why is an extra observer better than this? If you're not looking at the drawer, you also don't need to have the right version displaying...

keanulee: True, but this means you'll be doing extra work every time you toggle the drawer (e.g. navigating between 2.0 pages, or when closing the drawer and the listbox goes out of view). There's also cases where you can change the `activeDocsVersion` without toggling the drawer (e.g. switch versions, open the drawer, use the browser back button to switch back to the last version). Either way, it's more "correct" to update the listbox when the version has actually changed (not just because the drawer is toggled).

notwaldorf: Ugh, fineeee

app/elements/pw-shell.html
keanulee: For versioned pages (e.g. "/1.0/start/"), we only send v1 links (remember that even the drawer links on mobile are light DOM children of `<pw-shell>` that are _projected_ into the drawer through `<content select=".site-nav">`).

For non-versioned pages (e.g. "/blog/"), we could consider only including 2.0 links (forcing docs users to 2.0 even if they previously  browsed 1.0), but that's probably a little harsh. Another less expensive way to do this toggling (instead of iterating through all the links) is to make `<pw-shell>.activeDocsVersion` reflect to attribute, then style the other links in _main.scss, like:

```css
pw-shell[active-docs-version=1.0] a[data-version=2.0] {
  display: none;
}
```

app/elements/pw-shell.html
```diff
@@ -164,11 +173,17 @@
...
-            <a href="/2.0/blog/" name="1.0/blog">News</a>
-            <a href="https://elements.polymer-project.org/" target="_blank" name="catalog">Element Catalog</a>
+
+            <a href="/1.0/start/" name="1.0/start" hidden="[[_shouldShowV2(activeDocsVersion)]]">Start</a>
+            <a href="/1.0/docs/devguide/feature-overview" name="1.0/docs" hidden="[[_shouldShowV2(activeDocsVersion)]]">Polymer</a>
+            <a href="/1.0/toolbox/" name="1.0/toolbox" hidden="[[_shouldShowV2(activeDocsVersion)]]">App Toolbox</a>
+
+            <a href="/2.0/start/" name="2.0/start" hidden="[[!_shouldShowV2(activeDocsVersion)]]">Start</a>
+            <a href="/2.0/docs/about_20" name="2.0/docs" hidden="[[!_shouldShowV2(activeDocsVersion)]]">Polymer</a>
+            <a href="/2.0/toolbox/" name="2.0/toolbox" hidden="[[!_shouldShowV2(activeDocsVersion)]]">App Toolbox</a>
```
keanulee: My only other suggestion is to style away the irrelevant links here on the `active-docs-version` attribute as well (similar to the mobile drawer) to avoid the work of `_shouldShowV2`:

```css
:host([active-docs-version="1.0"]) .section-tabs a[data-version="2.0"] { ... }
```

notwaldorf: 😭 ok

PR #1881, opened 2017-02-17
Adds bundling to make things faster

notwaldorf:
I added a `polymer build` step so that we can bundle and lazy load elements and be lazier and faster and then I made everything lazier and faster (maybe).

Deployed site, 3G (start render: 5.372s, speed index: 5572)
https://www.webpagetest.org/result/170222_DQ_VB61/
Lighthouse:
master

This branch, 3G (start render: 4.796s, speed index: 4883 and a slightly slower first byte actually)
https://www.webpagetest.org/result/170222_3H_VB64/
Lighthouse:
branch

🎉🎉🎉 

Review time! @arthurevans (for FYI) @frankiefu (because he's been looking at it for all week and is the 👑 of perf)


arthurevans:
Two regressions vs. the original site:

- Visual flashes on load. Initial load of a guide page on a clean load shows unstyled left nav, then the drawer and header flash into place with the page content.

- API doc page doesn't load if you navigate directly to it (or hard-reload an existing page). For example:

    https://pee-wees-big-lazy-loading-adventure-dot-polymer-project.appspot.com/2.0/docs/api/array-selector

notwaldorf:
@arthurevans Pushed fixes for both!

notwaldorf:
Ok friends, pls review this for realsies. @arthurevans @frankiefu

frankiefu:
The blog links are not working (e.g. https://pee-wees-big-lazy-loading-adventure-dot-polymer-project.appspot.com/1.0/blog/2017-02-13-news-pwa.html), expected?

polymer.json
```diff
@@ -0,0 +1,19 @@
+{
+  "entrypoint": "app/1.0/index.html",
+  "shell": "app/elements/pw-shell.html",
+  "fragments": [
+    "app/elements/pw-shell.html",
```
frankiefu: Do we need to have `pw-shell.html` in the fragments?

gulpfile.js
```diff
@@ -268,7 +287,7 @@ gulp.task('copy', 'Copy site files (polyfills, templates, etc.) to dist/', funct
     .pipe(gulp.dest('dist'));
 
   let bower = gulp.src([
-      'app/bower_components/webcomponentsjs/webcomponents*.js'
+      'app/bower_components/webcomponentsjs/webcomponents*.js',
```
frankiefu: remove trailing comma?

polymer.json
notwaldorf: I think so, because we want it to get bundled, right? I'm basically only using what the `elements` folder contains, and nothing else in the `app` output.

notwaldorf:
@frankiefu confirmed the blog links don't work on master either

polymer.json
frankiefu: Since it's already in `"shell": "app/elements/pw-shell.html",`, then we don't need to put it in the fragments.

notwaldorf: Talked offline: then it puts it in a `1.0` folder in the output; if i add it in the fragments, then i can get all my bundled elements from the same place. I'm holding the tool weirdly, look away :D

notwaldorf:
@frankiefu fixed `app-drawer` fouc-ing on mobile

notwaldorf:
MERGINNNNN.

notwaldorf:
THANKS FRANDSSSSSS!!!

PR #1886, opened 2017-02-23
fix md doc quotes

notwaldorf:
Discovered in the perf branch that this page rendered funky, because a body style was leaking into the document. The backticks seemed to be mismatched/missing in places, so I tried to fix them.

Feel free to throw this away if someone else already has a fix locally or something.

@arthurevans @katejeffreys


arthurevans:
Thanks! This is also fixed in #1885... Since I made some other formatting fixes to this file there, I'm going to close this one.

PR #1887, opened 2017-02-24
Web font and copy changes

notwaldorf:
👀 at: https://web-font-and-copy-changes-dot-polymer-project.appspot.com

- ➕ partner logos to the landing page. there needs to be a spec for logos cause they're alllll different ratios.
- 📇 content to the community page (though @wmginsberg is going to clean it up)
- 🔥 deleted the old community pages
- 🔧 fixed the footer links
-💣 the `Roboto` font from the body content (the slab titles and code are still web fonty). This is part of my war on web fonts to prevent this sort of meh experience that was currently happening:
waiting-for-roboto
Now:
screen shot 2017-02-24 at 3 36 48

fyi @arthurevans


arthurevans:
Comparing the new and old versions, it seems like there's more line spacing in this version.

The regular text *seems* a bit more wide-spaced (although the numbers are the same, so I may be mistaken; it may just be the way the font looks).

But the extra padding on the bulleted lists makes them look funky when they have nested paragraphs. See, for example, the end of this page: 

https://web-font-and-copy-changes-dot-polymer-project.appspot.com/1.0/docs/devguide/feature-overview

arthurevans:
Other than the bulleted list note, LGTM!

notwaldorf:
@arthurevans ah, you're totally right, i'll just add the style to that one specific page

notwaldorf:
@arthurevans fixed and redeployed!
@tjsavage does it look good? main page and community pages, mostly

arthurevans:
Bulleted lists now LGTM.

At the bottom of the page, the news excerpts look a little weird. Like, the third one appears to be only half as wide as #1 and #2. Is that intentional?

arthurevans:
Also: dark blue link text on black section of community page is really hard to read.

notwaldorf:
@arthurevans huh, the weird columns are reproducible in https://www.polymer-project.org/1.0/ too, so i think this isn't a new bug (but can look into fixing)

arthurevans:
So it does. Weird. In that case, LGTM 🍡

tjsavage:
LGTM! Awesome work

arthurevans:
@notwaldorf should I go ahead and merge this?

notwaldorf:
@arthurevans yes please!

PR #1891, opened 2017-02-26
Update deploy script to work with latest gcloud.

arthurevans:
R: @notwaldorf


notwaldorf:
A very enthusiastic LGTM, since i always modify this by hand :)

PR #1892, opened 2017-02-27
Add new styling docs

katejeffreys:


app/2.0/docs/devguide/encapsulated-style.md
```diff
@@ -0,0 +1,639 @@
...
+```html
+<!-- Create a template for the custom element -->
+<dom-module id='custom-element'>
+  <template>
+    <h1>Heading!</h1>
+    <p>We are elements in custom-element's local DOM.</p>
+  </template>
+</dom-module>
+
+<!-- Drop the custom element on the page -->
```
arthurevans: This looks like you're putting this all in the index.html file, which is not going to work on non-Chrome browsers unless you have a workaround in there (it's documented here,  https://2-0-staging-dot-polymer-project.appspot.com/2.0/docs/devguide/registering-elements#main-document-definitions, but as it turns out that's going to need to be updated slightly for 2.0).

The best way to deal with this (and the reason we tend to use plunker) is so we can define custom elements in their own files (`custom-element.html`), and import and use them in another file (`index.html`).

Also, these plunkers need to have the polyfills added so they work on non-Chrome browsers.
Put these together, and this plunker would look like this:

https://plnkr.co/edit/dnL9SeRFZr0sQt0pqKaG?p=preview

app/2.0/docs/devguide/encapsulated-style.md
```diff
@@ -0,0 +1,639 @@
+---
+title: Encapsulated style with Polymer 2.0 and shadow DOM v1
+---
+
+<!-- toc -->
+
+## Styling your elements
+
+Polymer supports DOM templating and the shadow DOM API. When you use the `<dom-module>` and `<template>` blocks to template your custom element, Polymer attaches a shadow root to instances of this element. Polymer then copies in the contents of the template you provided for your element.
```
arthurevans: So the kindly folks in core just added a _new_ way to define a DOM template, in addition to the old standard `dom-module`. We talk about that in the DOM templating chapter, so maybe just be less specific here: "When you provide a DOM template for your custom element, ..."

app/2.0/docs/devguide/encapsulated-style.md
```diff
@@ -0,0 +1,639 @@
...
+  class CustomElement extends Polymer.Element {
+    static get is() {
+      return "custom-element";
+    }
+  }
+  customElements.define(CustomElement.is, CustomElement);
+</script>
+```
+
+The HTML elements in your template become children within your custom element's shadow DOM. Shadow DOM provides a mechanism for encapsulation, meaning that elements in side the shadow DOM are not accessible to styling rules external to the shadow DOM.
```
arthurevans: within => in
in side => inside
"not accessible to styling" Might want to say "don't match selectors outside the shadow DOM" or say what you did, but then add, "Specifically, elements inside a shadow tree don't match selectors specified outside that shadow tree—in the main document, or in another shadow tree."

does that make sense?

app/2.0/docs/devguide/encapsulated-style.md
```diff
@@ -0,0 +1,639 @@
...
+Shadow DOM permits encapsulation of styling rules for custom elements. You can freely define styling information for your elements, such as fonts, text colors, and classes, without fear of the styles applying outside the scope of your element.
+
+Here's an example:
+
+[See it on Plunker](http://plnkr.co/edit/cHSjdQTa0h6fWXAygkje?p=preview)
+
+```html
+<!-- Document-level stylesheet -->
+<style>
+	.myclass {
```
arthurevans: Indentation is off here... should be two-space indents like elsewhere.

app/2.0/docs/devguide/encapsulated-style.md
```diff
@@ -0,0 +1,639 @@
...
+		color:blue;
+	}
+</style>
+<dom-module id='x-foo'>
+  <template>
+    <!-- Encapsulated, element-level stylesheet -->
+    <style>
+		  p {
```
arthurevans: Also indentation.

app/2.0/docs/devguide/encapsulated-style.md
```diff
@@ -0,0 +1,639 @@
...
+
+Polymer supports DOM templating and the shadow DOM API. When you use the `<dom-module>` and `<template>` blocks to template your custom element, Polymer attaches a shadow root to instances of this element. Polymer then copies in the contents of the template you provided for your element.
+	
+Here's an example:
+
+[See it on Plunker](http://plnkr.co/edit/VPktVlI0PWMnYdYVsVBx?p=preview)
+
+```html
+<!-- Create a template for the custom element -->
+<dom-module id='custom-element'>
```
arthurevans: Random point of style here: our convention is to use double quotes in HTML (id="custom-element") and single quotes in JSON unless there's a compelling reason to do otherwise.

app/2.0/docs/devguide/encapsulated-style.md
```diff
@@ -0,0 +1,639 @@
...
+	  <p>I'm a shadow DOM child element of x-foo.</p>
+	  <p class="myclass">So am I.</p>
+  </template>
+</dom-module>
+<x-foo></x-foo>
+<!-- The following paragraph uses the document-level stylesheet. -->
+<p class="myclass">I have nothing to do with x-foo. Because of encapsulation, x-foo's styles won't leak to me.</p>
+```
+
+For a detailed explanation of shadowDOM as it applies to Polymer, see [arthur's article]().
```
arthurevans: shadowDOM => shadow DOM. 
`[arthur's article]()` => `[Shadow DOM concepts](shadow-dom)`

app/2.0/docs/devguide/encapsulated-style.md
```diff
@@ -0,0 +1,639 @@
...
+x-foo {
+	background-color: blue;
+}
+```
+
+#### Using selectors 
+
+You can use CSS selectors to determine when and how to style the host. In this code sample:
+
+* The rule `:host { font-family: sans-serif; }` always matches `<x-foo>` elements
```
arthurevans: Rather than have the whole rule, I'd just include the selector. "The selector `:host` matches any `<x-foo>` element."

app/2.0/docs/devguide/encapsulated-style.md
```diff
@@ -0,0 +1,639 @@
...
+* The rule `:host(.blue) {color: blue;}` matches `<x-foo>` elements of class `blue`
+* The rule `:host(.red) {color: red;}` matches `<x-foo>` elements of class `red`
+* The rule `:host(:hover)` matches `<x-foo>` elements when they are hovered over
+
+[See it on Plunker](http://plnkr.co/edit/FsXnCAz65SR6fZ7YKuy6?p=preview)
+
+```html
+<dom-module id="x-foo">
+	<template>
```
arthurevans: Indentation

app/2.0/docs/devguide/encapsulated-style.md
```diff
@@ -0,0 +1,639 @@
...
+		}
+		</style>
+    <p>Make this text red if x-foo has class "warning", and black otherwise.</p>
+	</template>
+</dom-module>
+<x-foo class="warning"></x-foo>
+<x-foo></x-foo>
+```
+
+Styling with the `:host` selector is one of two instances where rules inside a shadow tree can affect an element outside a shadow tree. The second instance uses the `::slotted()` syntax to apply styling rules to content populated from the main DOM at runtime.
```
arthurevans: "from the main DOM" is not quite right, because the distributed children could be coming from a parent shadow tree. I think you should say "distributed children" here. (Maybe reference the shadow DOM concepts chapter for a discussion, although that definition is kind of buried in the "Shadow DOM and composition" section.

app/2.0/docs/devguide/encapsulated-style.md
```diff
@@ -0,0 +1,639 @@
...
+
+### Styling slotted content (distributed children)
+
+You can create slots in an element's template that are populated at runtime using this syntax:
+
+[See it on Plunker](http://plnkr.co/edit/MRN9blKg6A3w8G0RkyJD?p=preview)
+
+```html
+<dom-module id="x-foo">
```
arthurevans: indentation

app/2.0/docs/devguide/encapsulated-style.md
```diff
@@ -0,0 +1,639 @@
...
+```
+
+To style slotted content, use the `::slotted()` syntax.
+
+`::slotted(*)` selects all slotted content:
+
+[See it on Plunker](http://plnkr.co/edit/cGrbntuantGtqj7Poqft?p=preview)
+
+```html
+<dom-module id="x-foo">
```
arthurevans: check indentation

app/2.0/docs/devguide/encapsulated-style.md
```diff
@@ -0,0 +1,639 @@
...
+</x-foo>   
+```
+
+[See it on Plunkr](http://plnkr.co/edit/suVhRDAB5Rq7Q7T6zbMI?p=info)
+
+You can select by element type:
+
+```html
+<dom-module id="x-foo">
+	<template>
```
arthurevans: check indentation... I'm going to stop marking these, I think just check the rest.

app/2.0/docs/devguide/encapsulated-style.md
```diff
@@ -0,0 +1,639 @@
...
+    </template>
+</dom-module>
+
+<x-foo>
+    <div slot="para1">I'm green!</div>
+    <div slot="para2">I'm not green.</div>
+</x-foo> 
+```
+
+For more examples, see [Rob Dodson's Shadow DOM CSS Cheat Sheet](http://robdodson.me/shadow-dom-css-cheat-sheet/).
```
arthurevans: Not sure this link is worth it... Not a lot of examples and some are likely to be confusing, like host-context which we don't support.  (also, the statements about browser support are really out of date.)

app/2.0/docs/devguide/encapsulated-style.md
```diff
@@ -0,0 +1,639 @@
...
+<dom-module id="my-style-module">
+	<template>
+		<style>
+			<!-- Styles to share go here! -->
+		</style>
+	</template>
+</dom-module>
+```
+
+When you create the element that will use the styles, include the style module in the opening tag of the style block:
```
arthurevans: An important missing bit here is that you almost invariably define the style module in a different file than you use it in. So you should show the style module with a filename (`my-style-module.html`) and show the custom element file (`new-element.html`) importing the style module at the top using a `<link rel="import" href="my-style-module.html">` (or whatever path you like).

app/2.0/docs/devguide/encapsulated-style.md
```diff
@@ -0,0 +1,639 @@
...
+<script>
+    class MyAwesomeButton extends Polymer.Element {
+      static get is() {
+        return "my-awesome-button";
+      }
+    }
+    customElements.define(MyAwesomeButton.is, MyAwesomeButton);
+</script>
+
+## Third-party libraries that modify local DOM {#scope-subtree}
```
arthurevans: This section can go, we don't need scopeSubtree in 2.0.

app/2.0/docs/devguide/encapsulated-style.md
```diff
@@ -0,0 +1,639 @@
...
+
+## Styling at the document level
+	
+Browsers that implement the current Shadow DOM v1 specifications will automatically encapsulate styles, scoping them to the elements in which they were defined.
+
+Some browsers have not implemented the Shadow DOM v1 specifications. To make sure your apps and elements display correctly in these browsers, you'll need to use `custom-style` to ensure that styling information doesn't "leak" into the local DOM of your elements.
+
+The `custom-style` extension enables a set of polyfills that ensure that styles in your apps and elements behave as you would expect from the Shadow DOM v1 specifications, even in browsers that don't implement these specifications.
+
+`custom-style` is an extension of the native `<style>` element. To ensure that your styles behave according to the Shadow DOM v1 specifications in all browsers, use `custom-style` when you define *document-level* styles.
```
arthurevans: The phrase "is an extension" could be confusing... In 2.0 use use a wrapper element (`<custom-style>`). In 1.x we had a "type-extension element" which... is an extension of the native `<style>` element.

app/2.0/docs/devguide/encapsulated-style.md
```diff
@@ -0,0 +1,639 @@
...
+  <dom-module id="x-foo">
+    <template>
+      <p>Paragraph B: I am in the local DOM of x-foo. I am black on all browsers.</p>
+    </template>
+  </dom-module>
+  <x-foo></x-foo>   
+```
+  
+### Syntax and compatibility
+
```
arthurevans: Maybe some words here.

app/2.0/docs/devguide/encapsulated-style.md
```diff
@@ -0,0 +1,639 @@
...
+#### Polymer 2.x
```
arthurevans: Style wise, for consistency when captioning a code sample you can use:

`Polymer 2.x {.caption}`

app/2.0/docs/devguide/encapsulated-style.md
```diff
@@ -0,0 +1,639 @@
...
+```html
+<custom-style>
+  <style>
+    p {
+		...
+    }
+  </style>
+</custom-style>
+````
+#### Compatibility between 1.x and 2.x
```
arthurevans: Hybrid (compatible with 1.x and 2.x)

app/2.0/docs/devguide/encapsulated-style.md
```diff
@@ -0,0 +1,639 @@
...
+<style is="custom-style">
+  p {
+	  ...
+  }
+</style>
+```
+
+### `custom-style` and custom properties
+
+In a DOM module template, you can define custom properties to be used by Polymer's shim for cross-scope styling. Then, use an `include` attribute in a `style` tag to share that style data between multiple custom-style elements.
```
arthurevans: This makes it sound like you need to use a style module to define custom properties, but that's not the case. They just need to be in a custom-style.

Maybe, since this doesn't introduce custom properties, this section should be in the custom properties chapter, and be called something like, "Setting custom properties at the document level" with a quick x-ref back here?

app/2.0/docs/devguide/encapsulated-style.md
```diff
@@ -0,0 +1,639 @@
...
+	
+Browsers that implement the current Shadow DOM v1 specifications will automatically encapsulate styles, scoping them to the elements in which they were defined.
+
+Some browsers have not implemented the Shadow DOM v1 specifications. To make sure your apps and elements display correctly in these browsers, you'll need to use `custom-style` to ensure that styling information doesn't "leak" into the local DOM of your elements.
+
+The `custom-style` extension enables a set of polyfills that ensure that styles in your apps and elements behave as you would expect from the Shadow DOM v1 specifications, even in browsers that don't implement these specifications.
+
+`custom-style` is an extension of the native `<style>` element. To ensure that your styles behave according to the Shadow DOM v1 specifications in all browsers, use `custom-style` when you define *document-level* styles.
+
+*Note: You should only use `custom-style` to define styles for the main document. To define styles for an element's local DOM, just use a `<style>` block.*
```
arthurevans: We need to add that this is optional and it needs to be imported separately (unless you're importing the full `polymer.html`). 

You could show an example of an `index.html` that imports the polyfills and then imports `polymer/lib/elements/custom-style.html` (since it's not _defining_ any Polymer elements, 
it doesn't need to import polymer.html or polymer-element.html). 

It would then import the custom element, and use a custom-style to apply styles... make sense?

app/2.0/docs/devguide/encapsulated-style.md
```diff
@@ -0,0 +1,639 @@
...
+[See it on Plunker](http://plnkr.co/edit/yiD9XWPHaMjHaGGwu4V9?p=preview)
+```html
+  <custom-style>
+  <style>
+      p {
+        color: red;
+      }
+    </style>
+  </custom-style>
+  <p>Paragraph A: I am in the main DOM. I am red.</p>
```
arthurevans: main DOM => main document

app/2.0/docs/devguide/encapsulated-style.md
```diff
@@ -0,0 +1,639 @@
...
+  <dom-module id="x-foo">
+    <template>
+      <p>Paragraph B: I am in the local DOM of x-foo. I am black on all browsers.</p>
```
arthurevans: local DOM => shadow DOM

arthurevans:
Will review the custom properties doc later tonight, but I wanted to send off the first set of comments. One thing I did note on the custom properties doc is that it doesn't cover the updateStyles API. It's essentially the same as in 1.x (see https://www.polymer-project.org/1.0/docs/devguide/styling#style-api), except that customStyles has been eliminated in favor of just using:

```
this.updateStyles({
  '--some-custom-style': 'green',
  '--another-custom-style': 'blue'
});
```

app/2.0/docs/devguide/encapsulated-style.md
katejeffreys: @arthurevans I thought I was including the polyfills by having this import link in the plunks. Am I mistaken.. or should I do it differently?

<link rel="import" href="https://polygit.org/polymer+:2.0-preview/webcomponentsjs+:v1/shadydom+webcomponents+:master/shadycss+webcomponents+:master/custom-elements+webcomponents+:master/components/polymer/polymer.html">

katejeffreys: Github comments system seems to have eaten my link, but it is the first line in every plunk

app/2.0/docs/devguide/encapsulated-style.md
katejeffreys: https://developers.google.com/web/fundamentals/getting-started/primers/shadowdom#composition_slot has a good description of it, so I linked to that

app/2.0/docs/devguide/encapsulated-style.md
arthurevans: The import link imports the polymer library. The polyfills are loaded using a `<script>` tag. In 1.x this was always `webcomponentsjs/webcomponents-lite.js` ... (Now we have a new option, `webcomponentsjs/webcomponents-loader.js`, which automatically loads the correct set of polyfills for the current browser.)

http://plnkr.co/edit/dnL9SeRFZr0sQt0pqKaG?p=preview

The polyfills can only be imported _once_, so you typically load them in the main document. 

HTML imports get deduplicated by the browser, so every custom element can have a `<link rel="import">` at the top of it for Polymer, but Polymer only gets loaded once. Each element should import its own dependencies this way.

app/2.0/docs/devguide/cross-scope-styling.md
```diff
@@ -0,0 +1,331 @@
...
+
+Thus, custom CSS properties are a powerful way for element authors to
+expose a theming API to their users in a way that naturally fits right alongside
+normal CSS styling. 
+
+### Custom CSS mixins
+
+It may be tedious (or impossible) for an element author to predict every
+CSS property that may be important for theming, let alone expose every
+property individually.
```
arthurevans: CSS mixins are a proposal that has not been implemented by any of the browsers. 
To use them, you need to import an optional polyfill. 

Any element using mixins should import the polyfill. 

```
<!-- import CSS mixins polyfill -->
<link rel="import" href="/bower_components/shadycss/apply-shim.html">
```
For backward compatibility, the `polymer.html` import includes the CSS mixins polyfill. No extra import is required when defining hybrid elements.

app/2.0/docs/devguide/cross-scope-styling.md
```diff
@@ -0,0 +1,331 @@
...
+future the `my-toolbar` author can rename the `title` class or
+restructure the internal details of `my-toolbar` without changing the custom
+property exposed to users.
+
+You can also include a default value in the `var()` function, to use in case the user
+doesn't set the custom property:
+
+```css
+color: var(--my-toolbar-title-color, blue);
+```
```
arthurevans: Should probably also show a default value that's a custom property. Some people get it wrong, and in 1.x we supported an incorrect syntax.

Proper syntax is:

`var(--my-color, var(--my-default-color))`

app/2.0/docs/devguide/cross-scope-styling.md
```diff
@@ -0,0 +1,331 @@
...
+
+### Custom CSS mixins
+
+It may be tedious (or impossible) for an element author to predict every
+CSS property that may be important for theming, let alone expose every
+property individually.
+
+An element author can define a set of CSS properties as a single custom property and
+then allow all properties in the set to be applied to a specific CSS rule
+in an element's local DOM. The extension enables this with a mixin capability
```
arthurevans: local DOM => shadow DOM

app/2.0/docs/devguide/cross-scope-styling.md
```diff
@@ -0,0 +1,331 @@
...
+        };
+      }
+    </style>
+    <my-toolbar title="This one is green."></my-toolbar>
+    <my-toolbar title="This one is green too."></my-toolbar>
+    <my-toolbar class="warning" title="This one is red."></my-toolbar>
+  </template>
+
+  <script>
+    Polymer({ is: 'my-element'});
```
arthurevans: update to class syntax

app/2.0/docs/devguide/cross-scope-styling.md
```diff
@@ -0,0 +1,331 @@
...
+  </script>
+</dom-module>
+```
+
+## Inheritance
+
+If an element doesn't override styling information, that element inherits styles from its parent:
+
+```
+<style is="custom-style">
```
arthurevans: This should be wrapped in a `<custom-style>`
Also, should probably show the import for the custom-style element.

app/2.0/docs/devguide/cross-scope-styling.md
```diff
@@ -0,0 +1,331 @@
...
+</style>
+```
+
+[See it on JSBin](https://jsbin.com/ruliruy/edit?html,output)
+
+Note that the font family is inherited, but the text color is not. This is because paper-checkbox overrides the text color.
+
+## Mixins
+
+It may be tedious (or impossible) for an element author to predict every CSS property that may be important for theming, let alone expose every property individually.
```
arthurevans: I think this section is repeated.

app/2.0/docs/devguide/encapsulated-style.md
katejeffreys: Could you explain a bit more... In what sense is it optional if document-level styles don't work without it?

app/2.0/docs/devguide/encapsulated-style.md
katejeffreys: I'm going to omit this section for now. It seems to be redundant IMO

app/2.0/docs/devguide/encapsulated-style.md
arthurevans: Ah, sorry.

The custom-style element is an optional _import_: That is, it isn't included in the `polymer-element.html` import. 

So the example might look like:

    index.html {.caption}

    ```
    <!-- Load polyfills -->
    <script src="/bower_components/webcomponentsjs/webcomponents-loader.js"></script>
    <!-- Load optional CSS mixin polyfill (if using) -->
    <link rel="import" href="/bower_components/shadycss/apply-shim.html">
    <!-- Load custom-style element -->
    <link rel="import" href="/bower_components/polymer/lib/elements/custom-style.html">
    
    <custom-style>
      <style>...</style>
    </custom-style>
    ```
For hybrid components (for example, if you're creating a default theme that's imported as a set of custom properties and mixins), you'd just import `polymer.html`, which is backward-compatible with 1.x (the mixins shim and the custom-style element are already included there, like they are in 1.x).

app/2.0/docs/devguide/encapsulated-style.md
katejeffreys: So if I add 

`<script src="webcomponentsjs/webcomponents-loader.js"></script>`

after the `<link rel=blablablahb>` to every index.html, would that work?

arthurevans:
I think we still need some changes here (polyfill limitations in the styling chapter, some code sample cleanup), but I'm going to merge this so we can stage the whole thing and we can iterate on it.

PR #1896, opened 2017-02-28
Update getting started material

katejeffreys:


app/2.0/start/samples/editable-name-tag/editable-name-tag.html
```diff
@@ -0,0 +1,37 @@
...
+      <input is="iron-input" placeholder="Your name here...">
+    </iron-input>
+  </template>
+
+  <script>
+    class EditableNameTag extends Polymer.Element {
+      static get is() { return "editable-name-tag"; }
+      
+      // configure the owner property
+      static get config() {
```
arthurevans: This should be:

```
static get properties() {
  return {
    owner: {
      type: String,
      value: 'Daniel'
    }
  }
}
```

arthurevans: Same change for configurable-name-tag sample (also, in that sample, the {{owner}} text binding should be changed to [[owner]]

app/2.0/start/quick-tour.md
```diff
@@ -134,10 +134,10 @@ It also shows configuring the owner property from markup in `index.html`.
 
 <demo-tabs selected="0" src="http://plnkr.co/edit/DhDSeqNrmLflcQ8UZI1R?p=preview">
```
arthurevans: This plunker link goes to a 1.x sample.

arthurevans:
We still need to:

- Update the two samples using the old syntax (configurable-name-tag and editable-name-tag), and change "local DOM" to "shadow DOM" throughout. 
- Fix the CTAs at the bottom of the doc to link to appropriate places.
- Also (sorry, missed this first time)--in the first example, please change the `ready()` callback to a `constructor()` like the `name-tag` example.

katejeffreys:
Changes made. Can't test because can't build docs site as discussed

katejeffreys:
@arthurevans built and tested. Ready for someone else to check it!

PR #1898, opened 2017-03-01
New section for the API docs

notwaldorf:
Ok I did a thing to help with the new API page maybe? Basically I versioned the `api` section which is only linkable from the `Polymer` nav section, unfortunately and therefore kind of weird, but also solved @justinfagnani's docs problems

(it still has some nav bugs in the mobile UI, but I didn't want to sink time into this if we think this is a bad idea)

Demo: https://new-docs-placeholder-dot-polymer-project.appspot.com

/cc @arthurevans @keanulee


keanulee:
On https://new-docs-placeholder-dot-polymer-project.appspot.com/1.0/api/array-selector, when I switch the desktop listbox from 1.0 to 2.0, it `pushState`s to https://new-docs-placeholder-dot-polymer-project.appspot.com/2.0/api/ which is a 404. I think you just need to set the default URL that it goes to after the switch to /2.0/api/Polymer.Base instead.

Also, when you go from 2.0 to 1.0, it goes to /1.0/api/. Even though that redirects to /1.0/api/Polymer.Base, that's not reflected by the selected state in left nav. Should change the default to /1.0/api/Polymer.Base as well (instead of /1.0/api/).

arthurevans:
As you said, this is kind of weird. If we're going to give the API reference its own left nav, it should just have its own tab, IMO. 

Start / Guides / API / App Toolbox doesn't exactly make sense, maybe. But maybe better than weird.

notwaldorf:
I think we can close this PR in favour of the `2.0-api` branch, right?

PR #1911, opened 2017-03-04
Give me my hrefs back plz.

arthurevans:
FYI: @notwaldorf in case this breaks anything else.

I was wondering why all the cross-references in the docs were broken, and I tracked it back to this eentsy change somewhere back a couple of weeks ago. Which somehow wasn't in the branch I was working on when I was testing cross-references earlier this week :(.

Tested, doesn't seem to break anything.

### DIFF
--- a/server.py
+++ b/server.py
@@ -264,7 +264,10 @@ class Site(http2push.PushHandler):
       data = {
         'nav': nav,
         'full_nav': full_nav,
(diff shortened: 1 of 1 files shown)


notwaldorf:
Argh, crap, i totally deleted this because I couldn't see it used anywhere. Sorry @arthurevans!

PR #1913, opened 2017-03-06
Some finishing touches

katejeffreys:


app/2.0/homepage/contact-card/contact-card.html
```diff
@@ -32,16 +32,18 @@
         display: none;
       }
     </style>
-    <content></content>
+    <slot></slot>
 
     <iron-icon icon="star" hidden$="{{!starred}}"></iron-icon>
   </template>
   <script>
-    Polymer({
```
notwaldorf: Hmm, this is code that's actually being run, right? In that case, the site is running on Polymer 1.0, which means we can't actually use a 2.0 element on that page :(

app/index.html
```diff
@@ -338,16 +338,18 @@ <h2 class="underline">Fully interoperable</h2>
...
       &lt;iron-icon icon=&quot;star&quot; hidden$=&quot;{{!starred}}&quot;&gt;&lt;/iron-icon&gt;
     &lt;/template&gt;
     &lt;script&gt;
-      Polymer({
-        is: 'contact-card',
-        properties: {
-          starred: Boolean
+      class ContactCard extends Polymer.Element {
+        static get is() { return "contact-card"; }
+          constructor() {
```
notwaldorf: Oh sorry, I think the `static` and `constructor` should also left align (maybe the easiest is to indent `static` by 2?)

katejeffreys:
Thanks!!

PR #1931, opened 2017-03-08
Update about_20.md

andersforsell:
Fix Polymer version


arthurevans:
Thanks for the fix! Looks like this has been fixed in master.

PR #1936, opened 2017-03-09
dom-module `element-name` => `x-custom`

noahlaux:
This PR changes the `id` in dom-module to match the corresponding `is` property on the element.


noahlaux:
I signed it!

PR #1942, opened 2017-03-11
Clean up server code; Use CSS instead of bindings to style devguide header

keanulee:
This started with me trying to use CSS to style the devguide instead of bindings, then I got carried away...


app/elements/pw-shell.html
```diff
@@ -360,23 +361,23 @@
       }
     },
 
-    _selectedSection: function(sectionCR) {
-      var data = sectionCR.base;
-      return data.version+'/'+data.section;
+    _computeSection: function(path) {
+      var match = /^\/([12]\.0\/[^\/]+|blog|community)\//.exec(path);
```
notwaldorf: omg wat is this even. pls halp with comment halp

app/elements/pw-shell.html
```diff
@@ -161,11 +158,11 @@
       }
     </style>
 
-    <app-location route="{{_sectionRoute}}"></app-location>
-    <app-route route="{{_sectionRoute}}" pattern="/:version/:section" data="{{_sectionData}}" tail="{{_subsectionRoute}}"></app-route>
-    <app-route route="{{_sectionRoute}}" pattern="/search/" query-params="{{_searchParams}}"></app-route>
+    <app-location route="{{_route}}" path="{{path}}"></app-location>
```
notwaldorf: I think this is the only part i find a bit scary because I don't know how app-location really works

arthurevans:
What's the benefit of these changes? I'd want to test them pretty thoroughly since there are substantial changes to the client and server logic, but I don't know why this is better than the old way.

keanulee:
For one, it fixes an issue where navigating from a non-homepage (e.g. https://www.polymer-project.org/2.0/docs/about_20) to the homepage (by clicking on the logo) doesn't clear the selected tab underline. The server logic hasn't changed - just renamed a bunch of stuff and eliminated dead code

notwaldorf:
tbh i think this PR is more likely to be merged sooner if we didn't do any of the renaming, and just fixed the bug, and left the refactoring for a time when we're not in crunch mode

app/elements/pw-shell.html
keanulee: `<app-location>` internally keeps `route.path` the same value as `path`, but in our case we can just use `path` directly and not rely on `this.set`. (We still need `route` for the `<app-routes>` below though.)

keanulee:
I renamed the variables on the python server because I encountered them while fixing the templates and the variable names were not clear to me. I don't really think that they make a huge difference, and going back to revert them seems counterproductive. Not a super urgent fix (given it's been an issue for a while), so happy to let this PR sit until it's well tested.

katejeffreys:
@keanulee @arthurevans Could you briefly outline the kind of things to look for in testing? What would I see if it breaks?

keanulee:
This PR touches routing stuff and switching between docs versions (1.0 and 2.0), so navigating between sections and pages, changing the version, etc. and noting if clicking on the link does navigate to the correct page without full page refresh, the header style is consistent with the docs version, search works as expected...

arthurevans:
@katejeffreys Everything Keanu said. Also some general notes, when testing site changes like, I usually want to test:

- Navigating to URLs by following links on the doc site, as well as by entering the full URL directly (which is equivalent to following a search link or other offsite link, because the full site has to load). You can also try refreshing the page and opening devtools and doing a cache flush & hard reload—the latter is theoretically equivalent to hitting the site fresh from an outside link.

-  Checking the landing page, blog pages and API reference pages. 

-  Make sure the Quick Tour samples still work correctly.

- Try entering search terms and changing the search filters. Then try reloading the search page and make sure the search stays the same. (For example, the URL https://www.polymer-project.org/search/?q=data%20binding+more:2.0 should show hits for data binding in the 2.0 docs. You want that to still work if someone copy-pastes that URL into slack, for example.)

katejeffreys:
Tested per above comments. LGTM.

PR #1944, opened 2017-03-12
@apply with parentheses is deprecated

rictic:
Noticed by Chad in `#2.0-preview` on slack

### DIFF
--- a/app/1.0/docs/devguide/styling.md
+++ b/app/1.0/docs/devguide/styling.md
@@ -273,7 +273,7 @@ proposal.
 
 Use `@apply` to apply a mixin:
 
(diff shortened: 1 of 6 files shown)


arthurevans:
Does this form work in 1.0?

rictic:
Yes, 1.0 works both with and without parens. 2.0 (at least at one point) only works without parens.

rictic:
69fa975 LGTM assuming that the `{.alert .alert-info}` is something that is handled specially by the docs project.

arthurevans:
Yeah, it gives us little alert/note boxes:

image

PR #1948, opened 2017-03-13
fixing a typo in devguide for custom-elements (Element upgrades)

elvomka:
There's a minor typo in the current documentation in the section for "Element upgrades".


elvomka:
Why do I need to create a google account to sign the CLA? That's stupid.

jsilvermist:
That's not a typo, the element is means your custom elements identifier, like `MyElement.is`.

The change I might want to make would be to surround `is` with backticks to make this more clear however.

elvomka:
@jsilvermist The visual mark would indeed be helpful.

PR #1950, opened 2017-03-13
Fix first editor tab name

stalina:
In the first polymer example, the tab is named "proto-element.html", but is referenced as custom-element.html in the index file (so as in the plunker), it has been changed to custom-element.html


arthurevans:
Thanks for the fix! Another community member submitted this as https://github.com/Polymer/docs/pull/1945 (now merged) so I'm closing this one.

PR #1953, opened 2017-03-14
Add backticks to "is" for better clarity

jsilvermist:
Add backticks to "is" to make clearer that "is" refers to a property (eg. `MyElement.is`), and is not just a grammatical error.

Prevent confusion such as in #1948.


arthurevans:
Actually, that was a typo and the original PR was correct. It should say, "the element has a valid custom element name"

arthurevans:
Fixed in https://github.com/Polymer/docs/pull/1984/commits/28c7d5545f2dbad964c75798c22f8b15df7ce127. Thanks for bringing this to my attention!

PR #1954, opened 2017-03-15
Fix broken links

arjunyel:
Fixed links that were broken because of folder structure.

The paper-checkbox link wasn't broken, just outdated :laughing:


arjunyel:
I signed it!

PR #1966, opened 2017-03-17
Add call to super on ready function

pomber:


arthurevans:
Thanks for the fix. You're quite right there should be a `super` call there, but I think the inclusion of this `ready` callback, which has nothing to do with properties, was an oversight, so I'm going to cut it. Thanks for bringing it to my attention!

PR #1973, opened 2017-03-23
Changed a MD table to HTML in upgrade.md

luis-puhl:
The markdown table is not rendering to html so I've changed it o an html table.
This way agrees with the file [`about_20.md`](https://github.com/Polymer/docs/blob/master/app/2.0/docs/about_20.md).

screenshot 9

It may be some deeper markdown error but I don't even know were to look (where is this markdown rendered into html?)

Also, no tests where made, except I got the following result when replacing the text in the browser myself:

screenshot 10

### DIFF
--- a/app/2.0/docs/upgrade.md
+++ b/app/2.0/docs/upgrade.md
@@ -79,12 +79,32 @@ break the existing, 1.x version of your element or app.
 
 Update the Polymer version in `bower.json` to the latest RC version.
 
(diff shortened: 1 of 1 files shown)


arthurevans:
This table should be fixed  by https://github.com/Polymer/docs/commit/c808e355fecf5e821e69dac5e62c7b2507e570ea, which has been merged but not yet deployed.

PR #1986, opened 2017-03-30
Fix Typo

billy-hardy:


arthurevans:
Hm. This should clearly be reworded, but the original intention here was that the _change_ data—that is, the new value or change record passed to the observer—might be out of date (that is, not match the current value of the property). Would probably be clearer to simply say, "... the parameters passed to the observer may not match the element's current property values."

PR #1988, opened 2017-03-30
Fix typos

tirithen:


tirithen:
I signed it!

arthurevans:
These docs are generated, so this should be fixed by this PR: https://github.com/Polymer/polymer/pull/4469

PR #1995, opened 2017-04-03
Update style-shadow-dom.md

baocang:
change `dom- module` to `dom-module` in `Use external stylesheets` secrtion


jsilvermist:
I believe that was split because the line was over 80 characters, in which case it might be better to push the entire tag to the next line.

baocang:
@jsilvermist  I wrap the entire to the next line, but I see there a lot of lines longer than 80 characters and without hart wrap.  80 characters is too short for wide screen, [GitHub](https://github.com) online editor also provide about 126 characters per line. and most editor preferred soft wrap of 80 characters, not hard wrap. so mostly I likes coding the documentation text、comments without hard wrap, and setting up code editor soft wrap preferred.

PR #2004, opened 2017-04-10
correct nav path for news case study

ddx:
Nav on Polymer App Toolbox page is linking to v1.0 news case study page instead of the v2.0 version.

See here:
https://www.polymer-project.org/2.0/toolbox/index


ddx:
I signed it!

arthurevans (approved):
LGTM, thanks!

PR #2006, opened 2017-04-11
Update index.html

renfeng:


ghost (commented):
Thanks @renfeng, it is important for us to be clear on the best ways for Polymer developers to interact with the team.

Alternative ways to address this:

* We could remove the link to the Google Plus Community if we no longer plan to use it, or 

* We could add a statement that it contains useful links to concepts and discussions from a couple of years ago and reinforce that the Slack channel is the best place for real-time interactions and questions.

@wmginsberg What do you think?

ghost (approved):

PR #2021, opened 2017-04-13
Add reference to new 'addPushManifest' build option

FredKSchott:
Ready for review, but shouldn't be released until the next CLI release is out.


arthurevans:
0.18.0 is out, so merging.

PR #2022, opened 2017-04-13
Update install guide and remove build instructions from it

katejeffreys:


app/2.0/start/install-2-0.md
```diff
@@ -4,107 +4,96 @@ title: Install Polymer 2.x
...
+1. Install Bower.
+
+    ```bash
+    npm install -g bower
+    ```
+
+2. Install the Polymer CLI
 
     ```bash
     npm install -g polymer-cli@next
```
arthurevans: I think we won't need the @next by the time we deploy these docs, so we should take a pass and remove them all.

PR #2025, opened 2017-04-14
Document the new implementation of `polymer lint`

rictic:
/cc @justinfagnani 

The new implementation will be released on Monday with the push of v0.18.0 to the `latest` npm tag.

### DIFF
--- a/app/1.0/docs/tools/polymer-cli.md
+++ b/app/1.0/docs/tools/polymer-cli.md
@@ -275,19 +275,44 @@ Open up a page other than the default `index.html` in a specific browser
 
     polymer serve --open app.html --browser Safari
 
-### Lint element(s) {#lint}
+### Lint {#lint}
 
-Check elements in your element project or app project for syntax errors, anti-patterns and more.
+Analyze your project for syntax errors, missing imports, bad databinding expressions and more.
 
-Element project example:
+Use it like so:
 
-    polymer lint --input my-el.html
+    polymer lint --rules=polymer-1
 
-App project example (linting multiple elements):
+This will lint all of the code in your project with the `polymer-1` ruleset, which is appropriate for projects using Polymer 1.0. If you're upgrading to Polymer 2.0, then the `polymer-2-hybrid` or `polymer-2` are a better choice, as they'll warn you about use of deprecated Polymer 1.0 features.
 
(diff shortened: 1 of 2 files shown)


app/2.0/docs/tools/polymer-cli.md
```diff
@@ -303,20 +303,43 @@ Open up a page other than the default `index.html` in a specific browser
 
 ### `polymer lint` {#lint}
 
-Lints element(s)..
+Analyze your project for syntax errors, missing imports, bad databinding expressions and more.
```
arthurevans: Is it worth adding a note here about what `polymer lint` _doesn't_ do? 

As I understand it, it doesn't catch general JavaScript issues, so people might want to supplement it with a general-purpose linter like ESLint or JSHint.

app/1.0/docs/tools/polymer-cli.md
```diff
@@ -275,19 +275,44 @@ Open up a page other than the default `index.html` in a specific browser
...
+The output of `polymer lint` looks like this:
+
+```
+            <iron-collapse>
+            ~~~~~~~~~~~~~~~
+
+index.html(83,12) warning [undefined-elements] - The element iron-collapse is not defined
+```
+
+This means that on line 83 of `index.html` there's an `<iron-collapse>` tag, but the linter can't find the definition of the `iron-collapse` custom element. This probably means that there's a missing HTML import in `index.html`, but if we wanted to ignore this warning, we'd add `"undefined-elements"` to the `"ignoreWarnings"` array in `polymer.json`.
```
arthurevans: Suggest breaking the last sentence and writing in 2nd person for consistency:

in `index.html`. To ignore this warning, add `"undefined-elements"` to the `"ignoreWarnings"` array in `polymer.json`.

arthurevans:
A few small suggestions otherwise LGTM.

app/1.0/docs/tools/polymer-cli.md
rictic: Done!

app/2.0/docs/tools/polymer-cli.md
rictic: +1, added some more text for clarity

rictic:
Comments addressed, please take a look!

PR #2027, opened 2017-04-17
Upgrade guide fixes & related content fixes.

arthurevans:


app/2.0/docs/devguide/events.md
```diff
@@ -132,7 +132,7 @@ shadow DOM boundaries, set the `composed` flag to true when you create the event
 var event = new CustomEvent('my-event', {bubbles: true, composed: true});
```
kevinpschaaf: Maybe worth noting that by default events also don't bubble (`fire` made them bubble by default).  So a note similar to the `composed` flag for `bubbles` would be good.

PR #2028, opened 2017-04-18
New build guide

katejeffreys:


app/2.0/docs/tools/build-for-production.md
```diff
@@ -0,0 +1,211 @@
...
+* [Compiling ES6 to ES5](#compiling)
+* [Bundling resources to reduce the total number of HTTP requests made by the user's browser](#bundling)
+
+### Splitting inline JavaScript and CSS from HTML files {#splitting}
+  
+  By default, the Polymer CLI build process uses a built in feature of `polymer-build` to split inline JavaScript and CSS from HTML files. The build process automatically splits these streams in order to process them appropriately - for example, for minification, compilation or both.
+
+  By default, the Polymer build tools rejoin JavaScript and HTML after processing.
+
+  To split your code from HTML in order to deploy into an environment that requires this, you can use [Crisper](https://github.com/PolymerLabs/crisper).
```
katejeffreys: @arthurevans What info from the old guide do you think should be included in the new build-for-production guide in the section on "Splitting inline JavaScript and CSS from HTML files"?

* Description of CSP and what it does
* How to install and run crisper
* Examples of Crisper command line use
* Examples of Crisper use with Gulp
* All of the above?
* Something different?

The 1.0 docs mention piping Vulcanize output to crisper. Vulcanize has been replaced by the bundler, Y/n? https://github.com/Polymer/polymer-build#projectbundler

Either way, this is confusing to me. Why would the user bundle their resources and then split them again?

@justinfagnani @FredKSchott could you help me out with an example use case?

app/2.0/docs/tools/build-for-production.md
```diff
@@ -0,0 +1,211 @@
...
+
+Your builds are output to two separate folders, corresponding to their names: `build/bundled` and `build/unbundled`. 
+
+## Building with `polymer-build` {#use-polymer-build}
+
+Consider using `polymer-build` instead of the CLI if you:
+
+* Want to customize your build(s) without using the Polymer CLI
+* Need to run your source code through custom optimizers/processors before, after, or during your build
+* Need to hook additional work into any part of the build process
```
katejeffreys: For now, I've used https://github.com/PolymerElements/generator-polymer-init-custom-build/blob/master/generators/app/gulpfile.js. If I can get a better example, will update.

katejeffreys:
@arthurevans This is ready for another look.

app/2.0/docs/tools/build-for-production.md
```diff
@@ -0,0 +1,222 @@
...
+```
+This flag adds the `custom-elements-es5-adapter.js` adapter for running ES5 code on browsers that support ES6.
+
+If you’re unsure what the best strategy is for your project, here’s a quick overview:
+
+|   | Easiest for cross-browser support  | Most optimal for WC v1 performance  |
+|---|-------|------|
+| **Server** | Any server works, including static ones | Differential serving required |
+| **Deployed Code** | ES5 transpiled | ES6|
+| **Polyfill Loader** | custom-elements-es5-adapter.js | webcomponents-loader.js|
```
arthurevans: This table still isn't right. And I think it's going to be hard to record in exactly this format.

Can probably shorten "Most optimal ..." to just "Best performance." This should perform better on all browsers that support ES6, not just ones with native CE (because the native ES6 will in most cases run faster than the transpiled version).

First row, "differential serving => dynamic serving"?
For deployed code, the first column should be "Single build, ES5 transpiled" (Or just ES5 like you have now), while the other should say, "Two builds, ES5 and ES6"

Polyfill loader: either drop this row or rewrite it as "Polyfills" ... Column one should be custom-elements-es5-adapter.js + webcomponents-loader.js and column two stays the same. (Really, it is either webcomponents-loader.js or webcomponents-lite.js. I suspect we're going to be pushing the latter as more production-friendly.

app/2.0/docs/tools/build-for-production.md
```diff
@@ -0,0 +1,222 @@
...
+* Edge 15.15063 or later.
+* Firefox 51 or later.
+
+You should serve compiled ES5 only to older browsers that don't support ES6.
+
+If you need to statically host your code and serve a single version to all browsers, however, you should compile all code to ES5. In this case, you can include a shim - a lightweight polyfill that lets compiled ES5 work on browsers that support native custom elements.
+  
+```bash
+polymer build --js-compile 
+```
```
arthurevans: Show just flag and equivalent polymer.json snippet here, like with minification?

app/2.0/nav.yaml
```diff
@@ -111,13 +111,16 @@
...
-  - title: Optimize for production
-    path: /2.0/docs/tools/optimize-for-production
+  - title: Build for production
+    path: /2.0/docs/tools/build-for-production
     indent: True
 #  - title: Publish an element
 #    path: /2.0/docs/tools/reusable-elements
 #    indent: True
-  - title: Services
+  - title: Advanced tools
```
arthurevans: Remove "Advanced tools" section—this is long gone.

app/2.0/docs/tools/build-for-production.md
```diff
@@ -0,0 +1,222 @@
...
+The Polymer CLI and `polymer-build` library support the following transforms:
+
+* [Splitting inline JavaScript and CSS from HTML files](#splitting)
+* [Minifying HTML, JavaScript and CSS](#minifying)
+* [Compiling ES6 to ES5](#compiling)
+* [Bundling resources to reduce the total number of HTTP requests made by the user's browser](#bundling)
+
+### Splitting inline JavaScript and CSS from HTML files {#splitting}
+  
+By default, the Polymer CLI build process uses a built in feature of `polymer-build` to split inline JavaScript and CSS from HTML files. The build process automatically splits these streams in order to process them appropriately - for example, for minification, compilation or both.
```
katejeffreys: I have left this out for now. It looks like the most helpful thing to do would be to include a section on CSP per the old docs. I created https://github.com/Polymer/docs/issues/2101 to track this.

Issue #1909, opened 2017-03-04
Add rootPath and importPath to DOM templates chapter


arthurevans:
https://github.com/Polymer/polymer/issues/4357

katejeffreys:
https://www.polymer-project.org/2.0/docs/devguide/dom-template
Contains info on rootPath and importPath
Closing.

Issue #1462, opened 2015-12-11
Documentation for `preserve-content`?


arthurevans:
_From @pkaske on August 22, 2015 13:0_

I scratched my head for quite some time till I found out about the `preserve-content` attribute on  templates to keep them untouched by polymer.
Is this somewhere in the docs? I couldn't find it. If it's missing, it's maybe something worth mentioning :wink: 

_Copied from original issue: Polymer/polymer#2346_

arthurevans:
_From @Sleeckx on August 26, 2015 14:45_

+1

Would like to know if this feature is here to stay as well.

arthurevans:
@kevinpschaaf Is this OK to document? If so, going to move it to the docs repo.

arthurevans:
_From @dlasky on December 10, 2015 22:42_

we would like to use this as well if its safe to do so

arthurevans:
_From @kevinpschaaf on December 11, 2015 1:1_

Yep, let's document.

Adding `preserve-content` to a template nested within a Polymer element's top-level template will:
* prevent bindings from being parsed (and removed) from the template
* prevent special handling of nested template's `content` fragment; specifically, the `content` fragment of a nested template is normally cached by Polymer and cleared, to be handed back to the template (as `_content`) prior to `ready` time.  This is done for a number of reasons including performance (avoids the cost of unnecessarily cloning elements in nested template contents) and to avoid `HTMLTemplateElement` bugs on certain platform/versions (older Chrome versions upgraded elements in nested templates when cloning outer templates, and Safari failed to deep-clone nested template contents).

mgiuffrida:
:+1: for documentation

arthurevans:
Was not sure where to put this. I guess we can add it to the local DOM doc. Maybe here https://www.polymer-project.org/1.0/docs/devguide/local-dom.html#template-stamping or more likely as a new section near the end of the doc.

mgiuffrida:
@arthurevans That's the section I checked first. It says:

> Polymer will automatically clone this template’s contents into the element’s local DOM.

which is only 100% true if `preserve-content` is specified, so that seems like an appropriate place to document it.

erichmx:
Definitely worth documenting, it took me 3 hours to finally find it in https://github.com/Polymer/polymer/blob/2.0-preview/src/template/annotations.html

bedeoverend:
[...]

PR #2035, opened 2017-04-19
Template updates.

### LINKED ISSUE #2019: Update DOM templating doc
- [ ] Add section on handling URLs in templates.
- [ ] Add reference to stamping in light DOM.

arthurevans:
Updates to the DOM template chapter. Fixes #2019, #1909, #1462.
Staged at: https://2-0-staging-dot-polymer-project.appspot.com/2.0/docs/devguide/dom-template

### DIFF
--- a/app/2.0/docs/devguide/dom-template.md
+++ b/app/2.0/docs/devguide/dom-template.md
@@ -93,7 +93,7 @@ template element. If you're going to modify the superclass template, there are a
 important rules:
 
 -   Don't modify the superclass template in place; make a copy before modifying.
--   If you're doing anything expensive, like copying or modifying an exisitng the template,
+-   If you're doing anything expensive, like copying or modifying an existing template,
     you should memoize the modified template so you don't have to regenerate it when the
     getter is called.
 
@@ -120,7 +120,7 @@ The following example shows a simple modification of a parent template:
 })();
 ```
 
-The following example shows a more complicated could wrap a superclass template with a child
+The following example shows how an element could wrap a superclass template with its own
 template.
 
 ```html
--- a/app/2.0/docs/devguide/dom-template.md
+++ b/app/2.0/docs/devguide/dom-template.md
@@ -170,10 +170,15 @@ define a `template` getter that returns a falsy value.
 
 ### URLs in templates {#urls-in-templates}
 
-A relative URL in a template may be relative to an application or to a specific component. For
-example, if a component includes images alongside an HTML import that defines an element, the image
-URL needs to be resolved relative to the import. However, an application-specific element may need
-to include links to URLs relative to the main document.
+A relative URL in a template may need to be relative to an application or to a specific component.
+For example, if a component includes images alongside an HTML import that defines an element, the
+image URL needs to be resolved relative to the import. However, an application-specific element may
+need to include links to URLs relative to the main document.
+
+By default, Polymer **does not modify URLs in templates**, so all relative URLs are treated as
+relative to  the main document URL. This is because when the template content is cloned and added
+to the main document, the browser evaluates the URLs  relative to the document (not to the original
+location of the template).
 
 To ensure URLs resolve properly, Polymer provides two properties that can be used in data bindings:
 
@@ -248,15 +253,24 @@ use the standard DOM `querySelector`  method.
 
 
 Add the `strip-whitespace` boolean attribute to a template to remove
-any empty text nodes from the template's contents. This can result in a
+any **empty** text nodes from the template's contents. This can result in a
 minor performance improvement.
+
+**What's an empty node?** `strip-whitespace` removes only text nodes that occur between
+elements in the template and are _empty_ (that is, they only contain whitespace characters).
+These nodes are created when two elements in the template are separated by whitespace (such as
+spaces or line breaks). It doesn't remove any whitespace from inside elements.
+{.alert .alert-info}
 
 With empty text nodes:
 
 
 ```html
 <dom-module id="has-whitespace">
-  <template> <div>A</div> <div>B</div> </template>
+  <template>
+    <div>Some Text</div>
+    <div>More Text</div>
+  </template>
   <script>
     class HasWhitespace extends Polymer.Element {
       static get is() { return  'has-whitespace' }
@@ -270,6 +284,14 @@ With empty text nodes:
 </dom-module>
 ```
 
+There are five nodes in this element's shadow tree because of the whitespace surrounding the `<div>`
+elements. The five child nodes are:
+
+  text node
+  `<div>Some Text</div>`
+  text node
+  `<div>More Text</div>`
+  text node
 
 Without empty text nodes:
 
@@ -292,6 +314,12 @@ Without empty text nodes:
   </script>
 </dom-module>
 ```
+
+Here, the shadow tree contains only the two `<div>` nodes:
+
+`<div>Some Text</div><div>More Text</div>`
+
+Note that the whitespace _inside_ the `<div>` elements isn't affected.
 
 ## Preserve template contents
 
@@ -377,7 +405,8 @@ _attachDom(dom) {
 ```
 
 When you stamp the DOM template to light DOM like this, data bindings and declarative event listeners
-work as usual, but you cannot use shadow DOM features, like `<slot>` and style encapsulation. A
-template stamped into light DOM shouldn't contain any `<style>` tags. Styles can be applied by an
+work as usual, but you cannot use shadow DOM features, like `<slot>` and style encapsulation.
+
+A template stamped into light DOM shouldn't contain any `<style>` tags. Styles can be applied by an
 enclosing host element, or at the document level if the element isn't used inside another element's
 shadow DOM.
--- a/app/2.0/docs/devguide/dom-template.md
+++ b/app/2.0/docs/devguide/dom-template.md
@@ -97,7 +97,7 @@ important rules:
     you should memoize the modified template so you don't have to regenerate it when the
     getter is called.
 
-The following example shows a simple modification of a parent template:
+The following example shows a simple modification based on a parent template:
 
 ```js
 (function() {
@@ -106,8 +106,8 @@ The following example shows a simple modification of a parent template:
   class MyExtension extends MySuperClass {
     static get template() {
       if (!memoizedTemplate) {
-        // retrieve superclass template
-        memoizedTemplate = MySuperClass.template;
+        // create a clone of superclass template (`true` = "deep" clone)
+        memoizedTemplate = MySuperClass.template.cloneNode(true);
         // add a node to the template.
         let div = document.createElement('div');
         div.textContent = 'Hello from an extended template.'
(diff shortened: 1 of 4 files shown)


app/2.0/docs/devguide/dom-template.md
```diff
@@ -10,7 +10,7 @@ By default, adding a DOM template to an element causes Polymer to create a shado
 
 The DOM template is also processed to enable data binding and declarative event handlers.
 
-## DOM template overview
+## Specify a DOM template
 
 To specify an element's DOM template:
```
kevinpschaaf: There's really a step (3) that goes along with 1 & 2, which is

> 3. Add a `static get is() { return 'element-name'; }` getter to your element's class that identifies the `dom-module` id to get the template from.

However, it's probably worth noting that we now support 3 ways of specifying a template:

1. `static get is()` that returns id of a `dom-module` containing a `<template>` (basically, the normal legacy way, which should be the most efficient way to package a template using HTMLImports)
1. `static get template()` that returns an `HTMLTemplateElement` reference
1. `static get template()` that returns a string (which Polymer will parse into a new HTMLTemplateElement)

And note that if you're not using method (1), there's now no reason to include the `static get is()` -- its only purpose is to identify the dom-module containing a template, and that is no longer the compulsory way to get a template.

There's some detail in the 3rd bullet in the bulleted list under https://github.com/Polymer/polymer/blob/master/README.md#20-es6-class-based-syntax

---

Sorry, I see you cover (3) below... however it may make sense to lay out the 3 ways up front and then describe each one with an example maybe?

app/2.0/docs/devguide/dom-template.md
```diff
@@ -44,9 +44,81 @@ except for testing. For caveats about defining elements in the main document,
...
+or a string template), then no shadow root is created for the element, _unless_ the element is
+extending another element that defines a DOM template.
+
+### Inherited templates
+
+An element that extends another Polymer element can inherit its template. If the element doesn't
+provide its own DOM template (using either a `<dom-module>` or a string template), Polymer uses the
+same template as the superclass, if any.
+
+If the superclass has no template, the subclass element doesn't create a shadow root.
```
kevinpschaaf: It would be nice in this section to describe how to actually _extend_ templates now that we have inheritance (in addition to this content which just discusses inheriting templates).

So, the way an element gets its template is that there is a default `static get template()` getter that uses the `static get is()` getter if provided to find the template in the dom-module. As described above, one can now override `static get template()` to return an arbitrary template (or string, which is parsed into a template), rather than using `is`+`dom-module`.

So you can now do this kind of thing to augment a super's template if needed:

```js
// clone super's template
let subTemplate = MySuperClass.template.cloneNode(true);
// modify it as desired
subTemplate.appendChild(...);

class MySubClass extends MySuperClass {
  static get template() { return subTemplate; }
}
```

app/2.0/docs/devguide/dom-template.md
```diff
@@ -83,4 +155,138 @@ use the standard DOM `querySelector`  method.
 
 
 
+## Remove empty text nodes {#strip-whitespace}
+
+
+Add the `strip-whitespace` boolean attribute to a template to remove
+any empty text nodes from the template's contents. This can result in a
+minor performance improvement.
```
kevinpschaaf: We recently had a user confused that `strip-whitespace` wasn't actually stripping whitespace out of the _inside_ of text nodes... it only strips whitespace-only text nodes between other elements.  Might be worth mentioning?

app/2.0/docs/devguide/dom-template.md
```diff
@@ -83,4 +155,138 @@ use the standard DOM `querySelector`  method.
...
+      }
+    }
+    customElements.define(NoWhitespace.is, NoWhitespace);
+  </script>
+</dom-module>
+```
+
+## Preserve template contents
+
+Polymer performs one-time processing on any Polymer-managed templates, including templates nested
```
kevinpschaaf: Polymer performs one-time processing _(such as parsing and removing binding annotations and caching and removing nested template contents for better performance)_...

kevinpschaaf: That is, `preserve-content` does more than just avoiding the `.content` being cached & removed, it makes Polymer totally skip processing on it, so it won't get bindings, etc.

app/2.0/docs/devguide/dom-template.md
kevinpschaaf: (3) is useful when elements are not being defined via HTML imports, e.g. when using ES modules.

(2) may be useful in the module case, or also to extend/modify super templates, share templates, etc.

app/2.0/docs/devguide/dom-template.md
arthurevans: Cool. I wanted to put in something like this, but I just didn't have all the bits together. Hmm... 

Stupid question, though. Is there a reason you're doing the `let subTemplate ... ` etc work up front as opposed to inside the `template` getter? Just want to understand the logic. 

Note to self: we actually need `subTemplate.content.appendChild(...)`

app/2.0/docs/devguide/dom-template.md
arthurevans: Ah, yeah. Nice, will incorporate language.

app/2.0/docs/devguide/dom-template.md
```diff
@@ -38,15 +55,159 @@ Example: { .caption }
...
+An element that extends another Polymer element can inherit its template. If the element doesn't
+provide its own DOM template (using either a `<dom-module>` or a string template), Polymer uses the
+same template as the superclass, if any.
+
+You can also modify a superclass template by defining a `template` getter that returns a modified
+template element. If you're going to modify the superclass template, there are a couple of
+important rules:
+
+-   Don't modify the superclass template in place; make a copy before modifying.
+-   If you're doing anything expensive, like copying or modifying an exisitng the template,
```
bicknellr: 'an exisitng the template'

app/2.0/docs/devguide/dom-template.md
```diff
@@ -38,15 +55,159 @@ Example: { .caption }
...
+        memoizedTemplate.content.appendChild(div);
+      }
+      return memoizedTemplate;
+    }
+  }
+
+})();
+```
+
+The following example shows a more complicated could wrap a superclass template with a child
```
bicknellr: 'a more complicated could': maybe 'element' or 'class' was mean't to go after 'complicated'?

arthurevans: Hah! That'll teach me to submit stuff on Monday that I was editing on Friday 😊.

Perhaps the sentence is complicated enough without "complicated" and could just be: "The following example shows how an element could wrap a superclass template with a child template."

app/2.0/docs/devguide/dom-template.md
```diff
@@ -38,18 +55,125 @@ Example: { .caption }
...
+If the element is extending another element that has a DOM template and you don't want a DOM template,
+define a `template` getter that returns a falsy value.
+
 ### URLs in templates {#urls-in-templates}
 
-A URL in a template may be relative to an application or to a specific component. For example,
-if a component includes images alongside an HTML import that defines an element, the image URL needs
-to be resolved relative to the import. However, an application-specific element may need to include
-links to URLs relative to the main document.
+A relative URL in a template may be relative to an application or to a specific component. For
```
kevinpschaaf: A relative URL in a template may **_need to be_** be relative...

(might clarify slightly)

app/2.0/docs/devguide/dom-template.md
```diff
@@ -38,18 +55,125 @@ Example: { .caption }
...
+example, if a component includes images alongside an HTML import that defines an element, the image
+URL needs to be resolved relative to the import. However, an application-specific element may need
+to include links to URLs relative to the main document.
 
 To ensure URLs resolve properly, Polymer provides two properties that can be used in data bindings:
```
kevinpschaaf: Maybe it would be good to say "By default, Polymer does not modify URL's in templates, so all relative URL's will be relative to the main document URL. This is because when the template content is cloned and added to the main document, the browser evaluates the URL's as relative to the document (not the original location of the template)."

app/2.0/docs/devguide/dom-template.md
```diff
@@ -83,4 +244,140 @@ use the standard DOM `querySelector`  method.
...
+  </script>
+</dom-module>
+```
+
+## Customize DOM initialization
+
+There are several points where you can customize how Polymer initializes your element's DOM. You
+can customize how the shadow root is created by creating it yourself. And you can override the
+`_attachDom` method to change how the the DOM tree is added to your element: for example, to
+stamp into light DOM instead of shadow DOM.
```
kevinpschaaf: Perhaps should note that when stamping into light DOM, styling those elements becomes a problem that the user has to solve, e.g. by placing a stylesheet in the body, etc.

arthurevans:
@katejeffreys would  you do a copyedit pass on this before I merge it? A bunch of new content in here, please let me know where it seems fuzzy or confusing.

usergenic:
Added a `cloneNode(true)` otherwise superclass template *is* modified.
```js
    static get template() {
      if (!memoizedTemplate) {
        // create a clone of superclass template (`true` = "deep" clone)
        memoizedTemplate = MySuperClass.template.cloneNode(true);
        // add a node to the template.
        let div = document.createElement('div');
        div.textContent = 'Hello from an extended template.'
        memoizedTemplate.content.appendChild(div);
      }
      return memoizedTemplate;
    }
```

usergenic:
Hmm now reading that section, it almost looks like the intent with that example was to show how NOT to do it, because then it's followed by an example that shows how you do clone ala `document.importNode()`.  So, I'm not clear as a reader about that.  If the first example shows modifying parent template and that's bad, then there should be a comment in the example code that says "See here, don't do this right here, because now the parent template is modified.  Bad!"

🐱

usergenic:
> Without empty text nodes:
>
> ```html
><dom-module id="no-whitespace">
>  <template strip-whitespace>
>    <div>A</div>
>    <div>B</div>
>  </template>
>  <script>
>    class NoWhitespace extends Polymer.Element {
>      static get is() { return  'no-whitespace' }
>      ready() {
>        super.ready();
>        console.log(this.shadowRoot.childNodes.length); // 2
>      }
>    }
>    customElements.define(NoWhitespace.is, NoWhitespace);
>  </script>
></dom-module>
> ```

I think in the above the divs should read like this instead, instead of "A" and "B":
```html
<template strip-whitespace>
  <div>Some Text</div>
  <div>More Text</div>
</template>
```

app/2.0/docs/devguide/dom-template.md
arthurevans: We mention that down below where we go into detail on this. 

https://2-0-staging-dot-polymer-project.appspot.com/2.0/docs/devguide/dom-template#stamp-templates-in-light-dom (last sentence on the page, actually)

Do you think we should also add that note in the intro?

katejeffreys:
LGTM.

PR #2038, opened 2017-04-19
Fix table formatting.

arthurevans:
@notwaldorf to veto or imma just gonna merge it.


arthurevans:
Checked all the tables I could find, and they look OK to me. So merging.

PR #2041, opened 2017-04-20
add a version label to search results

notwaldorf:
Looks like this. I don't think the search results are sorted by 2.0 first, which kinda sucks.

screen shot 2017-04-19 at 5 59 52

@keanulee @arthurevans

### DIFF
--- a/app/elements/pw-shell.html
+++ b/app/elements/pw-shell.html
@@ -458,9 +458,13 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       } else if (response.items) {
         section = resultsElement.querySelector('.yes-results');
 
+        // Clean up the version bit for easier using.
+        for (var item = 0; item < response.items.length; item++) {
+          response.items[item].version = response.items[item].labels[0].displayName;
+        }
+
         // Populate the results in the dom-repeat.
         section.querySelector('template').items = response.items;
-
         var queryPrefix = '?q=' + this._searchParams.q + '&start=';
         var button = section.querySelector('.next-results');
         if (response.queries.nextPage) {
--- a/app/elements/pw-shell.html
+++ b/app/elements/pw-shell.html
@@ -445,6 +445,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     },
 
     _handleSearchResponse: function(event) {
+      var i;  // the linter which may not be named has issues.
       var response = event.detail.response;
 
       // Hide the existing section being displayed.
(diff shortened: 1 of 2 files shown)


arthurevans:
OK, so here's how you can filter for _only_ a single label: add the string 'more:<label>' to the query. So if the user query is "data binding" and we want to show 2.0, q should be: `q=data+binding+more:2.0`... (Also we should be running the query though `encodeURIComponent` or something similar).

app/search/index.html
```diff
@@ -38,6 +38,26 @@
...
+
+      .search-result .version[data-version="1.0"] {
+        background-color: #37474f;
+      }
+
+      .search-result .version[data-version="2.0"] {
+        background-color: #1565c0;
+      }
+
+      .search-result .version[data-version="Blog"] {
```
keanulee: /s/version/label/, here and everwhere

notwaldorf: why? it's a version, no?

keanulee: "Blog" is not a version

notwaldorf: ok fair

notwaldorf:
@arthurevans @keanu. Added a filter for results. By default:
screen shot 2017-04-25 at 10 34 59

Only for 1.0:
screen shot 2017-04-25 at 10 35 05

app/elements/pw-shell.html
```diff
@@ -445,6 +448,7 @@
     },
 
     _handleSearchResponse: function(event) {
+      var i;  // the linter which may not be named has issues.
```
notwaldorf: A story: I tried to make the comment be "// jshint has issues with scoping" and then `jshint` tried to interpret "has issues with scoping" as weird config parameters and anyway, `jshint` is Voldemort now, this is our darkest timeline.

notwaldorf:
@arthurevans is this good to merge then?

app/elements/pw-shell.html
```diff
@@ -458,10 +462,41 @@
...
-        var queryPrefix = '?q=' + this._searchParams.q + '&start=';
+        // Restricted search buttons. Ugh, so. we use the (undocumented) +more:version
+        // search parameter, which means we need to uhhh remove the +more
+        // keyword by hand if it already exists.
+        var matches = this._searchParams.q.match(/(.*) more:(1\.0|2\.0|Blog)/);
+        var searchParams;
+        var activeLabel;
+        if (matches && matches.length > 2) {
+          searchParams = matches[1];
+          activeLabel = matches[2]
```
arthurevans: jshint is throwing a hissy about a missing semicolon here.

notwaldorf: oshit good find, thanks (mine isn't, but i also don't update my jshint out of fear)

arthurevans:
LGTM with one nit, when you select a link, it shows "Search results for test chai more:2.0"

Can we remove the "more:20" from the text?

notwaldorf:
@arthurevans donzo

arthurevans:
One more nitpick:

If I search for something that doesn't exist in a given category:

`/search/?q=chai+more:Blog`

The `more:` shows up on the "no results" page:

Search results for chai more:Blog
Your search — chai more:Blog — did not match any documents.

Also, ideally we'd include the filter links on this page so they can expand back to "all" or something.

notwaldorf:
@arthurevans done. it wasn't the greatest solution -- that page needs refactoring so that the common code (title + filters) is just in one place rather than in two, but it would take a while and maybe we want to do it in a separate PR

arthurevans:
Awesome! 🚀

PR #2043, opened 2017-04-21
Remove the gulp hack required for minifying bundles

keanulee:
Be sure to update your node_modules (`rm -rf node_modules; npm i`) before rebuilding


katejeffreys:
@keanulee can you suggest a reviewer for this?

templates/head-meta.html
```diff
@@ -21,5 +21,5 @@
   window.Polymer = window.Polymer || {dom: 'shadow', lazyRegister: true};
 </script>
 <!-- Node: the webcomponents polyfill is lazily loaded in app.js -->
-<link rel="import" href="/elements/pw-shell.html" async>
+<link rel="import" href="/elements/pw-shell.html">
```
keanulee: `async` was removed because the `custom-style` below uses modules defined in this import (for styling things in the light DOM).

keanulee:
@katejeffreys I would suggest @notwaldorf for code review, but you should also try rebuilding this branch and make sure it still works.

notwaldorf:
If this makes minified files which i think it does, then LGTM. (but if someone wants to double check by running it locally, it would be aces)

arthurevans:
@notwaldorf Tested, staged at https://remove-bundle-hack-dot-polymer-project.appspot.com/

As far as I can tell, minifies everything except for index.html and highlighter.js, which is the same as the current site. Pagespeed Insights gives similar speed scores to this version and the current site.

Issue #2012, opened 2017-04-11
Add/update deployment docs


arthurevans:
The "PRPL pattern" "App structure" and "Background: HTTP/2 and Server push" sections of Serve your app are relevant to both build & deployment.

This doc should probably be moved to the tools section. 

We also more concrete advice on how to serve bundled/unbundled resources.

PR #2047, opened 2017-04-21
Update text for 2.0, add images, change nav, move and rename file

katejeffreys:
https://github.com/Polymer/docs/issues/2012


app/2.0/docs/tools/deploy.md
```diff
@@ -0,0 +1,175 @@
+---
+title: Deploy your app
```
arthurevans: I think the title is probably off here, because the main subject of this doc is the PRPL pattern and the app shell structure that our build tools expect.

The "Build output" section is really discussing how the build chain bundles an app built with the PRPL app shell pattern.

Title suggestions? The PRPL pattern? Copying @kevinpschaaf @graynorton

arthurevans: For example, if we were to retitle this PRPL pattern, we could remove the first L2 head ("PRPL pattern") and remove the intro paragraphs leading up to it that talk about serving.

Then the "Serve your app" page could slim down to just the first few paragraphs here, along with some pointers to different pages in /tools/ and maybe a few notes on differential serving.

PR #2063, opened 2017-04-28
cleanup existing/out of date tooling documentation

FredKSchott:
For the Polymer Polish Party(tm) I went through the tooling documentation to look for anything that had gone out-of-date with our newest 1.0 tooling. I found and fixed everything out of date, clarified some confusing  bits, and moved some things to be less intimidating / easier to scan & read.

### Notable changes:
- `tools/overview.md`
  - Reword some wordy bits
- `tools/polymer-json.md`
  - Added link back to 'optimize-for-production.md' for builds config section
- **`tools/polymer-cli.md` *<-- biggest changes***
  - Currently: https://www.polymer-project.org/2.0/docs/tools/polymer-cli
  - Update deprecated installation instructions (git no longer needed, thank god)
  - Added an easy-to-grok overview to replace the pages of details about each command. The thinking is that those details are already found elsewhere on our docs site, the CLI repo README, and via the `polymer help COMMAND`. If more details are needed, pages like #2028 can focus on the actual use-case.
  - Focused instead on the more subtle assuptions that the CLI makes that aren't currently obvious in our docs: What app/element structures the CLI supports, relative vs. absolute paths, dependency paths, etc.
  - Added examples with info for element & app-shell applications
- `tools/services.md`
  - (Putting on my new user hat) I wasn't sure what "services" meant, so I renamed the section to "Web services" to be more clear that these are external web services.

---

@arthurevans @katejeffreys I'm happy to work you two to get this PR merged, but also won't be offended if you have your own plans for each of these docs. Even if that's the case hopefully there is something useful here you could use.

### DIFF
--- a/app/2.0/docs/tools/optimize-for-production.md
+++ b/app/2.0/docs/tools/optimize-for-production.md
@@ -4,13 +4,79 @@ title: Optimize for production
 
 <!-- toc -->
 
-[Polymer CLI](polymer-cli) is the recommended starting point for building
-Polymer applications. If for some reason it does not meet your needs, you
-can use the underlying libraries that power its build toolchain. See
-[Advanced tools](advanced#build) for a list of these tools.
+The Polymer CLI includes a `build` command to generate a production-ready build of your application. This process can include minifying code, compiling JavaScript, automatic service worker generation, and more.
 
-See the [polymer.json specification](polymer-json) for a reference to the available Polymer CLI
-build options.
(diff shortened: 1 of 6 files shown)


FredKSchott:
After the fact Arthur pointed me to https://github.com/Polymer/docs/pull/2028 which adds a "building for production" page. IMO that PR should supersede my `tools/optimize-for-production.md` rewrite: they both go over the same features but @katejeffreys's PR does a much better job at explaining why/when to use them.

katejeffreys:
Thanks @FredKSchott. I am including your updates to fix out-of-date info in https://github.com/Polymer/docs/pull/2028 so will close this PR.

Issue #1752, opened 2016-09-22
Update Deploy with Firebase with HTTP/2


abdonrd:
Announce: [HTTP/2 comes to Firebase Hosting](https://firebase.googleblog.com/2016/09/http2-comes-to-firebase-hosting.html)

In the docs: Build an app with App Toolbox > Step 4. Deploy > [Deploy with Firebase](https://www.polymer-project.org/1.0/start/toolbox/deploy#deploy-with-firebase)

Especially mention to Server Push for PRPL pattern.

cc/ @arthurevans @robdodson

abdonrd:
Friendly ping to @arthurevans and @robdodson. 🙂

katejeffreys:
Thanks for raising this, @abdonrd. To help me understand better, is the purpose of this issue to document the availability of HTTP/2 in Firebase in order to tell developers that they can use this to implement the PRPL pattern?

Would it be better to update this page instead, or as well? https://www.polymer-project.org/1.0/toolbox/server#background-http2-and-http2-server-push

abdonrd:
@katejeffreys exactly! How to setup a Polymer project with the PRPL pattern in Firebase using HTTP/2 and Server Push.

Maybe "Deploy with Firebase" is a better place (with any reference to "Background: HTTP/2 and HTTP/2 server push"). I say this to keep this page free of services.

johnlim:
Hi,
The latest docs seem to steer away from mentioning how to deploy to firebase using `unbundled` builds, using HTTP2 and the PRPL pattern. Is using `bundled` builds the official recommendation now or.....?

PR #2082, opened 2017-05-05
Address #1752

katejeffreys:
@abdonrd - Attempting to address https://github.com/Polymer/docs/issues/1752

### DIFF
--- a/app/2.0/start/toolbox/deploy.md
+++ b/app/2.0/start/toolbox/deploy.md
@@ -195,3 +195,10 @@ guide](https://www.firebase.com/docs/hosting/quickstart.html).
     The URL to your live site is listed in the output. You can also open
     the site in your default browser by running `firebase open hosting:site`.
 
+#### Optional: Edit your Firebase configuration to enable server push
+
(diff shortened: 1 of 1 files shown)


app/2.0/start/toolbox/deploy.md
```diff
@@ -195,3 +195,10 @@ guide](https://www.firebase.com/docs/hosting/quickstart.html).
     The URL to your live site is listed in the output. You can also open
     the site in your default browser by running `firebase open hosting:site`.
 
+#### Optional: Edit your Firebase configuration to enable server push
```
arthurevans: Can we verify whether the default build (or the default build when we launch) includes generating the push manifest? Might be that this should live in a different doc if it's more advanced (i.e., if the generated push manifest might need manual tweaking). Recommend we talk to the tools folks for a recommendation.

abdonrd:
The polymer-build now can generate a push-manifest.json file. Maybe can reference?

arthurevans:
I have concerns about putting this here, in a get-up-and-running tutorial, because it really is pretty advanced stuff. Let's hang on to this and see whether we can slot it into another section?

Ideally, we'd talk about generating the push manifest, editing the push manifest, and setting up push in the same place.

katejeffreys:
Ok - I will think about where to put it. Thanks for the feedback!

Issue #2080, opened 2017-05-05
`polymer serve` on Build an element tutorial repo gives the wrong URL


arthurevans:
_From @wmginsberg on April 28, 2017 18:5_

### Description
While doing the Build an element tutorial, I ran polymer serve and noticed that the URL suggested from the CLI was leading to a Not Found page. The URL sent out was http://127.0.0.1:8081/components/icon-toggle/  so I tried http://127.0.0.1:8081/components/icon-toggle/demo and it worked. Unsure if this is a CLI issue or a github repo/starter code issue. 

### Versions & Environment
- Polymer CLI: 0.18.1
- node: v6.4.0
- Operating System: macOS Sierra 10.12.4

#### Expected Results
I can see the demo code from the returned URL 

#### Actual Results
I receive a Not Found error page when trying to see demo code

_Copied from original issue: Polymer/polymer-cli#705_

arthurevans:
_From @wmginsberg on April 28, 2017 18:7_

There is a note in the tutorial later on but it's farther down and I didn't see it til I kept going 

"If this text doesn’t appear, make sure you’re looking at the the demo folder itself, or at demo/index.html, not at toggle-icon.html or toggle-icon-demo.html."

arthurevans:
_From @justinfagnani on May 3, 2017 22:10_

I think this should be a docs issue. We point to the root of the component folder, and yes, there's no guarantee that there's something loadable there.

arthurevans:
Hm. We could add an index or something that just redirects to the /demo/ directory, but I think that would be confusing since it's not a standard part of an element project.

Aha! How about if we add an `index.html` that mentions the name of the project, and tells people that they if they're doing the tutorial they probably want to be looking at the demo directory?

arthurevans:
Hey Kate. If you add an `index.html` file at the top-level of the project, you can add a link to the demo file, something like this: 

```
<html>
<head>
<title>Build your first element tutorial project</title>
<style>
  body { font-family: sans-serif; }
</style>
</head>
<body>
  <p>If you're testing your element, you probably want to
        look at the <a href="/components/icon-toggle/demo/">Demo</a>.</p>
        <p>You can find the tutorial associated with this project at
        <a href="https://www.polymer-project.org/2.0/start/first-element/intro">Build your
                                        first Polymer element on polymer-project.org.</a>
</body>
</html>
```

[...]

PR #2083, opened 2017-05-05
Address #2080

katejeffreys:
https://github.com/Polymer/docs/issues/2080
https://github.com/PolymerLabs/polymer-2-first-element/pull/4

### DIFF
--- a/app/2.0/start/first-element/intro.md
+++ b/app/2.0/start/first-element/intro.md
@@ -106,7 +106,7 @@ components.
     You’ll see some text where the icon toggles should appear. It doesn’t look
     very interesting, but it shows that everything is working.
  
-    (Note that the path uses `icon-toggle`—the component name listed in this element’s `bower.json` file—rather than the actual directory name. If you’re wondering why `polymer serve` does this, see [HTML imports and dependency management](/2.0/docs/tools/polymer-cli#element-project-layout).)
+    (Note that the URL includes `icon-toggle`—the component name listed in this element’s `bower.json` file—rather than the actual directory name. If you’re wondering why `polymer serve` does this, see [HTML imports and dependency management](/2.0/docs/tools/polymer-cli#element-project-layout).)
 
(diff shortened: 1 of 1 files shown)


app/2.0/start/first-element/intro.md
```diff
@@ -95,30 +96,15 @@ components.
...
 
-        polymer serve
-
-3.  `polymer serve` gives you two URLs to choose from - one for an application, another for 
-reusable components. Open the URL for **reusable components**. 
-
-    <img src="/images/2.0/first-element/polymer-serve-urls.png" alt="Output from the polymer serve 
-command, showing URLs" title="Local server URLs">
-
-    (Note that the path uses `icon-toggle`—the
```
arthurevans: I wonder if this note should be retained in some form. Although they don't need to type in this URL, it'll show up in the URL bar, and this URL confuses some people.

Maybe add as a note `{.alert .alert-info}` at the very end?

katejeffreys: Added it back in

app/2.0/start/first-element/intro.md
```diff
@@ -95,30 +96,17 @@ components.
...
 
     You’ll see some text where the icon toggles should appear. It doesn’t look
     very interesting, but it shows that everything is working.
-
-**If this text doesn’t appear**, make sure you’re looking at the the demo folder itself, or at 
-`demo/index.html`,
-not at `toggle-icon.html` or `demo-element.html`.
-{ .alert .alert-info }
+ 
+    (Note that the path uses `icon-toggle`—the component name listed in this element’s `bower.json` file—rather than the actual directory name. If you’re wondering why `polymer serve` does this, see [HTML imports and dependency management](/2.0/docs/tools/polymer-cli#element-project-layout).)
```
arthurevans: Since we don't show the path explicitly anymore—maybe instead of "the path uses", say, "the URL includes"

PR #2086, opened 2017-05-05
Upgrade feedback

arthurevans:
cc: @samuelli

### DIFF
--- a/app/2.0/docs/upgrade.md
+++ b/app/2.0/docs/upgrade.md
@@ -720,7 +720,7 @@ If your element doesn't do any imperative DOM manipulation, you can skip this se
 
 **Hybrid elements** need to keep using the `Polymer.dom` API. However, note that in 2.0, for native
 methods and properties that return a `NodeList,` **<code>Polymer.dom</code> APIs also return
-<code>NodeList</code>, not <code>Array</code>.</strong> If you're using any native
+<code>NodeList</code>, not <code>Array</code>.** If you're using any native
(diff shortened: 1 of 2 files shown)


app/2.0/docs/upgrade.md
```diff
@@ -72,48 +72,60 @@ npm update polymer-cli@next
...
+    | Component | Version |
+    |-----------|---------|
+    | Polymer   | `^2.0.0-rc.3` |
+    | webcomponentsjs | `^1.0.0-rc.7` |
+    | web-component-tester | `^6.0.0-prerelease.6` |
+    | Polymer elements | `2.0-preview` |
+
+    Note that Polymer, webcomponentsjs, and web-component-tester have prerelease tags. You can use
+    a range like `^2.0.0-rc.3` to get all future prerelease tags. To use prerelease elements,
+    you need to refer to the `2.0-preview` branch. To get  updates to elements, you may need to remove
```
katejeffreys: "To get updates to elements, you may need to remove your `bower_components` folder and reinstall all components"

We have instructed the user to do so, so might be able to remove this statement. idk, not too important.

arthurevans: With the preview elements, you'll need to delete the `bower_components` folder each time you update. For polymer itself (and other tagged stuff), you should only need to delete it once, after updating your `bower.json`.

We should be able to remove this whole paragraph when we release the elements. But I'll leave it in for now on the off chance that we do another docs push before then.

Issue #2078, opened 2017-05-05
Correct project template names in CLI doc


arthurevans:
From the polish party doc:

"Select `application`" -- there is no such option, should be `polymer-2-application`

Same for "element" project.

Same issue occurs in toolbox/templates...

katejeffreys:
See #2097 for fix

PR #2088, opened 2017-05-05
Correct names of element and app options in polymer init command

katejeffreys:
#2078


arthurevans:
Note we need the same changes in https://www.polymer-project.org/2.0/toolbox/templates

PR #2094, opened 2017-05-08
document build presets

FredKSchott:
We've added the ability to inherit build config from named presets. Here's all the info I think a user might need, @arthurevans @katejeffreys as always let me know if you'd like this re-organized / re-formatted / re-written in any way.

### DIFF
--- a/app/2.0/docs/tools/polymer-json.md
+++ b/app/2.0/docs/tools/polymer-json.md
@@ -135,6 +135,7 @@ for every run:
 
 * `name`: An optional name for your build. If multiple builds are defined, the `name` property is
 required.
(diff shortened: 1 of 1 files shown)


arthurevans:
Does the CLI support presets?

FredKSchott:
As of the next release it will

arthurevans:
🎈 :tada: 🎈  :tada: 🎈

FredKSchott:
I know!!! :party-parrot:

arthurevans:
LGTM2. @katejeffreys Why don't you go ahead and merge this and you can do any formatting/etc. you want to. 

Might be good to make a header for "Build presets," for instance. When the next CLI comes out, we can add the presets in the CLI doc and in the Build doc, too.

PR #2097, opened 2017-05-09
Fix names of templates

katejeffreys:


arthurevans:
1 more to add: in the templates doc, starter-kit is now polymer-2-starter-kit.

Shop is 1.x only (I think), so the template name is the same. News actually doesn't have a template, we should probably remove that one.

arthurevans:
Also (though less importantly) we should update the corresponding 1.x docs to reflect the naming in the latest CLI.

PR #2099, opened 2017-05-10
Spelling mistake

daneb:


daneb:
@katejeffreys apologies, I signed it, but didn't confirm with a reply here ;)

katejeffreys:
Oops, my bad. Thanks for the catch :) I have fixed it in another PR.

PR #2116, opened 2017-05-12
slot id --> slot name

katejeffreys:


app/2.0/docs/devguide/shadow-dom.md
```diff
@@ -199,7 +199,7 @@ example:
 ```
 <fancy-note>
   #shadow-root
-    <slot id="icon">
+    <slot name="icon">
       <img src="note.png">
     </slot>
     <slot></slot>
```
arthurevans: This one is good. The following ones are all supposed to be `id`.  They're default slots (no name) and we just use the ID so we can tell them apart.

PR #2128, opened 2017-05-15
use preset-env instead of preset-es2015

hzoo:
We recommend using env instead (the default is the same behavior) (I work on Babel)

### DIFF
--- a/app/2.0/docs/es6.md
+++ b/app/2.0/docs/es6.md
@@ -48,7 +48,7 @@ on browsers that support native custom elements.
 If you're putting together a custom build:
 
 -   Polymer requires all ES6 features except modules. (If you're using Babel, you can use
-    `babel-preset-es2015` with the `modules` option set to false.)
+    [`babel-preset-env`](https://github.com/babel/babel-preset-env) with the `modules` option set to false.)
 -   You must compile all your elements, the Polymer library, and any third-party elements you're
     using, but _not_ the polyfills.
 -   If you're serving compiled code to browsers that support native custom elements, inject the


abdonrd:
Ping to @justinfagnani :)

justinfagnani:
I don't quite have time to look at the differences here. That preset-env uses babel-preset-latest by default might be a problem...

hzoo:
preset-env currently = preset-es2015,16,17, just means you don't need to make a new entry for each year

katejeffreys:
Is there some testing that can be done to verify @hzoo's suggested alternative? I can do that if that would help.

justinfagnani:
The only issue is whether preset-env support features that aren't supported across all the evergreen browsers, and whether it will in the future. We're pretty conservative about what we support so that early-stage features are compiled out before hitting our toolchain.

It might be ok, I just don't have time to check it out until after I/O

hzoo:
The point of preset-env is that it adds whatever is in the latest spec. It doesn't support anything < Stage 4 and is basically the new "default" for Babel. The reasoning is that if you pass down your supported browsers it will not compile what is natively supported. And inherently all Stage plugins aren't implemented in browsers so preset-env won't include them

arthurevans:
OK, so making this change would require changing all of our other guidance on building and deploying Polymer apps—we currently recommend compiling for only a few browsers, and serving straight ES6 to the rest. I don't think that's a change we're going to make.

I think if we include this, we should include it as a note for advanced developers who want to use es2016, es2017 features. But if they do, it sounds like they're going to need to deal with making multiple builds and figuring out which ones to serve where.

Or am I missing something,  @justinfagnani?

hzoo:
Why do they need to create multiple builds? To be clear, this just allows for the compilation of additional features in es2016/es2017 (in addition to es2015) so not sure why this would require a huge change to the docs? Users are going to want to use those features anyway later, not just ES6.

justinfagnani:
I think we need to be clear about our goals for compilation: it is not to support language features based on the their status wrt to spec stages, it is to support the features that already implemented in our evergreen browsers, in older browsers like IE11.

So if it's not supported in Chrome, Safari, Edge and Firefox right now, we won't support it. If users want to use such features, they need to compile their source before it hits our tool chain. So, afaict, we really do need to care about which features browsers support.

hzoo:
Yeah that's the purpose of passing in the options into the preset, you can specify those browsers if you need to, otherwise it compiles it all down.

hzoo:
> So if it's not supported in Chrome, Safari, Edge and Firefox right now, we won't support it

If it's Stage 4 it should be supported, but I'l close if we don't want to do this

Issue #2071, opened 2017-05-02
Address issues with Polymer CLI docs (see #2063)


katejeffreys:
Address issues with Polymer CLI documentation raised by @FredKSchott 
in https://github.com/Polymer/docs/pull/2063

"...clearly highlight the added information about URLs and project structure... the current page is long and unwieldy which makes it hard to find that information when its most relevant. Maybe two separate "Element Project Structure" & "Application Project Structure" pages would do the trick?" -fks

PR #2130, opened 2017-05-16
Restructure polymer cli docs

katejeffreys:
Addresses #2071 in response to @FredKSchott's comments in #2063


katejeffreys:
Open to suggestions for other ways of doing this. Also needs a nav fix which I will do last.

app/2.0/docs/tools/create-app-polymer-cli.md
```diff
@@ -0,0 +1,102 @@
+---
+title: Create an application project with the Polymer CLI
+---
+
+<!-- toc -->
+
+Polymer CLI supports initializing a project folder with one of several application templates.  The 
+CLI comes with a `basic` template, which is the most basic starting point for a Polymer-based
```
FredKSchott: Confusing: Putting basic in ```s makes it feel like a proper name, but the actual term "basic" is never used by the CLI & there's no one "basic" template. 

Suggestion: "The CLI comes with a basic `polymer-2-app` template, which is the..."

app/2.0/docs/tools/create-app-polymer-cli.md
```diff
@@ -0,0 +1,102 @@
...
+Polymer CLI supports initializing a project folder with one of several application templates.  The 
+CLI comes with a `basic` template, which is the most basic starting point for a Polymer-based 
+application, as well as others that introduce more complex layout and application patterns.
+
+This chapter teaches you more about `basic` app projects.  See [Polymer App Toolbox 
+templates](../../toolbox/templates) for more details on other templates.
+
+## App project architecture {#app-architecture}
+
+Polymer CLI is designed for apps that follow the [app shell
```
FredKSchott: New user 🎩: Wait, Is it designed *only* for apps that follow the app shell architecture? If I'm not following that, i can't use the CLI?

app/2.0/docs/tools/create-app-polymer-cli.md
```diff
@@ -0,0 +1,102 @@
...
+templates](../../toolbox/templates) for more details on other templates.
+
+## App project architecture {#app-architecture}
+
+Polymer CLI is designed for apps that follow the [app shell 
+architecture](https://developers.google.com/web/updates/2015/11/app-shell).
+
+There are fundamental concepts of the app shell architecture that you should understand before 
+creating your app project with Polymer CLI: the entrypoint, the shell, and fragments. See [App 
+structure](../../toolbox/prpl#app-structure) from the App Toolbox docs for an in-depth overview of
```
FredKSchott: 🎩: I just read the developers.google.com link above, is this different? How? Do I need to read both?

app/2.0/docs/tools/create-app-polymer-cli.md
```diff
@@ -0,0 +1,102 @@
...
+architecture](https://developers.google.com/web/updates/2015/11/app-shell).
+
+There are fundamental concepts of the app shell architecture that you should understand before 
+creating your app project with Polymer CLI: the entrypoint, the shell, and fragments. See [App 
+structure](../../toolbox/prpl#app-structure) from the App Toolbox docs for an in-depth overview of 
+these concepts.
+
+## Set up basic app project {#basic-app}
+
+Follow the steps below to get your `basic` app project set up.
```
FredKSchott: Okay I think I see now that you're using "`basic`" as a name for the static, not-app-shell applications. This isn't clearly defined anywhere I've seen, if we want to use this we should explicitly describe what a `basic` app is somewhere here, or elsewhere that we can link to.

app/2.0/docs/tools/create-app-polymer-cli.md
```diff
@@ -0,0 +1,102 @@
...
+    Where <code><var>app</var></code> is the name of your project directory.
+
+1.  Initialize your app. Polymer CLI asks you a few questions
+    as it sets up your app.
+
+        polymer init
+
+1.  Select `polymer-2-application`.
+
+1.  Enter a name for your app. Defaults to the name of the current directory.
```
FredKSchott: do we need to describe each step here? These steps are subject to change through PRs, it would be unreasonable to expect our entire team to remember to come back here and update each time.

Suggestion: Replace these 3 steps with a summary "follow the prompts to create your app" step. The prompts are meant to be easy to follow for beginners and should contain enough info for the beginning user.

app/2.0/docs/tools/create-app-polymer-cli.md
```diff
@@ -0,0 +1,102 @@
...
+
+    The code samples throughout this doc use the example app element name
+    <code><var>my-app</var></code>. When creating your app you'll want to
+    replace any instance of <code><var>my-app</var></code> with the name of
+    your main element.
+    {.alert .alert-info}
+
+1.  Enter a description for your app.
+
+At this point, Polymer CLI generates files and directories for your app, and installs your
```
FredKSchott: Suggestion: "the Polymer CLI generates your app and installs its dependencies."

app/2.0/docs/tools/create-app-polymer-cli.md
```diff
@@ -0,0 +1,102 @@
...
+    {.alert .alert-info}
+
+1.  Enter a description for your app.
+
+At this point, Polymer CLI generates files and directories for your app, and installs your 
+project's dependencies.
+
+## App project layout
+
+After the initialization process Polymer CLI generates the following files and directories.
```
FredKSchott: Suggestion: "initialization process" feels a little 🤖

app/2.0/docs/tools/create-app-polymer-cli.md
```diff
@@ -0,0 +1,102 @@
...
+*   `bower_components/`. Project dependencies. See [Manage dependencies](#dependencies).
+*   `index.html`. Entrypoint page of the app.
+*   `src/`<code><var>my-app</var></code>/<code><var>my-app</var></code>`.html`.
+    Source code for main element.
+*    `test/`<code><var>my-app</var></code>/<code><var>my-app</var></code>`_test.html`. Tests for 
+main element.
+
+### Add elements
+
+You may want to compose your main element out of smaller, application-specific elements. These
```
FredKSchott: 🎩: What's a main element? Where is that? What's an application-specific element?

app/2.0/docs/tools/create-app-polymer-cli.md
```diff
@@ -0,0 +1,102 @@
...
+You may want to compose your main element out of smaller, application-specific elements. These 
+app-specific elements should be defined in the `src` directory, at the same level as 
+<code><var>my-app</var></code>.
+
+    app/
+      src/
+        my-app/
+        my-el/
+
+Currently there is no Polymer CLI command to generate application-specific elements. You should do
```
FredKSchott: No need to highlight a shortcoming of the CLI/experience if we don't need to :) 

Suggestion: Remove, or replace with something like "Polymer applications can be composed of many elements in your `src/` folder"

app/2.0/docs/tools/create-app-polymer-cli.md
```diff
@@ -0,0 +1,102 @@
...
+## Manage dependencies in an application project {#dependencies}
+
+Applications should use relative URLs to import other source files and dependencies. But because 
+applications are served independently, they can properly reach into the dependencies directory for 
+dependencies.
+
+    <!-- from a 'src/some-application.html' file in your application -->
+    <link rel="import" href="../bower_components/polymer/polymer.html">
+
+Only one file should use absolute URLs for imports: your "entrypoint" (defaults to "index.html").
```
FredKSchott: 🎩: wait, what defaults to "index.html"?

Suggestion: `(default: "index.html")`

app/2.0/docs/tools/create-element-polymer-cli.md
```diff
@@ -0,0 +1,121 @@
+---
+title: Create an element project with the Polymer CLI
+---
+
+<!-- toc -->
+
+This section shows you how to start an element project.
+
+1.  Create a directory for your element project. It's best practice for the name
```
FredKSchott: Not sure if the "It's best practice" line is necessary / overall directory naming strategies are probably out of scope for this doc. 

If you do keep it, then explain why its a best practice.

app/2.0/docs/tools/create-element-polymer-cli.md
```diff
@@ -0,0 +1,121 @@
...
+## Manage dependencies and HTML imports {#dependencies}
+
+### Use Bower to manage dependencies
+
+Polymer CLI uses [Bower](http://bower.io) for dependency management.
+
+Dependencies are stored in the `bower_components` directory. You should never manually alter the
+contents of this directory.
+
+Use the Bower CLI to manage dependencies.
```
FredKSchott: 🎩 what about that `polymer install` command?

app/2.0/docs/tools/create-element-polymer-cli.md
```diff
@@ -0,0 +1,121 @@
...
+*   `test/`<code><var>my-el</var></code>`_test.html`. Unit tests for
+    the element.
+
+## Manage dependencies and HTML imports {#dependencies}
+
+### Use Bower to manage dependencies
+
+Polymer CLI uses [Bower](http://bower.io) for dependency management.
+
+Dependencies are stored in the `bower_components` directory. You should never manually alter the
```
FredKSchott: How bower works is probably out of scope for this document as well.

app/2.0/docs/tools/create-element-polymer-cli.md
```diff
@@ -0,0 +1,121 @@
...
+contents of this directory.
+
+Use the Bower CLI to manage dependencies.
+
+This command downloads a dependency to `bower_components/` (the `--save` flag saves the new
```
FredKSchott: Which command?

(also, see above about Bower behavior probably being out of scope for this doc)

app/2.0/docs/tools/create-element-polymer-cli.md
```diff
@@ -0,0 +1,121 @@
...
+contents of this directory.
+
+Use the Bower CLI to manage dependencies.
+
+This command downloads a dependency to `bower_components/` (the `--save` flag saves the new 
+dependency to `bower.json`):
+
+    bower install --save PolymerElements/iron-ajax
+
+This command removes the dependency from `bower_components` and `bower.json`:
```
FredKSchott: Which command?

app/2.0/docs/tools/create-element-polymer-cli.md
```diff
@@ -0,0 +1,121 @@
...
+When importing dependencies inside your element, you should always use a relative URL  *as if the 
+dependency is a sibling of your project.*
+
+```
+<!-- from a top-level 'some-element.html' -->
+<link rel="import" href="../polymer/polymer.html">
+<link rel="import" href="../paper-elements/paper-button.html">
+```
+
+Using relative URLs may feel weird, but remember that dependencies are siblings of each other. So
```
FredKSchott: highlight exactly why it feels weird, in case a user doesn't know why its weird or if a user is feeling weird but unsure if its for the right reason.

Also, "remember that dependencies are siblings of each other *once installed*".

app/2.0/docs/tools/create-element-polymer-cli.md
```diff
@@ -0,0 +1,121 @@
...
+<link rel="import" href="../paper-elements/paper-button.html">
+```
+
+Using relative URLs may feel weird, but remember that dependencies are siblings of each other. So 
+write your paths as if the element is already being served as a dependency, and during development 
+the CLI will properly serve your project as a sibling of your dependencies.
+
+#### Why use relative URLs?
+
+Suppose that you ran Polymer CLI to generate an element project. Your element is named `my-el`. You
```
FredKSchott: You already explain this above. No need to explain in two places.

katejeffreys:
Thanks for the suggestions @FredKSchott! I have revised the pages and am happy to answer questions about any of the choices.

After review from @arthurevans this can be merged!

app/2.0/docs/tools/create-app-polymer-cli.md
katejeffreys: I agree that it would be unreasonable to expect the team to remember to update the CLI docs with these changes. I will maintain this documentation :)

For now I'm happy to leave these steps here with explanations for each. For instance, the user is asked to give a name to their "main element" and as you've noted below, the concept of a "main element" is not clear without an explanation.

app/2.0/docs/tools/create-app-polymer-cli.md
arthurevans: Yeah, this shouldn't be in backticks `basic` unless it's a literal string (like `polymer-2-application`).

app/2.0/docs/tools/create-app-polymer-cli.md
arthurevans: If we think this is going to be changing rapidly, we could move to just describing the "main element" concept up front, like, "Follow the prompts to create your app. Among other things, the CLI prompts your for the main element in your project..." 

But why don't we go with the steps for now and revisit if they end up changing frequently?

arthurevans:
Do you have this staged anywhere? Having trouble visualizing.

app/2.0/docs/tools/create-app-polymer-cli.md
```diff
@@ -0,0 +1,101 @@
...
+The `polymer-2-application` template is the most basic starting point for a Polymer-based 
+application. Other templates introduce more complex layout and application patterns.
+
+This chapter teaches you more about the `polymer-2-application` template.  
+See [Polymer App Toolbox templates](../../toolbox/templates) for more details on other templates.
+
+## App project architecture {#app-architecture}
+
+The `polymer-2-application` template follows the [app shell 
+architecture](https://developers.google.com/web/updates/2015/11/app-shell) pattern.
```
arthurevans: I think this whole section is incorrect, or at least misleading. The `polymer-2-application` template doesn't really have any structure at all. The PSK template follows the app shell architecture.

app/2.0/docs/tools/create-app-polymer-cli.md
arthurevans: I think we kind of need to highlight this to prevent people from trying to do init element inside an app project. But instead of saying, "There is currently no..." We could say,

To add another element to the project:

1. Create a new folder under src.

        mkdir src/my-el

2. Create an HTML import for the new element. You can use the existing app element as a starting point.

3. To use the new element, you'll need to import it into your application element.

 Don't use `polymer init` to create an element project inside your app project.

This adds more detail without being entirely satisfying (i.e., doesn't show adding the `<link rel="import">` or using the element in the app element). So alternately, just add the .html files to the diagram above for clarity:


```
my-example/
     src/
       my-example-app/
         my-example-app.html
       my-el/
         my-el.html
```

You can copy the existing app element as a starting point for your next element.

 Don't use `polymer init` to create an element project inside your app project.

=== end suggestion

I changed the example here to "my-example" so the directory names would make sense. If you're using `my-app` as a placeholder, should be `<var>my-app</var>` by style. Downside: examples might need to be formatted in HTML instead of fenced code blocks, because fenced code blocks escape the HTML markup.

app/2.0/docs/tools/create-element-polymer-cli.md
arthurevans: Since we're just about the only folks who use it, seems like it's probably OK. If you're comfortable with Bower and CLI tools, chances are you aren't even reading the "create an element project/app project" docs. I wouldn't want to give a full tutorial, but a few "don't stick your fork in this here socket" pointers seem apropos?

app/2.0/docs/tools/create-element-polymer-cli.md
arthurevans: This is a good question. Don't want to hang up on it too much here (we use bower directly other places), but we should discuss whether to move to that in general (or offer both options).

app/2.0/docs/tools/create-element-polymer-cli.md
arthurevans: Rather than "This command" form, maybe replace these with:

To add a new Bower dependency to your project:

    <bower command>

As Fred suggested, "This command..." is a little ambiguous about whether you're talking about the command above, or the command below.

app/2.0/docs/tools/polymer-cli-commands.md
```diff
@@ -0,0 +1,278 @@
...
+* [Global options for Polymer CLI commands](#global)
+
+## `polymer build` {#build}
+
+*This command is for app projects only.*
+
+Generates a production-ready build of your app. This process includes minifying the HTML, CSS, and
+JS of the application dependencies, and generating a service worker to pre-cache dependencies.
+
+Polymer CLI's build process is designed for apps that follow the [app shell
```
arthurevans: I think this link is a little confusing because some of our app structure is Polymer- or PRPL-specific (Like, the terms "entrypoint" and "fragment"). And that doc is full of a lot of implementation details that won't be helpful to the average Polymer user.

It would make more sense to link here to the PRPL doc, and have only that doc link to the app shell article for background on the app shell pattern.

app/2.0/docs/tools/polymer-cli.md
```diff
@@ -39,500 +39,28 @@ You're all set. Run `polymer help` to view a list of commands.
 
 ## Overview {#overview}
 
-Polymer CLI is a command-line interface for Polymer projects. It includes a build pipeline, a
-boilerplate generator for creating elements and apps, a linter, a development server, and a test
-runner.
+Polymer CLI is our official command line tool for Polymer projects and Web Components. It includes
```
arthurevans: our => the (see style/person)

app/2.0/docs/tools/polymer-cli.md
```diff
@@ -39,500 +39,28 @@ You're all set. Run `polymer help` to view a list of commands.
 
 ## Overview {#overview}
```
arthurevans: I know this organization is not new, but it strikes me as strange that the overview follows the install. 

Maybe move the first sentence to the head of the article (before the Install heading) and retitle this section "Types of projects" or "CLI project types".

arthurevans: And maybe `## Install` => `## Install Polymer CLI`?

Not sure. Do what feels right.

app/2.0/docs/tools/polymer-cli-commands.md
```diff
@@ -0,0 +1,278 @@
...
+
+* [`polymer build` (for app projects only)](#build)
+* [`polymer init`](#init)
+* [`polymer install`](#install)
+* [`polymer lint`](#lint)
+* [`polymer serve`](#serve)
+* [`polymer test`](#tests)
+* [Global options for Polymer CLI commands](#global)
+
+## `polymer build` {#build}
```
arthurevans: As discussed, remove code formatting in headers.

katejeffreys:
Thanks, updated with feedback.

katejeffreys:
Updated nav. Staged here:
https://restructure-cli-docs-dot-polymer-project.appspot.com/2.0/docs/devguide/feature-overview

app/2.0/docs/tools/create-app-polymer-cli.md
```diff
@@ -0,0 +1,124 @@
...
+
+Polymer CLI supports initializing a project folder with one of several application templates.  
+The `polymer-2-application` template is the most basic starting point for a Polymer-based 
+application. Other templates introduce more complex layout and application patterns.
+
+This chapter teaches you more about the `polymer-2-application` template.  
+See [Polymer App Toolbox templates](../../toolbox/templates) for more details on other templates.
+
+For a more full-featured progressive web app template, you can use the starter kit template 
+(`polymer-2-starter-kit`). See [/{{{polymer_version_dir}}}/start/toolbox/set-up](the Polymer Starter Kit tutorial)
```
arthurevans: lInk is backwards, should be `[text](url)`

app/2.0/docs/tools/create-app-polymer-cli.md
arthurevans: Still have the backticks.

app/2.0/docs/tools/create-app-polymer-cli.md
arthurevans: Still true.

app/2.0/docs/tools/create-app-polymer-cli.md
```diff
@@ -0,0 +1,124 @@
...
+*   `test/`<code><var>my-app</var></code>/<code><var>my-app</var></code>`_test.html`. Tests 
+    for main element.
+
+### Add elements
+
+You may want to compose your main element out of smaller elements specific to your app. These 
+application-specific elements should be defined in the `src` directory, at the same level as 
+<code><var>my-app</var></code>.
+
+<code>
```
arthurevans: Needs to be wrapped in a `<pre>` tag, otherwise the formatting gets lost.

app/2.0/docs/tools/create-app-polymer-cli.md
```diff
@@ -0,0 +1,124 @@
...
+          <var>my-app</var>.html
+        <var>my-el</var>/
+          <var>my-el</var>.html
+</code>
+
+To add another element to the project:
+
+1. Create a new folder under src.
+   
+<code>
```
arthurevans: This is getting outdented relative to the list. I _think_ it'll work OK if you intend this whole block four spaces.

app/2.0/docs/tools/create-app-polymer-cli.md
```diff
@@ -0,0 +1,124 @@
...
+   mkdir src/<var>my-el</var>
+</code>
+   
+2. Create an HTML import for the new element. You can use the existing app element as a starting point.
+   
+3. To use the new element, you'll need to import it into your application element (for example, <code><var>my-app</var>.html</code>) with an "import" link:
+   
+<code>
+   <link rel="import" href="/src/<var>my-el</var>/<var>my-el</var>.html">
```
arthurevans: Again, indent four spaces. Plus, you'll need to replace the `<` and `>` with entities: `&gt;` in place of `<` and `&lt;` in place of `>` ... The markdown blocks escape these automatically. When you do a `<code>` block yourself, you need to escape them or this gets interpreted as a literal `<link>` tag and it is hidden from the user.

app/2.0/docs/tools/create-app-polymer-cli.md
```diff
@@ -0,0 +1,124 @@
...
+## Manage dependencies in an application project {#dependencies}
+
+Applications should use relative URLs to import other source files and dependencies. But because 
+applications are served independently, they can properly reach into the dependencies directory for 
+dependencies.
+
+    <!-- from a 'src/some-application.html' file in your application -->
+    <link rel="import" href="../bower_components/polymer/polymer.html">
+
+Only one file should use absolute URLs for imports: your "entrypoint" (default: "index.html").
```
arthurevans: I think we can delete these two paragraphs (this and the following). Although the current basic app template uses absolute URLs, PSK now uses relative URLs in the entrypoint and I think we'll be going that way in the future. So we probably don't need to call this out.

app/2.0/nav.yaml
```diff
@@ -133,6 +133,15 @@
   - title: Polymer CLI
     path: /2.0/docs/tools/polymer-cli
     indent: True
+  - title: Polymer CLI commands
```
arthurevans: The left nav looks pretty cluttered with all of the "Polymer CLIs" in there... Maybe we could shorten these three to:
CLI commands
Create an element project
Create an app project

(Just in the left nav. Full titles in the docs are a-ok.)

app/2.0/docs/tools/create-app-polymer-cli.md
katejeffreys: Could you point out the paragraphs that need to be removed by their starting words, please? I can't disambiguate.

arthurevans: "Only one file should use absolute URLs for imports: your "entrypoint" (default: "index.html").  ... and the following code sample."

arthurevans: In other words, from L100 on.

app/2.0/docs/tools/polymer-cli.md
```diff
@@ -4,14 +4,18 @@ title: Polymer CLI
...
 <!-- toc -->
 
-## Install {#install}
+Polymer CLI is the official command line tool for Polymer projects and Web Components. It includes 
+a build pipeline, a boilerplate generator for creating elements and apps, a linter, a development
+server, and a test runner.
+
+## Install Polymer CLI {#install}
 
 1.  Make sure you have installed a verson of Node.js supported by Polymer. To check the version
```
arthurevans: <typo>

app/2.0/docs/tools/polymer-cli-commands.md
```diff
@@ -0,0 +1,277 @@
...
+
+## polymer lint {#lint}
+
+Analyze your project for syntax errors, missing imports, bad databinding expressions and more. `polymer lint` helps with identifying issues across your HTML, JS, and CSS based on an in-depth analysis of web components in source code. It does not reinvent the wheel though, it focuses on issues specific to web components and Polymer, so it is a good adjunct to other tools like [`eslint`](http://eslint.org/) and [`htmlhint`](http://htmlhint.com/).
+
+Use it like so:
+
+    polymer lint --rules=polymer-2
+
+This will lint all of the code in your project with the `polymer-2` ruleset, which is appropriate for projects using Polymer 2.0. If your code is hybrid and should with with either Polymer 1.x or 2.x then `polymer-2-hybrid` is a better choice, as it will warn you about use of features that do not exist in Polymer 2.x.
```
arthurevans: repeated word (with with)

PR #2134, opened 2017-05-16
Clarify "shady DOM"

MarcusJT:
Surely a typo? (Or does "shady" mean something else, in which case that needs to be explained or another choice of words used)


MarcusJT:
I signed it!

web-padawan:
Yes, it is [shady](https://github.com/webcomponents/shadydom).

PR #2143, opened 2017-05-18
ES6: var -> let

dandv:


jsilvermist:
@dandv In this context wouldn't `const` be more appropriate? The reference wouldn't need to change for the element, and there's no evidence the variable is being used for multiple items.

dandv:
@jsilvermist: Sure. Please squash :)

PR #2170, opened 2017-05-28
fix(first_element): fix documentation to use iron-icons.html

### LINKED ISSUE #2169: first element tutorial: iron-icon.html does not show the polymer icon
# Issue

I got stuck at this step where the polymer icon was not appearing:
```
<iron-icon icon="polymer"></iron-icon>
```

# Fix

iron-icon.html does not contain the definition for the polymer icon or the star icon used in step 2 and step 4. This should be changed to use iron-icons.html. Iron-icons has the following line that makes the icon available:

```
<g id="polymer"><path d="M19 4h-4L7.11 16.63 4.5 12 9 4H5L.5 12 5 20h4l7.89-12.63L19.5 12 15 20h4l4.5-8z"/></g>
```

So, this should be changed from:

```
<link rel="import" href="../iron-icon/iron-icon.html">
```

 to 
```
<link rel="import" href="../iron-icons/iron-icons.html">
```

cnishina:
iron-icon.html does not contain the definition for the polymer icon or
the star icon used in step 2 and step 4. This should be changed to use
iron-icons.html.

closes #2169

### DIFF
--- a/app/2.0/start/first-element/step-2.md
+++ b/app/2.0/start/first-element/step-2.md
@@ -32,7 +32,7 @@ Starting code—HTML imports { .caption }
 
 ```html
 <link rel="import" href="../polymer/polymer-element.html">
-<link rel="import" href="../iron-icon/iron-icon.html">
+<link rel="import" href="../iron-icons/iron-icons.html">
 ```
 
 Key information:
@@ -223,7 +223,7 @@ icon-toggle.html { .caption }
 
 ```html
 <link rel="import" href="../polymer/polymer-element.html">
-<link rel="import" href="../iron-icon/iron-icon.html">
+<link rel="import" href="../iron-icons/iron-icons.html">
 <dom-module id="icon-toggle">
   <template>
     <style>
--- a/app/2.0/start/first-element/step-4.md
+++ b/app/2.0/start/first-element/step-4.md
@@ -16,7 +16,7 @@ Import the `Polymer.GestureEventListeners` mixin by adding it to the HTML Import
 ```html
 <link rel="import" href="../polymer/polymer-element.html">
 <link rel="import" href="../polymer/lib/mixins/gesture-event-listeners.html">
-<link rel="import" href="../iron-icon/iron-icon.html">
+<link rel="import" href="../iron-icons/iron-icons.html">
 ```
 
 Use the mixin in the class declaration: 


arthurevans:
The element intentionally doesn't use `iron-icons`. The element shouldn't have any opinion about what icon it's going to display. `iron-icons` is a large component, and it may not contain the icon the user actually wants to use. Like `paper-icon-button`, this component uses `iron-icon`, and lets the user import or define whatever icon they want to use.

The demo element includes `iron-icons`, because it uses the star element:

https://github.com/PolymerElements/paper-icon-button/blob/master/paper-icon-button.html#L12

cnishina:
I understand that iron icons might be a really big component to include; however, for a first time user to Polymer, the starter documentation does not detail how to get the star or polymer icon. I think that the starter tutorial should just work based on the instructions given. Either it should include all the icons with iron-icons and note that this is a large set to include OR it should bring in the iron-icon.html and discuss how to bring in individual icons.

Since there was no discussion about where the polymer icon comes from, I spent more time than I wanted to figure out why this was not working. Since there was no discussion about this, I went through and repeated the steps to figure out where I did something wrong. 

The barrier of entry here might discourage people from using polymer. As a new user, you are asking them to understand how to bring in the polymer icon and star icon; however, from the tutorial it is not obvious.

cnishina:
Ok, I did not start from the "starter" git. I see that the iron-icons were included in the demo/demo-element.html. It might be nice just to note where the icons come from in the tutorial.

arthurevans:
Fair point. We could add a note about the difference between `iron-icon` and `iron-icons`, or perhaps touch on the icons when we walk through demo-element.  WDYT @katejeffreys?

cnishina:
This is totally my mistake. I thought it would be cool to take the Polymer CLI for a spin and created my own project. That's probably where I screwed up...although I think this is a to show off the CLI and have a project that starts from scratch.

katejeffreys:
The documentation touches on what's going on in Step 2 under "Create the local DOM structure":

>Key information:
>
>The <iron-icon> element is a custom element that renders an icon. Here it's hard-coded to use an icon named "polymer".

How about I link to iron-icon on webcomponents for more info? https://www.webcomponents.org/element/PolymerElements/iron-icon/elements/iron-icon

PR #2175, opened 2017-05-30
Link out to webocmponents.org

katejeffreys:


katejeffreys:
Fixes #2169

app/2.0/start/first-element/step-2.md
```diff
@@ -168,7 +168,7 @@ icon-toggle.html—after { .caption }
 Key information:
 
   * The `<iron-icon>` element is a custom element that renders an icon. Here it's hard-coded to use
-an icon named "polymer".
+an icon named "polymer". Find out more about `<iron-icon>` on [WebComponents.org](https://www.webcomponents.org/element/PolymerElements/iron-icon/elements/iron-icon).
```
arthurevans: I think the original confusion here stemmed from the fact that we're using an icon ("polymer") that's defined in `iron-icons`. But we don't actually import `iron-icons` here, we import it in the demo. In retrospect this is kind of a flaw in the tutorial flow, but I don't see an easy fix here. 

I'm not sure we can explain why this works without offering way too much information.

katejeffreys: Cool. Closing this PR.

PR #2176, opened 2017-05-31
Remove mention of the "next" version of Polymer CLI from upgrade.md

Haprog:
The command for updating Polymer CLI previously had the `@next` suffix, but since it's not used anymore, also remove the mention of the `next` version from the paragraph above the command.


Haprog:
I signed it!

PR #2182, opened 2017-06-01
Add resolutions for hybrid iron-doc-viewer.

aomarks:
https://github.com/PolymerElements/iron-doc-viewer/pull/134 updates the iron-doc-viewer bower config to be hybrid. The docs site needs a few resolutions.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -21,6 +21,8 @@
     "firebase-element": "PolymerElements/firebase-element#^1.0.12"
   },
   "resolutions": {
+    "iron-location": "1 - 2",
+    "app-route": "1 - 2",
     "iron-doc-viewer": "2.0-analyzer"
   }
 }


bower.json
```diff
@@ -21,6 +21,8 @@
     "firebase-element": "PolymerElements/firebase-element#^1.0.12"
   },
   "resolutions": {
+    "iron-location": "1 - 2",
```
arthurevans: D'oh. Had no idea we could do it so simply. In the docs I think we currently have something like `^1.9.1 || ^2.0.0` ... And we could get away with `1 - 2`? sheesh.

aomarks: That's what bower made for me :)

Also https://semver.npmjs.com/ is amazing for playing with semver expressions.

PR #2184, opened 2017-06-03
Analyzer-powered docs for Polymer 1.0

aomarks:
This updates polymer-project.org with new 1.0 documentation powered by Analyzer and the new versions of the `<iron-doc-*>` components, and drops the legacy Hydrolysis powered version.

@arthurevans It would be awesome if you have some time to check this branch out and see how you think it looks versus the current 1.0 docs.

You'll probably want to review the 3 commits separately, since one of them is purely the output of the script.

Also note that the script is just a copy of the existing 2.0 script with some small modifications, so I wouldn't really suggest reviewing that for style... if there's time I might clean up or unify both scripts at some point.

Once this is in we won't have any dependency on the Hydrolysis elements anymore, so then we can start ripping them out, and then finally merge the `2.0-analyzer` branch to master.


arthurevans:
LGTM. Docs seem like an improvement over the previous version (formatting, defaults). 

I thought the new system had fixed this bug with signatures (https://github.com/Polymer/polymer-analyzer/issues/231) but I think the case is that you can express varargs with ES6, so for example `Polymer.PropertyEffects.splice` generates the correct docs:

https://github.com/Polymer/polymer/blob/520407e9caa6289d270d451221c0a9786395b878/lib/mixins/property-effects.html#L1779

https://www.polymer-project.org/2.0/docs/api/mixins/Polymer.PropertyEffects#method-splice

While the 1.x equivalent (`Polymer.Base.splice`) produces broken docs:

https://github.com/Polymer/polymer/blob/1.x/src/standard/notify-path.html#L505

https://www.polymer-project.org/1.0/docs/api/Polymer.Base#method-splice

Since we're not going back to ES5, and this bug will probably never get fixed, we could probably close it out.

Cheers,
Arthur

PR #2188, opened 2017-06-05
Update step-2.md

joemarini:
Make the code copy-pasteable by closing the dom-module tag.

### DIFF
--- a/app/2.0/start/first-element/step-2.md
+++ b/app/2.0/start/first-element/step-2.md
@@ -62,6 +62,7 @@ Starting code—local DOM template { .caption }
     <!-- shadow DOM goes here -->
     <span>Not much here yet.</span>
   </template>
+</dom-module>
 ```
 
 Key information:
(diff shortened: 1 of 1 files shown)


joemarini:
I signed it!

app/2.0/start/first-element/step-2.md
```diff
@@ -62,6 +62,7 @@ Starting code—local DOM template { .caption }
     <!-- shadow DOM goes here -->
     <span>Not much here yet.</span>
   </template>
+</dom-module>
```
arthurevans: This is just showing the contents of the existing files, so it doesn't need to be copy-pastable. (The closing `<dom-module>` tag follows the `<script>`, shown in the next listing.)

app/2.0/start/first-element/step-2.md
```diff
@@ -91,7 +92,7 @@ element. If the element has a `<dom-module>`, this script is usually placed
 <em>inside</em> the `<dom-module>` to keep everything together.
 
 
-Starting code—element registration { .caption }
+Starting `code—element` registration { .caption }
```
arthurevans: That's an em dash, not a hyphen. The is the starting code, and this part deals with element registration.

We could change all of these em dashes to colon if you think it'd be clearer:

Starting code: HTML imports
Starting code: local DOM template  
Starting code: element registration

I think on the page the em dashes are pretty clear, but they look weird in markdown because the code font em dash is more or less identical to a hyphen.

PR #2194, opened 2017-06-09
Remove devguide shell

keanulee:
`<devguide-shell>` served mainly to layout the content on the split-view pages. It also created a stylized header for markdown pages using the page title specified by attributes on the element. This PR moves this layout work into the light DOM and server-side (run-time) templating, since in this case there's no benefit to doing this client-side (no bytes saved, since there's no repetition of content) and ever drawbacks (unnecessary client-side JS work).

`<devguide-shell>` also contained the docs version selector. This has been moved to a new `<pw-version-selector>` element which is now in the light DOM and updates the window location directly (the version selector in the drawer in `<pw-shell>` is also replaced by this).

This PR unblocks future work on the docs site, with the eventual goal of minimizing JS work required for first meaningful paint.

@arthurevans Can you please deploy this to a test server for more extensive testing?


keanulee:
Demo: https://no-devguide-shell-dot-polymer-project.appspot.com/

First meaningful paint on devguide pages (e.g. https://no-devguide-shell-dot-polymer-project.appspot.com/2.0/docs/devguide/feature-overview) is substantially faster since it is no longer blocked by lazy imports.
- Before: https://www.webpagetest.org/video/compare.php?tests=170622_NK_a2505fa61abdd97f8f4f3da6a35d218c-r:2-c:0
- After: https://www.webpagetest.org/video/compare.php?tests=170623_WM_6680c396fb3a413bdb339478c5296415-r:3-c:0

Other changes:
- Removed highlight.js from client-side since its use in pw-shell is broken and causes markdown pages to be slower (already highlighted by server)
- Move SW code to load listener to improve perf

notwaldorf:
Doing review now, but: I played around with it and the interaction looks good!

app/elements/pw-version-selector.html
```diff
@@ -0,0 +1,84 @@
...
+Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+-->
+<link rel="import" href="../bower_components/paper-dropdown-menu/paper-dropdown-menu-light.html">
+
+<dom-module id="pw-version-selector">
```
notwaldorf: 🙌

app/js/app.js
```diff
@@ -94,38 +94,40 @@ window.showToast = function(message) {
...
-        switch (installingWorker.state) {
-          case 'installed':
-            if (!navigator.serviceWorker.controller) {
-              window.showToast('Service Worker installed! Pages you view are cached for offline use.');
-            }
-            break;
-
-          case 'redundant':
-            throw Error('The installing service worker became redundant.');
+  window.addEventListener('load', function() {
```
notwaldorf: Why the extra event?

gulpfile.js
```diff
@@ -121,7 +121,6 @@ function writeServiceWorkerFile() {
       `${rootDir}/js/*.js`,
       `${rootDir}/css/*.css`,
       `${rootDir}/bower_components/**/webcomponents-lite.min.js`,
-      `${rootDir}/bower_components/highlight/highlight.js`,
```
notwaldorf: Ugh, we still have to figure out how to fix this. I'm ok with removing it for now, but maybe uncommenting it instead of deleting?
@arthurevans who has been looking into why it's hosed

polymer.json
```diff
@@ -2,14 +2,7 @@
   "entrypoint": "app/index.html",
   "shell": "app/elements/pw-shell.html",
   "fragments": [
-    "app/elements/pw-shell.html",
```
notwaldorf: Why don't we need pw-shell and friends as fragments?

app/js/app.js
keanulee: Perf - in the old timeline, the sw is loaded at the same time as the lazy imports (which should have priority over sw).

gulpfile.js
keanulee: Going forward, I suggest doing all highlighting on the server side (since I see little benefit of making the client doing the work).

polymer.json
keanulee: These are now all bundled in lazy-elements.html

app/js/app.js
notwaldorf: Please add a comment for future us, then?

polymer.json
notwaldorf: SG

notwaldorf:
This LGTM but i'd add a comment for that `load` event

keanulee:
If you load this site with an old SW, you will see broken styles (since static assets like pw-shell.html and site.css will be stale). The SW will be updated and the user will be prompted with the "Refresh site" toast, but until the refresh styles will be broken. I can't see a way around this (except for serving the old version pre-SW-upgrade and the new version post-SW-upgrade, but this means having multiple versions of the site active at the same time).

If this is acceptable, let's merge this PR. Otherwise, let me know.

arthurevans:
I think we can deal with that. 

I have a PR for server-side highlighting out. There's a slight hitch in it at present, but we could probably live with it if we wanted to. Waiting to see if the package owners will update an npm package or if we should hack around it...

Anyway, this LGTM.

PR #2202, opened 2017-06-14
Add automated validation of all external links in 2.0 docs

limonte:
[awesome_bot](https://github.com/dkhamsing/awesome_bot) is used on many awesome_* repositories, it will guarantee that there are no broken external links in the documentation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,11 +4,13 @@ python:
 install:
   - npm install
   - pip install WebTest
+  - gem install awesome_bot
 before_script:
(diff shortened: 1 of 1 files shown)


arthurevans:
Thanks for the PR! awesome_bot sounds, well... awesome. But we don't currently depend on any ruby packages, and I'm loathe to add another set of dependencies into this overloaded repo.

We currently have an external link checker bot that runs on the docs, so I'm going to close this.

PR #2196, opened 2017-06-12
Update iron-doc-viewer

keanulee:
Previous tag was deleted - this fixes `bower i` and `polymer build` steps

PR #2203, opened 2017-06-14
Updates for iron-doc-viewer 3.0 RC.

aomarks:
See https://github.com/Polymer/docs/pull/2196#issuecomment-308538751


arthurevans:
This all looks great to me, except that the 1.x behaviors docs don't match the rest of the docs. I should say, BEHAVIOR doc, because I think there's only one:

http://localhost:8080/1.0/docs/api/behaviors/Polymer.Templatizer

Maybe some style needs to be applied somewhere for the new element?

arthurevans:
Ah: tag name is wrong: https://github.com/Polymer/docs/blob/master/app/sass/_main.scss#L306

PR #2205, opened 2017-06-15
Fix iron-doc-default-theme include.

aomarks:


arthurevans:
Tested, staged to https://doc-theme-dot-polymer-project.appspot.com/2.0/docs/api/

I'm gonna kick this one over to @notwaldorf who has thoughts and opinions on the subject of font.

In particular, our regular body font is larger (16px) but thinner (300 weight), and I don't know offhand whether we should coordinate the API docs with the body and heading fonts used elsewhere.

arthurevans:
I'm pretty much always in favor of bigger fonts, myself. 👓

aomarks:
To be clear, there's no change to fonts, unless I am missing something. The theme here was just factored out because webcomponents.org overrides them. Previously these exact styles we're just inlined in the doc viewer.

aomarks:
Gonna bring this in with https://github.com/Polymer/docs/pull/2207.

PR #2206, opened 2017-06-15
Fix wrong option for entrypoint

inureyes:
* `--entry` should be `--entrypoint` for polymer CLI documentation


daKmoR:
@arthurevans looks about right

PR #2207, opened 2017-06-17
Add classes to 2.0 API docs.

aomarks:
Fixes https://github.com/PolymerElements/iron-doc-viewer/issues/122

Specifically we get `Polymer.Debouncer` and `Polymer.FlattenedNodesObserver`.


aomarks:
@arthurevans This is ready to go in now. Make sure to bower update, because this depends on some bug fixes that I just published to iron-doc-viewer 3.0.1.

arthurevans:
I am seeing random sorting of members. On the front page, I sometimes see Polymer.Element in the second position, sometimes in the last. And on, say, Polymer.Element page, the methods in both the public and protected sections seem to be ordered randomly.

(I also have kind of mixed feelings about the "protected methods" bar, because you have to scroll all the way to the bottom of the method list to see it. Not a big deal on most pages, but for big ones like Polymer.Element, you can imagine people giving up before they get to the bottom.)

aomarks:
For me Polymer.Element is consistently in the 2nd position in the list of elements, because it's by class name. For methods and properties, the ordering should be:

- public before protected
- static before not static
- not inherited before inherited
- ordered by inherited from class name
- ordered by method/property name

As for the bar, in the old layout when you checked "show protected", you then had to scroll down to find the protected things. So isn't that kind of the same? My issue with the previous design was that you have to scroll *up and then back down* to see protected methods, and that you don't know if there even *are* any protected methods to show. IMO the new one flows more logically (scroll through all the methods, then optionally see the additional protected ones if you want and there are any).

arthurevans:
Hmm... I think having multiple ordering like that in a single list is going to be really confusing.  

There are also just some things that I don't understand.

So here's the element ordering I see: 

```
<dom-module> (Polymer.DomModule)
Polymer.Element
<dom-bind> (Polymer.DomBind)
<dom-repeat> (Polymer.DomRepeat)
<dom-if> (Polymer.DomIf)
<array-selector> (Polymer.ArraySelector)
<custom-style> (Polymer.CustomStyle)
```

No matter *what* we're ordering by, array-selector should be first here, right?

Inside Polymer.Element, first I see `static finalize()` followed by `static createPropertiesForAttributes()`. They're both inherited (from different mixins). I guess this is explained by your ordering rules, but it just looks confusing to me.

Then in the regular methods, we have `attributeChangedCallback`,
`resolveUrl`, `updateStyles`, (good so far) followed by `disconnectedCallback`, and `connectedCallback`. in that order. All five of these are inherited from the same mixin, so I can't explain this at all.

I'd suggest that if we're sort into buckets, each should be marked out separately so users can tell what's going on. 

It's not too bad if static methods are at the start of a list of public methods, since the "static" at the beginning of the signature makes it easy to see why they're sorted there. But ordering by inherited-by is confusing because it's not the first thing you scan for on the line.

So maybe something like this?

Public methods
  static followed by instance

Protected methods
  static followed by instance

Inherited methods
  Inherited from <foo>
  Inherited from <bar>
 
  Or: all together, static followed by instance

This is more or less the way the Android docs are organized (more or less standard JavaDoc, I think? https://developer.android.com/reference/android/gesture/GestureUtils.html). But I do feel like if you want to order by inherited-from, each inherited-from superclass needs a little divider or miniheader so you know why the alphabetical order is changing. 

Alternately, the Sencha docs just have two buckets:

Instance methods
Static methods

And show/hide buttons for inherited, protected, etc., but that was the interface you didn't like before. (And I get it--it's a pain going back up to switch views. I don't think there's a perfect solution here.)

One thing that's a little tricky here is that for the core APIs, pretty much all of the API on Polymer.Element is inherited, and most of the APIs on many mixins are protected, so if those are collapsed, most of the API will be hidden. 

On the other hand, if you're documenting an element built with `Polymer.Element`, having the protected & inherited things hidden is exactly what you want. 

So... that inclines me towards something like the JavaDoc division, above, with "show protected" and maybe also "show inherited" bars? It at least makes it a lot clearer that there are hidden items, and how many there are.

aomarks:
Sounds like there must be a bug with sorting that you're hitting, I'll look into that.

Totally agree that the multi-sorting (even when it's working) is confusing, given that the buckets are not visually apparent. How about for this PR I'll remove inheritance from the sorting logic, so the sorting will be aligned exactly visually (visibility, static, name)? Then we can iterate on adding sections/groups for inheritance? I think I like the idea of being able to expand/contract the methods from each class you inherit from, like the Android docs,.

Also can you link to the Sencha docs you mentioned?

arthurevans:
Sounds like a plan.

Here are the Sencha docs. 

https://docs.sencha.com/extjs/6.0.2/classic/Ext.app.Application.html

Their doc interface is a little elaborate. People tend to either love or hate them. They use their own tooling (a project called JSDuck).

aomarks:
Hey @arthurevans, I just published a new `iron-doc-viewer` that fixes the sorting bug, and ignores inheritance as discussed. (Also a bunch of other fixes Sam found: https://github.com/PolymerElements/iron-doc-viewer/pull/158).

arthurevans:
Hey @aomarks ... Sorry for the late review. All looks good now, except that we have now developed conflicts. (I think because @keanulee's PR to remove the devguide shell got merged.) Can you resolve?

aomarks:
Resolved merge conflicts.

templates/base-devguide.html
```diff
@@ -12,6 +12,7 @@
 <head>
 <title>{% block title %}{% endblock %} - Polymer Project</title>
 {% include 'templates/head-meta.html' %}
+<style is="custom-style" include="iron-doc-default-theme"></style>
```
keanulee: When navigating from a different page on the docs site, the head is not copied, so these styles won't be available. Also with the upcoming changes in #2207, Polymer (and therefore `custom-style` and `<dom-module id="iron-doc-default-theme">`) will be async loaded, so including these in `<head>` is not appropriate. I moved these to iron-doc-viewer.html in commit https://github.com/Polymer/docs/pull/2232/commits/331c84719e3845310aadba2fdabf57edf10e1da2 (in PR #2207).

aomarks: Ah I see, thanks for fixing!

PR #2208, opened 2017-06-21
add static getter is to the custom element class

stkevintan:
As the doc reads:  
> The class must have a static `is` getter that returns the HTML tag name for your custom element.


stkevintan:
I signed it!

arthurevans:
Here, the sentence is out of date, not the example. The `is` getter is no longer required in all cases. It's only required when you specify a DOM template using a `dom-module`.

stkevintan:
thx! I see.

PR #2214, opened 2017-06-26
Set cache-control header for SW to 1s

keanulee:
Currently all static assets are cached for 10 minutes (App Engine default -
 https://cloud.google.com/appengine/docs/standard/python/config/appref#default_expiration). This adds an exception for the SW since it should be updated as soon as possible (https://developers.google.com/web/fundamentals/instant-and-offline/service-worker/lifecycle#updates)

Staged at https://service-worker-dot-polymer-project.appspot.com/

### DIFF
--- a/app.yaml
+++ b/app.yaml
@@ -45,6 +45,7 @@ handlers:
   static_files: service-worker.js
   upload: service-worker.js
   secure: always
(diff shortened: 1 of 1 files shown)


keanulee:
Re-staged at https://2017-06-26-dot-polymer-project.appspot.com

PR #2218, opened 2017-06-27
Clarify automatic listener addition and removal

ernsheong:
in Events guide


ernsheong:
This was removed in v2 docs, but was present in v1 (https://www.polymer-project.org/1.0/docs/devguide/events) Is this still the case now?

PR #2220, opened 2017-06-27
Upgrade old css selector

stkevintan:
`::content` selector belongs to Polymer 1.x  and it has been upgrade to `::slotted()` selector in Polymer 2 , according to the [official upgrade doc](https://www.polymer-project.org/2.0/docs/upgrade).

### DIFF
--- a/app/2.0/docs/devguide/custom-css-properties.md
+++ b/app/2.0/docs/devguide/custom-css-properties.md
@@ -488,7 +488,7 @@ The custom properties shim doesn't support styling distributed elements.
 
 ```css
 /* Not supported */
-:host ::slotted(*) div {
+:host ::slotted(div) {
   --custom-color: red;
 }
 ```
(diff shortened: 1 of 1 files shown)


app/2.0/docs/devguide/custom-css-properties.md
```diff
@@ -488,7 +488,7 @@ The custom properties shim doesn't support styling distributed elements.
 
 ```css
 /* Not supported */
-:host ::content div {
+:host ::slotted(*) div {
```
jsilvermist: I believe this is invalid as nothing should be to the right of `::slotted()`, I'm pretty sure the correct syntax would be `::slotted(div)`.

stkevintan: Sorry, it's my  misunderstanding.

katejeffreys:
@arthurevans Sanity check pls, when you get the chance?

arthurevans:
LGTM.

PR #2227, opened 2017-06-29
Update gesture-events.md

keanulee:
Ref https://github.com/PolymerElements/app-layout/issues/465#issuecomment-298401945


arthurevans:
Awesome, thanks. @katejeffreys we should look to update the element tutorial (https://www.polymer-project.org/2.0/start/first-element/step-4)... Sounds like we can simplify this step by removing the gesture event mixin.

abdonrd:
I added another PR about `on-tap` mention: https://github.com/Polymer/docs/pull/2283

PR #2232, opened 2017-06-30
First paint

keanulee:
Remove [unresolved] styling

Site is staged at https://first-paint-dot-polymer-project.appspot.com


app/404.html
```diff
@@ -13,8 +13,7 @@
 <title>404 - Polymer Project</title>
 {% include 'templates/head-meta.html' %}
 </head>
-<body class="fullbleed" unresolved>
```
keanulee: `.fullbleed` doesn't actually do anything (no CSS rule)

katejeffreys:
Cool cool, this looks like a non-trivial change so I will test it. Anything specific I should look for?

keanulee:
Site is staged at https://first-paint-dot-polymer-project.appspot.com

@katejeffreys nothing specific, just making sure the page content load and style correctly (even before the header nav is fully loaded) and SW shows cached pages when offline (note that the polymerosaurus is now removed when you try to load an uncached page from SW - this is so it doesn't appear while the app shell tries to fetch fresh content from the server).

katejeffreys:
LGTM.

app/sass/_layout.scss
```diff
@@ -0,0 +1,294 @@
...
+This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+*/
+
+/**
+ * There styles were pulled from iron-flex-layout/iron-flex-layout-classes.html
+ * so that they are in the main stylesheet.
```
frankiefu: Should we remove `iron-flex-layout` import in [pw-shell.html](https://github.com/Polymer/docs/blob/0f704a4840b27e5d9ad5d110b21dd5adc4b35fcc/app/elements/pw-shell.html#L18)?

frankiefu:
When I clicked on the "Community" tab, I saw error in the console:
```
pw-shell.html:formatted:10426 Uncaught TypeError: Cannot read property 'opened' of null
    at HTMLElement._styleLevel1Link (pw-shell.html:formatted:10426)
    at HTMLElement.styleLinkPath (pw-shell.html:formatted:10416)
    at HTMLElement._handleResponse (pw-shell.html:formatted:10380)
    at HTMLElement.S (pw-shell.html:formatted:2671)
    at HTMLElement.fire (pw-shell.html:formatted:3237)
    at HTMLElement._handleResponse (pw-shell.html:formatted:8462)
    at <anonymous>
```

app/sass/_layout.scss
keanulee: No, there are still other elements (e.g. pw-shell, demo-tabs, pw-footer) that rely on those classes (and because they're in shadow roots, they don't share the same stylesheet).

keanulee:
Restaged at https://first-paint-dot-polymer-project.appspot.com

keanulee:
Merged in changes from `master`, but having trouble redeploying to staging. Will try again a little later.

keanulee:
Redeployed to staging (after updating gcloud CLI tools)

keanulee:
Restaged after latest changes

notwaldorf:
If I compare loading https://polymer-project.appspot.com/2.0/docs/upgrade#remove-type-extension-elements to loading https://first-paint-dot-polymer-project.appspot.com/2.0/docs/upgrade#remove-type-extension-elements, there's a definite first paint jank. In the staged version you see the whole footer painting before you get any content. This used to be done with the `unresolved` attribute peppered everywhere, and only removed when we were fully upgraded.

arthurevans:
Hm. I'm not seeing the footer painting, but I am seeing the content being painted once then repainted at a different scroll position (or perhaps just scrolled). This is with network throttled to 3G fast and SW unregistered to make it easier for me to see what's going on.

Also, I'm not sure if this is just devtools weirdness, but when I go to this link in device emulation mode, it paints first in the correct scroll position, then a second time in the _wrong_ scroll position (that is, the "Remove type-extension elements" header is not on screen. Instead, it's somewhere near the DOM APIs header, which is quite a ways away.

keanulee:
@notwaldorf @arthurevans Hash ID linking is default browser behavior, and the initial flash of unscrolled content is kind of unavoidable since the browser hasn't rendered the markup of the referenced ID yet. This wasn't noticeable before because everything was `opacity: 0`. When upgraded, `pw-shell` has code that scrolls back to that element if it becomes out of view (e.g. if the user scrolls after the browser get there, but before `pw-shell` upgrades), and that's needed for the XHR navigation case.

If you refresh the page after SW installs, you'll get app-shell.html that has some placeholder whitespace of `100vh` (https://github.com/Polymer/docs/pull/2232/files#diff-3a37d6f68dd88b4742f81c247719f85eR21). However, if you refresh while scrolled, the browser will keep the scroll position so you'll see the footer before the XHR for page content is done. This could be mitigated by making that `100vh` a larger value (e.g. `10000px`), but that seems like an unwarranted hack just to serve this use case.

tl;dr - hash linking is special; not too keen on making the first paint experience perfect for this use case (since that's not guaranteed by the browser for regular web pages anyway - try this link for example https://github.com/Polymer/docs/pull/2232/files#diff-b9e12334e9eafd8341a6107dd98510c9R99).

arthurevans:
I'm still seeing a little weirdness when reloading a page at an anchor (it first scrolls to the anchor, then when the page finishes reloading, it scrolls up or down a bit). However, it seems to work well on initial page load. Guess I should test on an iDevice to see how it works without SW.

I don't have any strong feelings on this. @notwaldorf WDYT?

keanulee:
Restaged changes at https://first-paint-dot-polymer-project.appspot.com/2.0/docs/api/. @notwaldorf PTAL?

notwaldorf:
LGTM. Not super concerned about the hash linking, since I've seen problems on the live site, so I don't think this PR is changing that

PR #2260, opened 2017-07-20
Update polymer json

### LINKED ISSUE #2255: Document the basePath entry in polymer.json

katejeffreys:
Fixes #2255

### DIFF
--- a/app/2.0/docs/tools/polymer-json.md
+++ b/app/2.0/docs/tools/polymer-json.md
@@ -138,29 +138,66 @@ the `builds` property. This is equivalent to passing different CLI flags to the 
 storing them here will configure the build
 for every run:
 
-* `name`: An optional name for your build. If multiple builds are defined, the `name` property is
-required.
-* `preset`: An optional preset name that your build configuration can inherit from. See below for more information.
-* `addServiceWorker`: If `true`, generate a service worker for your application.
-* `addPushManifest`: If `true`, generate an [HTTP/2 Push Manifest](https://github.com/GoogleChrome/http2-push-manifest) for your application.
-* `swPrecacheConfig`: An optional configuration file for the generated service worker.
-* `insertPrefetchLinks`: If `true`, insert prefetch link elements into your fragments so that all
-dependencies are prefetched immediately.
-* `bundle`: May be boolean, or an object bag of settings matching specified [`polymer-bundler`](https://github.com/Polymer/polymer-bundler) constructor options. This functionality emulates the behavior of the Polymer 1.x build tools.
-  * If `true`, bundle your application.
-  * If an object is provided with further configuration, the options are:
-    * `excludes`: Optional, `String[]`. Excludes the listed paths of files or folders from inlining.
-    * `inlineCss`: Optional, `Boolean`. Specifies whether to inline external CSS file contents into `<style>` tags.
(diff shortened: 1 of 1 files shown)


app/2.0/docs/tools/polymer-json.md
```diff
@@ -138,29 +138,66 @@ the `builds` property. This is equivalent to passing different CLI flags to the
...
+    *   `inlineScripts`: Inline external Javascript file contents into `<script>` tags.
+    
+    *   `rewriteUrlsInTemplates`: Rewrite element attributes inside templates when inlining html.
+
+    *   `sourcemaps`: Create identity source maps for inline scripts.
+
+    *   `stripComments`: Remove all comments except those tagged `@license`, or starting with
+        `<!--!` or `<!--#`.
+
+*   `html`:
```
justinfagnani: should these say they're an object to be extra clear about what a nested item means?

app/2.0/docs/tools/polymer-json.md
```diff
@@ -190,12 +227,12 @@ And here is a configuration to generate two optimized builds: One bundled and on
   }]
 ```
 
-Here is a configuration to generate a bundled, minified application build with the following bundling options:
+The following configuration generates a bundled, minified application build with bundling options:
```
justinfagnani: Is "build" (in "application build") a noun or verb? Should this be "application built with..."?

app/2.0/docs/tools/polymer-json.md
katejeffreys: Could you clarify? Currently the text states that `bundle` is an object:

>`bundle`: You may provide a boolean (`true` to bundle your application; default is `false`), or an object containing bundling configuration options...

app/2.0/docs/tools/polymer-json.md
katejeffreys: It's a noun.

app/2.0/docs/tools/polymer-json.md
justinfagnani: I just mean does that also for for `html`, etc

app/2.0/docs/tools/polymer-json.md
```diff
@@ -138,29 +138,66 @@ the `builds` property. This is equivalent to passing different CLI flags to the
...
+    *   `minify`: If `true`, minify all CSS.
+
+*   `js`: An object containing configuration options for Javascript:
+
+    *   `minify`: If `true`, minify all JS.
+
+    *   `compile`: If `true`, use babel to compile all ES6 JS down to ES5.
+
+*   `browserCapabilities`: Capabilities required for a browser to consume this build. 
+    Values include `es2015` and `push`. [See the canonical list](https://github.com/Polymer/prpl-server-node/blob/master/src/capabilities.ts).
```
arthurevans: I think it would be far better to link to the README:

https://github.com/Polymer/prpl-server-node#capabilities

Also, there are currently only three values, `es2015`, `push`, and `serviceworker`, so perhaps just list all three and say, "For more information, see the `prpl-server-node` documentation."

Longer term, think about including these definitions in this file, the way we do the presets.

arthurevans:
A few comments otherwise LGTM.

PR #2264, opened 2017-07-21
Array mutations not tracked in data bindings

### LINKED ISSUE #2228: Binding to array elements
[Data binding](https://www.polymer-project.org/2.0/docs/devguide/data-binding#array-binding) docs says not to use `<span>{{array.0}}</span>`.  This actually works.  Is this a bug in polymer or in the documentation?  Is there some limitation to array element binding in this form?

katejeffreys:
Array mutations are not tracked and won't reflect in static bindings. Add note to that effect.

Fixes #2228

### DIFF
--- a/app/2.0/docs/devguide/data-binding.md
+++ b/app/2.0/docs/devguide/data-binding.md
@@ -524,6 +524,9 @@ item**.
 <span>{{array.0}}</span>
 ```
 
+**Array binding does not track array mutations.** 
+An array mutation that reorders an array (for example, using the [splice method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice)) will not be reflected in static bindings.
+{ .alert .alert-info }
 
 There are several ways to interact with arrays in a data binding:
 
--- a/app/2.0/docs/devguide/data-binding.md
+++ b/app/2.0/docs/devguide/data-binding.md
@@ -518,15 +518,11 @@ To keep annotation parsing simple, **Polymer doesn't provide a way to bind direc
 item**.
 
 ```
-<!-- Don't do this! -->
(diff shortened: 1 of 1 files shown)


app/2.0/docs/devguide/data-binding.md
```diff
@@ -524,6 +524,9 @@ item**.
 <span>{{array.0}}</span>
 ```
 
+**Array binding does not track array mutations.** 
+An array mutation that reorders an array (for example, using the [splice method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice)) will not be reflected in static bindings.
```
arthurevans: This doesn't exactly follow, because we just told then that these bindings don't work.

You might say something like,

A binding in the `[[array.0]]` format may _appear_ to work, but it's severely limited:

-   The array index can only be a number literal, not a variable. For example, `[[array.i]]` binds
     to the `i` property of `array`, not `array[i]`. 
-   The binding doesn't update  when the array is mutated. For example, the following would
     update the binding:
    
        this.array[0] = 'New value';
 
     But the following mutation would not:

        // replace first array item
        this.splice('array', 0, 1, 'New value');

That would be correct. But it's a pretty lengthy digression that detracts from the main point of the section, which is "don't do `[[array.0]]`.

Alternately, we could just embed comments in the code snippet.

 ```
 <!-- Don't do this! This format doesn't work -->
 <span>[[array[0]]]</span>
 <!-- Don't do this! Data may display, but won't be updated correctly -->
  <span>[[array.0]]</span>
  ```

katejeffreys: Thanks! I've gone with the code comments option to avoid the digression.

PR #2265, opened 2017-07-21
Fix error with force polyfill snippet

TimvdLippe:
If you would use the snippet as is, the following error is thrown:

> Uncaught ReferenceError: Invalid left-hand side in assignment

Changing it to an `if` does work :smile:


arthurevans:
Oop. LGTM, thanks.

PR #2269, opened 2017-07-26
Warn about infinite loops

katejeffreys:


katejeffreys:
@rianby64 This replaces your previous PR :)

rianby64:
I think the same idea applies to Polymer 2.0...

PR #2272, opened 2017-07-27
Add promise polyfill for IE.

### LINKED ISSUE #2271: Navigation fails on IE.
Initial page load works OK, but on page refresh, I'm seeing no change, and a `Promise is undefined` error in the console.

I believe this is the result of upgrading the the hybrid iron-ajax (2.0.0) which requires a promise polyfill on IE. The polyfill was included in iron-ajax 1.x, and is included in the webcomponents v1 polyfill. Running iron-ajax 2.0.0 with Polymer 1.x, you need the polyfill.

arthurevans:
Fixes #2271.

This fix is currently live in production.


keanulee:
Thanks Arthur

PR #2289, opened 2017-08-09
Add note about behavior of rewriteUrlsInTemplates

katejeffreys:
Behaves differently for Polymer 2.x vs 1.x. Add note to that effect.


arthurevans:
Might want to reference https://www.polymer-project.org/2.0/docs/upgrade#urls-in-templates here to explain the difference.

PR #2296, opened 2017-08-16
Add speaker spotlight for Kunal

katejeffreys:


katejeffreys:
@wmginsberg @arthurevans

PR #2306, opened 2017-08-28
Fix typo data system#make-observable-changes

poliak9:
Fix typo in sentence "This mixin may trade(s)" as the third person "s" should be dropped with the use of modal verb "may"


katejeffreys:
Replaced by #2310

PR #2307, opened 2017-08-28
edit intro text to be more accessible to beginners; add edit suggestions

eblaine:


e111077:
@eblaine next step would be to deploy

app/2.0/start/quick-tour.md
```diff
@@ -1,21 +1,29 @@
 ---
-title: Quick tour of Polymer
+title: Try Polymer
```
arthurevans: If we change the title here, we need to change the title in the navigation as well. And ideally check the other pages that reference that and update any references to "The quick tour" (although I think they mostly say something like "take a quick tour," which is generic enough that it's fine).

app/2.0/start/quick-tour.md
```diff
@@ -41,6 +49,11 @@ contain a dash (-)**.
   <iframe frameborder="0" src="samples/custom-element/index.html" width="100%" height="40"></iframe>
 </demo-tabs>
 
+* Try modifying the contents of `this.textContent`.
```
arthurevans: :heart: the prompts. Is it obvious enough that they need to click the plunker buttons to make these changes?

app/2.0/start/quick-tour.md
```diff
@@ -93,6 +106,9 @@ with a styled `<div>` tag.
   <iframe frameborder="0" src="samples/picture-frame/index.html" width="100%" height="60"></iframe>
 </demo-tabs>
 
+* Try adding divs in index.html to observe how they are unaffected by the styles applied to the divs in picture-frame’s shadow DOM.
```
arthurevans: We generally use code font and brackets for element names, and code font for filenames. "to observe how they are unaffected" is a little awkward, maybe rephrase as a question:

Try adding a `<div>` to `index.html`; is it affected by the styles in `<picture-frame>`'s shadow DOM?

e111077:
Ah sorry about that, Arthur, we got an LGTM from Kate on Slack. We probably should have made that more-apparent on the PR.

arthurevans:
👍  just wanted to make sure.

If we want to re-title this, we should change at least the navigation (`app/2.0/nav.yaml`) and probably the Take the Tour CTA  here: https://www.polymer-project.org/2.0/start/

katejeffreys:
Agree with all @arthurevans' comments, I will incorporate them. Thanks!

PR #2316, opened 2017-08-31
Clarify reflectToAttribute mapping

TimvdLippe:
As reported in the linked issue, clarify the mapping of how properties are mapped to attributes.

Closes https://github.com/Polymer/polymer/issues/4749

### DIFF
--- a/app/2.0/docs/devguide/properties.md
+++ b/app/2.0/docs/devguide/properties.md
@@ -374,9 +374,11 @@ For more on read-only properties and data binding, see
 
 In specific cases, it may be useful to keep an HTML attribute value in sync with
 a property value.  This may be achieved by setting `reflectToAttribute: true` on
-a property in the `properties` configuration object.  This will cause any
-[observable change](data-system#observable-changes) to the property to be serialized out to an
-attribute according to the reverse mapping as described in <a href="#property-name-mapping">Property name to attribute name mapping</a>.
+a property in the `properties` configuration object. This causes any observable
+change to the property to trigger an update to the corresponding attribute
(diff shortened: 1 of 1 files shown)


app/2.0/docs/devguide/properties.md
```diff
@@ -376,7 +376,7 @@ In specific cases, it may be useful to keep an HTML attribute value in sync with
 a property value.  This may be achieved by setting `reflectToAttribute: true` on
 a property in the `properties` configuration object.  This will cause any
 [observable change](data-system#observable-changes) to the property to be serialized out to an
-attribute of the same name.
+attribute according to the reverse mapping as described in <a href="#property-name-mapping">Property name to attribute name mapping</a>.
```
arthurevans: Going to defer to @katejeffreys on this because I've read this stuff so many times...

Does this make sense? 

Or, "serialized out to the corresponding attribute as described in ..."

Or (because we don't define serialized until the next section), perhaps should be something like: 

This causes any observable change to the property to trigger an update to the corresponding attribute (as described in Property name to attribute name mapping). Since attributes only take string values, the property value is serialized to a string, as described in Attribute serialization(link).

arthurevans:
Thank you @TimvdLippe Ping @katejeffreys can you take a look at the above ^^ and see what makes sense to you?

app/2.0/docs/devguide/properties.md
katejeffreys: I like this best:

This causes any observable change to the property to trigger an update to the corresponding attribute (as described in Property name to attribute name mapping). Since attributes only take string values, the property value is serialized to a string, as described in Attribute serialization(link).

TimvdLippe:
Updated :tada:

PR #2318, opened 2017-09-05
Update site to use Polymer 2.0

### LINKED ISSUE #2242: Update Polymer/docs code to Polymer 2.x

keanulee:
Fixes #2242 

Staged at https://polymer-2-dot-polymer-project.appspot.com/ - please leave feedback as comments here on this PR.

TODOs:
- [x] Make elements hybrid
- [x] Compile to ES5 and make sure webcomponents-loader.js works
- [x] Transpile-to-ES5/minify/bundle demos in /1.0/samples/homepage/
  - [x] Optional: Make homepage demos use webcomponents-loader.js instead of -lite.js
    - ~Not possible since `customElements.define()` call is bundled in index.html and must be polyfilled before in index.html~
    - The 2 homepage demos are now treated as standalone projects with their own `polymer.json` and built/bundled using `polymer build`. Element definitions are in the shell HTML import, so polyfills are guaranteed to be run by the time the shell is imported.
- [x] Page navigation seems broken in Firefox/Edge/IE (does full-page reload)
- [x] [IE] Promise.catch missing in IE - see https://github.com/webcomponents/webcomponentsjs/issues/837
  - [x] [IE] `<a>` link click handler doesn't stop page navigation (version selector now works though) https://github.com/PolymerElements/iron-location/issues/93
  - [x] Waiting for the above 2 patches to be merged and tagged
- [x] [IE] Issue with template polyfill (t.content.querySelectorAll missing) on homepage demos
- [x] Port summit 2015 and 2016 sites, or check in Polymer 1 deps in those dirs
- [x] Service worker should precache webcomponents-loader.js and runtime cache needed polyfill


keanulee:
The staging website (https://polymer-2-dot-polymer-project.appspot.com) is now ready for testing on all browsers, including IE 11! I'm waiting until patched dependencies are merged and tagged before merging this PR though.

notwaldorf:
I have been browsing and refreshing this for the last 5 minute on IE11, and it's pretty aces!

app/2.0/samples/homepage/contact-card/index.html
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
-    <script src="../../../../bower_components/webcomponentsjs/webcomponents-lite.min.js"></script>
+    <script src="bower_components/webcomponentsjs/webcomponents-loader.js"></script>
```
notwaldorf: yasssssss

app/2.0/samples/homepage/google-map/bower_components
```diff
@@ -0,0 +1 @@
+../../../../bower_components
```
notwaldorf: oh is this a link? smart!

app/elements/demo-tabs.html
```diff
@@ -93,12 +93,12 @@
...
     <div class="layout horizontal">
       <iron-pages selected="[[_selected]]">
-        <content id="contentpanels" select="demo-tab"></content>
+        <slot id="contentpanels" name="demo-tab"></slot>
       </iron-pages>
     </div>
 
     <div class="results">
-      <content select="iframe"></content>
+      <slot select="iframe"></slot>
```
notwaldorf: this should be `name="iframe"`, no?

app/elements/lazy-elements.html
```diff
@@ -7,15 +7,15 @@
...
 -->
+<link rel="import" href="icons.html">
+
 <link rel="import" href="../bower_components/app-layout/helpers/helpers.html">
 <link rel="import" href="../bower_components/iron-collapse/iron-collapse.html">
 <link rel="import" href="../bower_components/iron-selector/iron-selector.html">
 <link rel="import" href="../bower_components/paper-icon-button/paper-icon-button.html">
 <link rel="import" href="../bower_components/paper-item/paper-item.html">
 <link rel="import" href="../bower_components/paper-listbox/paper-listbox.html">
-<link rel="import" href="../bower_components/paper-toast/paper-toast.html">
```
notwaldorf: isn't the toast needed by the service worker code?

notwaldorf: (like, why isn't it lazy anymore)

app/elements/pw-shell.html
```diff
@@ -340,15 +341,18 @@
...
-        while (shell.firstChild) {
-          fragment.appendChild(shell.firstChild);
+        while (docShell.firstChild) {
+          fragment.appendChild(docShell.firstChild);
         }
 
-        Polymer.dom(this).appendChild(document.importNode(fragment, true));
+        // Replace shell contents.
+        var shell = Polymer.dom(this);
+        while (shell.firstChild) {
```
notwaldorf: Why is this better than `shell.innerHTML = '';`?

bower.json
```diff
@@ -29,13 +30,14 @@
...
+    "google-youtube": "GoogleWebComponents/google-youtube#2",
+    "google-map": "GoogleWebComponents/google-map#2",
+    "web-animations-js": "web-animations/web-animations-js#^2.2.5",
+    "webcomponentsjs": "master"
   },
   "resolutions": {
-    "iron-selector": "2",
-    "webcomponentsjs": "^0.7.24"
+    "iron-location": "fix-ie-capture",
+    "webcomponentsjs": "master"
```
notwaldorf: but you're going to update this to depend on real releases before you merge this, right?

templates/head-meta.html
```diff
@@ -14,8 +14,11 @@
...
...function"!=typeof r?e:r},r.set=function e(r,t,n,o){var i=Object.getOwnPropertyDescriptor(r,t);if(void 0===i){var a=Object.getPrototypeOf(r);null!==a&&e(a,t,n,o)}else if("value"in i&&i.writable)i.value=n;else{var u=i.set;void 0!==u&&u.call(o,n)}return n},r.slicedToArray=function(){function e(e,r){var t=[],n=!0,o=!1,i=void 0;try{for(var a,u=e[Symbol.iterator]();!(n=(a=u.next()).done)&&(t.push(a.value),!r||t.length!==r);n=!0);}catch(e){o=!0,i=e}finally{try{!n&&u.return&&u.return()}finally{if(o)throw i}}return t}return function(r,t){if(Array.isArray(r))return r;if(Symbol.iterator in Object(r))return e(r,t);throw new TypeError("Invalid attempt to destructure non-iterable instance")}}(),r.taggedTemplateLiteral=function(e,r){return Object.freeze(Object.defineProperties(e,{raw:{value:Object.freeze(r)}}))},r.temporalRef=function(e,r,t){if(e===t)throw new ReferenceError(r+" is not defined - temporal dead zone");return e},r.temporalUndefined={},r.toArray=function(e){return Array.isArray(e)?e:Array.from(e)},r.toConsumableArray=function(e){if(Array.isArray(e)){for(var r=0,t=Array(e.length);r<e.length;r++)t[r]=e[r];return t}return Array.from(e)}}("undefined"==typeof global?self:global);</script>
```
notwaldorf: Yum. What is this?

app/elements/pw-shell.html
keanulee: DOM manipulations are faster than invoking the parser, I think (even for empty strings), though the real motivation behind this change was that setting `innerHTML` was causing issues.

app/elements/lazy-elements.html
keanulee: On Firefox (at least), if you navigate to a non-cached page, the toast won't be upgraded and you don't see it (and get an ugly "show is not a function" error).

app/elements/demo-tabs.html
keanulee: It _should_ be `<slot name="iframe">` with the corresponding iframes updated with `slot="iframe"`, but since there's no default slot before I'm just going to make this the default (and remove `select="iframe"`).

FluorescentHallucinogen:
@keanulee Could you please also update the old and/or deprecated versions of dependencies such as `app-layout#^0.0.14` and `firebase-element#^1.0.12` to actual versions?

FluorescentHallucinogen:
@keanulee Oh, dependenies are already updated. Please ignore my previous message. One question: why you don't use version ranges in `bower.json` for some dependenies (e.g. `paper-item#^2.0.0` instead of `paper-item#2`)?

keanulee:
@FluorescentHallucinogen I like `#2` because it's succinct. It equivalent to `#^2.0.0`, just personal preference 😉  (Note that `#2.2.5` is NOT equivalent to `#^2.2.5` - `2` means versions that "start with 2, where the latter is an exact match)

arthurevans:
Is the IE iron-location issue the last issue blocking this?

keanulee:
The iron-location fix is now published and I updated the staging site with the latest bower deps (https://polymer-2-dot-polymer-project.appspot.com/). I tested the latest update and everything seems to be good. @arthurevans wanna give it one last check and merge? (I believe I already addressed all @notwaldorf 's feedback.)

bower.json
```diff
@@ -29,13 +29,9 @@
     "paper-listbox": "PolymerElements/paper-listbox#2",
     "paper-tabs": "PolymerElements/paper-tabs#2",
     "paper-toast": "PolymerElements/paper-toast#2",
-    "promise-polyfill": "polymerlabs/promise-polyfill#^1.0.0",
-    "google-youtube": "GoogleWebComponents/google-youtube#^1.2.1",
-    "google-map": "GoogleWebComponents/google-map#^1.1.10",
+    "promise-polyfill": "polymerlabs/promise-polyfill#2",
```
arthurevans: I think we can drop the separate promise-polyfill now... There's a promise polyfill in the v1 WC polyfills, so WC-loader should take care of this, I think. (Only needed this to run the hybrid iron-ajax against 1.x, IIRC).

arthurevans:
Will pull and do a test build tomorrow to make sure I don't have any issues building locally.

PR #2323, opened 2017-09-11
Update templates.md

ecollett1:
Wouldn't work with isEngineer: function(item) {...} Need it this way for ECMAScript 6


ecollett1:
I signed it!

PR #2324, opened 2017-09-12
`!/__/**` rewrite is unnecessary and breaks things

Scarygami:
According to the [Firebase Hosting docs](https://firebase.google.com/docs/hosting/url-redirects-rewrites#section-priorities) the reserved `/__/**` urls will always have priority so rewriting them isn't necessary.

And since this rewrite rule is more or less a catch-all, it will already trigger for everything before the second rule can take effect.

In a normal Polymer app this means that trying to access https://my-app.firebaseapp.com/doesnotexist will try to load `src/my-doesnotexist.html` which gets rewritten to index.html and therefore doesn't trigger the 404 fallback.

### DIFF
--- a/app/2.0/start/toolbox/deploy.md
+++ b/app/2.0/start/toolbox/deploy.md
@@ -186,10 +186,6 @@ guide](https://www.firebase.com/docs/hosting/quickstart.html).
         "public": "build/es5-bundled/",
         "rewrites": [
           {
-            "source": "!/__/**",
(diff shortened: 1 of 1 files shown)


keanulee:
@Scarygami Thank you for your PR. The points you made are all valid, but I'm going to close this in favor of https://github.com/Polymer/docs/pull/2333 since it includes this change and also in a few other places.

PR #2325, opened 2017-09-13
Clarify private setter of readOnly properties

TimvdLippe:
Add an extra clarification on how readOnly property names are used in a private setter.

Fixes https://github.com/Polymer/polymer/issues/4085

### DIFF
--- a/app/2.0/docs/devguide/properties.md
+++ b/app/2.0/docs/devguide/properties.md
@@ -344,7 +344,9 @@ When a property only "produces" data and never consumes data, this can be made
 explicit to avoid accidental changes from the host by setting the `readOnly`
 flag to `true` in the `properties` property definition.  In order for the
 element to actually change the value of the property, it must use a private
-generated setter of the convention <code>\_set<var>Property</var>(value)</code>.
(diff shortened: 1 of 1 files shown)


app/2.0/docs/devguide/properties.md
```diff
@@ -344,7 +344,9 @@ When a property only "produces" data and never consumes data, this can be made
 explicit to avoid accidental changes from the host by setting the `readOnly`
 flag to `true` in the `properties` property definition.  In order for the
 element to actually change the value of the property, it must use a private
-generated setter of the convention <code>\_set<var>Property</var>(value)</code>.
+generated setter of the convention <code>\_set<var>Property</var>(value)</code>
+where <code><var>Property</var></code> is transformed as
```
arthurevans: I'd prefer words, perhaps with examples. 

where `<code><var>Property</var></code>` is the property name, with the first character converted to uppercase (if alphabetic). For example, the setter for `oneProperty` is `setOneProperty`, and the setter
for _privateProperty is `set_privateProperty`.

WDYT?

TimvdLippe: Sounds great, updated! :+1:

PR #2326, opened 2017-09-13
Fix computed style example

arthurevans:
Per https://github.com/Polymer/polymer/issues/4374#issuecomment-291229633

Already correct in the upgrade guide, just matching it here.


arthurevans:
Ping @katejeffreys (mostly FYI as it's a pretty simple fix).

PR #2329, opened 2017-09-14
chore(step-4.md): Removed obsolete code

noahlaux:
It seems like there's a superfluous event listener.

### DIFF
--- a/app/2.0/start/first-element/step-4.md
+++ b/app/2.0/start/first-element/step-4.md
@@ -87,7 +87,6 @@ icon-toggle.html { .caption }
     }
     constructor() {
       super();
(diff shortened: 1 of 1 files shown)


arthurevans:
Seems like this section is out of date and to match with the finished version in the repo, we should just be adding an on-click listener to the icon? I thought we'd updated this section, perhaps @katejeffreys will remember?

katejeffreys:
I thought I updated it too. It's definitely out of date because on-tap and the gesture event mixin isn't necessary any more. I have updated the code and will close this PR in favor of the new one. Thanks @noahlaux for bringing this to our attention!

PR #2333, opened 2017-09-21
Update firebase deploy instructions

keanulee:
The rewrite rule for `!/__/**` is unnecessary since Firebase Hosting already handles those URLs without any configuration (see https://firebase.google.com/docs/hosting/url-redirects-rewrites#section-priorities). Moreover, this results in URLs with extensions (like /src/my-nonexistent-view.html) to incorrectly rewrite to index.html and causes issues like https://github.com/PolymerElements/polymer-starter-kit/issues/1061.

(The `!/__/**` rule was first added in https://github.com/Polymer/docs/pull/1815)


jsilvermist:
Should we also add `*.gif` as it already includes the other popular image formats?

PR #2334, opened 2017-09-21
Add fallback in pwShellSWHandler

### LINKED ISSUE #2319: Intermittent load failures
When I navigate to the site via a link, I sometimes get a loading error like this:

This site can’t be reached

The webpage at https://www.polymer-project.org/2.0/docs/devguide/hybrid-elements might be temporarily down or it may have moved permanently to a new web address.
ERR_FAILED

This appears to be related to the service worker. Simply reloading the page doesn't help. Empty cache+hard reload seems to fix it. And if I load it in an incognito window, it works fine.

keanulee:
_Should_ fix #2319 

I suspect the issue occurs when `/app-shell.html` is missing from the precache (maybe it became stale?). By specifically deleting `/app-shell.html` from the cache, I can replicate the `an object that was not a Response was passed to respondWith().` warning.

screen shot 2017-09-21 at 4 23 09 pm

This PR makes sure we gracefully handle that case.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -100,7 +100,18 @@ function writeServiceWorkerFile() {
     // Only serve the app shell for navigate requests (not XHRs) of non-sample content.
     if ((request.mode === 'navigate') && !request.url.match('samples')) {
       return caches.open(cacheName).then(function(cache) {
-        return cache.match(urlsToCacheKeys.get(new URL('/app-shell.html', self.location).toString()));
+        const url = new URL('/app-shell.html', self.location).toString();
+        return cache.match(urlsToCacheKeys.get(url)).then(function(response) {
+          if (response) {
+            return response;
+          }
+          throw Error('/app-shell.html missing from cache.');
+        });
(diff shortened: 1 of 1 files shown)


katejeffreys:
Sooo since you say it _should_ fix the issue, I'm going to guess that it's not testable locally?

Is there a way to confirm that if it doesn't fix the issue it will just do nothing, rather than make anything worse? I am happy to merge if that is the case

@arthurevans @keanulee

keanulee:
Maybe deploy a version to app engine and play around with it? @arthurevans created the issue, so he's best suited to check.

arthurevans:
@keanulee SGTM. Fix looks good, will test tomorrow (new laptop at work today, doesn't have my development environment on it yet).

arthurevans:
Tested, LGTM. Should we merge this first, or the 2.0 update?

keanulee:
Merge order shouldn't matter, so whichever's ready first I guess

PR #2336, opened 2017-09-25
Add page on e2e testing and browser automation

katejeffreys:
First draft of this page. @arthurevans @justinfagnani please comment on what else needs to be included.

### DIFF
--- a/app/2.0/docs/tools/browser-automation.md
+++ b/app/2.0/docs/tools/browser-automation.md
@@ -1,0 +1,16 @@
+---
+title: Test your Polymer app with a browser automation tool
+
+---
+
+<!-- toc -->
+
+Many app developers use browser automation tools such as WebDriver.io and Selenium to help with end-to-end testing. 
+
+Without shadow DOM, you can use [the `querySelector` method](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector) to access an element from the DOM, such as a button, in order to simulate a user action such as a click. However, if the button in question is encapsulated inside shadow DOM, it is out of scope of `querySelector` and will not be accessible.
+
+You may still want to perform end-to-end testing of your Polymer application with a browser automation tool. 
+
+[Chad Killingsworth has written a gist](https://gist.github.com/ChadKillingsworth/d4cb3d30b9d7fbc3fd0af93c2a133a53) with code for a custom WebDriver.io command to retrieve elements from inside shadow DOM. 
+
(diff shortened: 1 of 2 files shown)


katejeffreys:
@arthurevans

arthurevans:
@justin @ChadKillingsworth @ergo @andreasgalster Does this seem like a helpful starting point? (Tagging the folks who were discussing this issue on Slack.)

AndreasGalster:
I tried to get basic WebDriver tests going a while ago, for some reason only Phantom JS worked on my machine, which does fail even the most basic test to check if my app-shell has a specific property set.

I'm guessing it's a good idea to run WebDrvier tests with the new headless Chrome option?

katejeffreys:
@AndreasGalster I will ask @justinfagnani to answer that question.

Did you try the process in this doc? If it didn't work, can you say what happened when you tried it? Also you stated PhantomJS worked, but that it failed a very basic test - does this mean you couldn't use PhantomJS either?

AndreasGalster:
I just tried to load my page and get a property on my app-shell. Page loads but no property available. Even with waiting. I can do some more tests tonight if it's helpful. So yes, ultimately I wasn't able to run with PhantomJS and I couldn't try the process in the doc.

Unfortunately for reasons unknown to me yet all my other browsers time out (even Chrome, so I wouldn't be able to run headless). I haven't been able to figure out what that issue this, though I'm guessing this is unrelated to overall webcomponents/Polymer testing.

katejeffreys:
Thanks for your feedback @AndreasGalster.

I am not sure what the next step is and hoping y'all can help me to work it out. Currently the info I've provided in the proposed document is:

* Why automated tests won't work with Polymer elements
* How to make them work with Chad's custom WebDriver code

From reading Andreas' response, the issues he experienced in attempting this task do not seem to be related to Chad's code.

What further info needs to be provided to enable users to conduct automated tests on Polymer elements?

arthurevans:
We have a conflict here, and we never did get input from eng. I feel like there's a lot of new stuff going on with Puppeteer etc., so we should maybe hit eng. up and see if we should merge this or do something larger for 3.x. Or both.

katejeffreys:
https://github.com/SeleniumHQ/selenium/pull/5762

43081j:
FYI regarding the PR @katejeffreys , it has been merged now.

However, it doesn't introduce an accessor to `shadowRoot` so the gist from chad is still useful (i may one day get around to adding that but its a laborious PR).

The parts it does fix are things like (C# in my examples here) `Element.Text`, `Element.IsVisible()` and other "helper" members. They now traverse shadow dom and follow slots, etc.

PR #2337, opened 2017-09-26
Add `@namespace` and `@memberof` to supported tags

### LINKED ISSUE #2314: "Document your elements" doesn't mention @namespace or @memberof
These annotations are needed to render API docs correctly, and to convert with modulizer.

katejeffreys:
Fixes #2314

### DIFF
--- a/app/2.0/docs/tools/documentation.md
+++ b/app/2.0/docs/tools/documentation.md
@@ -308,6 +308,47 @@ properties and mixins in a table in the main element description:
     `--paper-button-ink-color` | Background color of the ripple | Based on the button's color
     `--paper-button` | Mixin applied to the button | `{}`
 
+### Namespaces and members of a namespace
+
+If you declare a namespace, annotate it with the `@namespace` tag. For example:
+
+```js
+<!-- zombie-base.html -->
+<script>
+  // A bunch of elements, for zombies, by zombies.
+  // @namespace ZombieElements
+  window.ZombieElements = ZombieElements || {};
+
+  // @memberof ZombieElements
+  // @polymer 
+  // @mixinFunction
+  ZombieElements.ZombieMixin = (base) => class extends base {
+    // braaains
+  }
+  // etc
+</script>
+```
+
+Annotate classes, functions and values assigned to a a namespace with the `@memberof` tag:
+
+```js
+<!--conehead-zombie.html -->
+<link rel="import" href="zombie-base.html">
+<script>
+  // A zombie with a cone on his head. Festive!
+  // @polymer
+  // @customElement
+  // @memberof ZombieElements
+  // @appliesMixin ZombieElements.ZombieMixin
+  class ConeheadZombie extends ZombieMixin(Polymer.Element) {
+    static get headProtection() {
+      return 'cone';
+    }
+  }
+  customElements.define('conehead-zombie', ConeheadZombie);
+  ZombieElements.ConeheadZombie = ConeheadZombie;
+</script>
+```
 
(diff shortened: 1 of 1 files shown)


katejeffreys:
@justinfagnani, @arthurevans has raised the question of providing some context for these tags and when a Polymer user would need to use them. Could you outline when a Polymer user would need to use `@namespace` and `@memberof`?

justinfagnani:
The usages are the same as described in the jsdoc docs, here: http://usejsdoc.org/tags-memberof.html and here http://usejsdoc.org/tags-namespace.html

You use `@namespace` when you're declaring a new namespace (which might need a definition in the glossary) and `@memberof` when you're declaring something that is part of a namespace.

katejeffreys:
Thanks @justinfagnani. 

Can we elaborate on when a Polymer user would need to declare a new namespace or declare something that is part of a namespace, in the context of creating a Polymer app or element?

@arthurevans any ideas?

katejeffreys:
@arthurevans

arthurevans:
@justinfagnani Yeah, I'm a little confused here. I haven't seen any use of @namespace or @memberof outside of the Polymer core. In particular, we have a bunch of behaviors with names like `Polymer.FooBarBehavior`, which don't use either @namespace or @memberof, but which document correctly.

Can you give us some examples of when real people would use these annotations for elements, mixins, or behaviors?

justinfagnani:
@arthurevans it may be that we've made the analyzer infer namespaces better over time, but that's fragile. It's also likely that we're just not noticing the lack of `@memberof` because we're looking at projects that write flatly into the `Polymer` namespace and we're not expecting any organization of the API docs via namespace. If a entire _repo_ contains just a single namespace, it's not noticable. I looked at Vaadin elements, and it seems like while they correctly use `@memberof`, they don't use `@namespace` and so their large element collections are not organized by namespace: https://github.com/vaadin/vaadin-grid/blob/master/vaadin-grid-column.html#L16

`@namespace` and `memberof` should be used consistently. When you declare a namespace, like `Polymer`, it should have a `@namespace` annotation, and when you declare a class/function/value that's assigned to a namespace it should have the `@memberof` annotation.

arthurevans:
OK, so to translate:

If I'm just registering an element, I may not care too much.

Like, if I just declare an element inside an IIFE and register it, I'm not creating a namespace nor a member of any namespace.

```
(function() {
  Polymer({ is: 'foo-bar', ... });
  // or...
  class FooBar extends Polymer.Element {
    ...
  }
  customElements.define(FooBar.is, FooBar);
})();
```

But if want to create a namespace, I could do something like:

```
<!-- zombie-base.html -->
<script>
// A bunch of elements, for zombies, by zombies.
// @namespace ZombieElements
window.ZombieElements = ZombieElements || {};

// @memberof ZombieElements
// @polymer 
// @mixinFunction
ZombieElements.ZombieMixin = (base) => class extends base {
  // braaains
}
// etc.
</script>
```

And ...

```
<!--conehead-zombie.html -->
<link rel="import" href="zombie-base.html">
<script>
// A zombie with a cone on his head. Festive!
// @polymer
// @customElement
// @memberof ZombieElements
// @appliesMixin ZombieElements.ZombieMixin
class ConeheadZombie extends ZombieMixin(Polymer.Element) {
  static get headProtection() {
    return 'cone';
  }
}
customElements.define('conehead-zombie', ConeheadZombie);
ZombieElements.ConeheadZombie = ConeheadZombie;
</script>
```

justinfagnani:
yes :)

PR #2338, opened 2017-09-26
How to style undefined elements

katejeffreys:


katejeffreys:
@arthurevans

app/2.0/docs/devguide/style-shadow-dom.md
```diff
@@ -444,6 +443,52 @@ And you can select by slot name:
 </x-foo>
 ```
 
+### Style undefined elements
+
+To avoid FOUC (flash of unstyled content), you might want to style custom elements before they are defined (that is, before the browser has attached their class definition to their markup tag). If you don't, the browser may not apply any styles to the element at first paint.
+
+In CSS you can use the `:defined` pseudo-class selector to target elements that are defined. In this case, styling elements that are not defined uses a simple inversion (`:not(:defined)`):
+
+```css
```
arthurevans: Since we're telling people not to use this, I don't think we should include the sample of the-way-that-doesn't work. Instead, I'd condense these two paras (above and below sample) into one: "There is a specification for a `:defined` pseudo-class selector to target elements that have been defined, but the custom elements polyfill doesn't support this selector."

app/2.0/docs/devguide/style-shadow-dom.md
```diff
@@ -444,6 +443,52 @@ And you can select by slot name:
...
+class myElement extends Polymer.Element(){
+  ...
+  ready(){
+    super.ready();
+    this.removeAttribute('unresolved');
+    ...
+  }
+  ...
+}
+```
```
arthurevans: Maybe add something (here or at the beginning?): Typically, you'll want to add styling for a few, top-level elements so your application's layout displays while the element definitions are being loaded.

katejeffreys: done

PR #2342, opened 2017-09-28
Add preserve-content docs for 1.0. Fixes #1462.

### LINKED ISSUE #1462: Documentation for `preserve-content`?

arthurevans:


app/1.0/docs/devguide/local-dom.md
```diff
@@ -560,4 +560,42 @@ Without empty text nodes:
...
 
 Starting in release 1.8.0, `strip-whitespace` is recursive, affecting any nested `dom-if` and
-`dom-repeat` instances inside the  template.
\ No newline at end of file
+`dom-repeat` instances inside the  template.
+
+## Preserve template contents
+
+Polymer performs one-time processing on your DOM template, such as parsing and removing binding
+annotations and caching  and removing nested template contents for better performance. This processing
```
katejeffreys: >Polymer performs one-time processing on your DOM template, such as parsing and removing binding annotations and caching  and removing nested template contents for better performance.

This is syntactically unclear to me. Probably it is obvious from context, but the multiple 'and's coupled with the fact that 'caching' could be a noun or a verb make it weird. 

i.e. is the correct sense of caching "parsing and removing binding annotations and caching", or "caching and removing nested template contents"?

katejeffreys: also extra space snuck in here: "caching  and removing"

arthurevans: Good catch. Rewrote this as a bulleted list to make it easier to read (here & in 2.0).

PR #2343, opened 2017-10-02
Clarify use of this in addEventListener()

MajorBreakfast:
After watching a recent episode of A11ycasts by Rob Dodson I was surprised that it's not necessary to use `bind()` when passing a callback to `addEventListener()`. So, I checked MDN why that works. I've updated the relevant section in the Polymer documentation to explain this properly.

https://youtu.be/16gvkPfPIx4?t=14m22s
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#The_value_of_this_within_the_handler

### DIFF
--- a/app/2.0/docs/devguide/events.md
+++ b/app/2.0/docs/devguide/events.md
@@ -77,7 +77,7 @@ the element that the event listener is attached to, in this case, the custom ele
 The recommended way for setting up a listener on a child element of the custom element is to
 use an [annotated event listener](events#annotated-listeners) inside the template.
 
-If you need to imperatively set up the listner, it is important to bind the `this` value
+If you need to imperatively set up the listener, it is important to bind the `this` value
 using `.bind()` or using an arrow function.
 
 ```js
--- a/app/2.0/docs/devguide/events.md
+++ b/app/2.0/docs/devguide/events.md
@@ -51,8 +51,8 @@ methods to add and remove event listeners imperatively.
 
 ### Listener on the custom element
 
-Listeners on our custom element can be set up in `ready()` using `this.addEventListener()`.
-The listener will be set up the first time our custom element is attached to the DOM.
+Listeners on a custom element can be set up in `ready()` using `this.addEventListener()`.
+The listener will be set up the first time the custom element is attached to the DOM.
 
 ```js
 ready() {
--- a/app/2.0/docs/devguide/events.md
+++ b/app/2.0/docs/devguide/events.md
@@ -49,16 +49,51 @@ lowercase event names.**
 You can use the standard `addEventListener` and `removeEventListener`
 methods to add and remove event listeners imperatively.
 
+### Listener on the custom element
+
+Listeners on our custom element can be set up in `ready()` using `this.addEventListener()`.
+The listener will be set up the first time our custom element is attached to the DOM.
+
 ```js
 ready() {
   super.ready();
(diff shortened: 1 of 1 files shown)


MajorBreakfast:
@katejeffreys I removed the explanation, because there's no need to bind the `this` using an arrow function or using `.bind()`. The `this ` will be correct due to how `addEventListener` calls the callback. The explanation sounds like it the arrow function is a necessity, but that's simply not true.

jsilvermist:
Though it may be bound to `this` in this case, the majority (in my experience) of event listeners probably won't be.

If this is to be changed I feel you would have to word it quite carefully to avoid confusion further on.

app/2.0/docs/devguide/events.md
```diff
@@ -56,9 +56,21 @@ ready() {
 }
 ```
```
arthurevans: The sample above (with arrow function) should be cut.

app/2.0/docs/devguide/events.md
```diff
@@ -56,9 +56,21 @@ ready() {
...
+ready() {
+  super.ready();
+  this.addEventListener('click', this._onClick);
+}
+
+_onClick(event) { /* ... */ }
+```
+
+The `this` value in `_onClick()` in the example above is the custom element because the callback is
+called with the event's current target as `this`. Also, the use of `ready()` is important, because,
```
arthurevans: The first sentence is kind of half the story—or, to put it another way, it assumes that you _do_ know how currentTarget is defined (i.e., that it's always the node that you added the listener to), but that you _don't_ know that currentTarget and this are the same. 

So we could say here something like:

By default, an event handler is called with the `this` value set to the event's _current target_. The current target is always equal to the element that the event listener is attached to, in this case, the custom element itself.

MajorBreakfast: Good point. I'll rephrase that, so that it's clearer.

MajorBreakfast:
@jsilvermist You're right. I think the solution is to point out more clearly that this section is only concerned with `this.addEventListener()` (with `this` being our custom element). For the other cases are better solutions:
- child elements: Use the `on-myevent="myFn"` syntax inside the template
- other elements: Use the approach presented next in the docs. In this case it is very important to remove the listener on when the element is detached to avoid memory leaks.

MajorBreakfast:
Okay, I think I've addressed all concerns now. Improvements:
- Add a section for each of the three cases (custom element, child element, outside element)
- Clearer explanations for the technical details. Also, I moved them into info boxes.

@arthurevans I've copied your explanation relating to the current target. It's really easy to understand.

katejeffreys:
Minor corrections for style. I like it! Thanks @MajorBreakfast. I will merge if @arthurevans is okay with it.

MajorBreakfast:
Perfect! :)

MajorBreakfast:
Wait I've found a typo...

MajorBreakfast:
@arthurevans Could you have a look?

arthurevans:
At some point we might want update this with other recommendations on when to register (for example, you can use afterNextRender to defer adding listeners until after first paint, and calling afterNextRender from the constructor is probably ~= to calling it from ready. However, this is a big improvement on the previous, so many thanks!

MajorBreakfast:
@arthurevans I'll keep an eye out for that. Thx for merging!

PR #2345, opened 2017-10-02
add info on detail.value

### LINKED ISSUE #2344: "Change notification events" section cut short
https://www.polymer-project.org/1.0/docs/devguide/data-system#data-flow-objects-arrays

"
For an array mutation, the detail.path field is an array mutation path, such as "myArray.splices", and the detail.value
"

That sentence needs finishing - it surely requires significant detail, or a reference to where that detail is elsewhere.

katejeffreys:
Fixes #2344

### DIFF
--- a/app/1.0/docs/devguide/data-system.md
+++ b/app/1.0/docs/devguide/data-system.md
@@ -698,7 +698,9 @@ The contents of the event vary depending on the change.
 *   For a subproperty change, the _path_ to the subproperty is included in the `detail.path` field,
     and the new value is included in the `detail.value` field.
 *   For an array mutation, the `detail.path` field is an array mutation path, such as
-    "myArray.splices", and the `detail.value`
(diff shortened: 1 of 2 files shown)


app/1.0/docs/devguide/data-system.md
```diff
@@ -698,7 +698,9 @@ The contents of the event vary depending on the change.
 *   For a subproperty change, the _path_ to the subproperty is included in the `detail.path` field,
     and the new value is included in the `detail.value` field.
 *   For an array mutation, the `detail.path` field is an array mutation path, such as
-    "myArray.splices", and the `detail.value`
+    "myArray.splices", and the `detail.value` field is a change record, like the one described in the documentation on [array observation](/{{{polymer_version_dir}}}/docs/devguide/observers#array-observation).
+    
+    When you mutate an array, a `.length` change event containing the updated length of the array is always generated, in addition to any other change events (such as `.splices`).
```
arthurevans: To clarify a bit, I'd rephrase this as:

When you mutate an array, Polymer _also_ generates a change event for the array's `length` property (for example, `detail.path` is "myArray.length" and `detail.value` is the new length of the array).

PR #2353, opened 2017-10-04
Remove exhortation to use tap instead of click

### LINKED ISSUE #2352: Gesture events recommended or not ?
It's not clear if gesture events are recommended or not.

In https://www.polymer-project.org/2.0/docs/devguide/gesture-events, it is written 
> Polymer provides optional support for custom "gesture" events for certain user interactions. These events fire consistently on both touch and mouse environments, so we recommend using these events instead of their mouse- or touch-specific event counterparts. 

But in a recent PR (#2283), tip about using `tap`  rather the `click` was removed from the documentation.

Also, this issue https://github.com/PolymerElements/app-layout/issues/465 recommends using native `click`.

As gesture events are an option in Polymer 2.x,  maybe they should not be recommended in the documentation.

katejeffreys:
Using tap instead of click is no longer required. Removed advice to use it. Also changed examples to use other events than tap. Fixes #2352

### DIFF
--- a/app/2.0/docs/devguide/gesture-events.md
+++ b/app/2.0/docs/devguide/gesture-events.md
@@ -5,10 +5,7 @@ title: Gesture events
 <!-- toc -->
 
 Polymer provides optional support for custom "gesture" events for certain user
-interactions. These events fire consistently on both touch and mouse environments,
-so we recommend using these events instead of their mouse- or
-touch-specific event counterparts. This provides better interoperability with both touch and
-mouse devices.
+interactions. 
 
 ## Using gesture events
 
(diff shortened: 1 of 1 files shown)


app/2.0/docs/devguide/gesture-events.md
```diff
@@ -5,10 +5,7 @@ title: Gesture events
 <!-- toc -->
 
 Polymer provides optional support for custom "gesture" events for certain user
-interactions. These events fire consistently on both touch and mouse environments,
```
arthurevans: I think this is still true for the up, down, and track events...

azakus:
LGTM. There's still some nuance that could be mentioned for the other gestures, but I'm not sure where that would go.

arthurevans:
@azakus Specifically, should we be cutting the text at the top that says, 

"These events fire consistently on both touch and mouse environments, so we recommend using these events instead of their mouse- or -touch-specific event counterparts. This provides better interoperability with both touch and mouse devices."

katejeffreys:
@azakus @arthurevans any issues with me merging?

azakus:
@arthurevans yeah, I think that's fine to remove that text.

PR #2355, opened 2017-10-05
No template[is=dom-repeat], especially document-level

keanulee:
This fixes the empty search results page. This change has already been deployed.


jsilvermist:
I understand document level, but why is this necessary inside `demo-tabs`?

keanulee:
@jsilvermist It's not strictly necessary (since Polymer will convert `<template is="dom-repeat">` to `<dom-repeat>` automatically), and we generally recommend against it if you need to support hybrid mode and if you use it inside certain tags (like `<table>` and `<select>`). In this case though it's 2.0 only and isn't in those tags so it works as well.

https://www.polymer-project.org/2.0/docs/upgrade#remove-type-extension-elements

PR #2358, opened 2017-10-12
Fix search results in shady dom

keanulee:
I noticed that the search results don't show on Shady DOM (e.g. Firefox) if you initiate a search from outside the search page (e.g. the homepage). This should fix it. I made a separate `<pw-search-results>` element that is lazy-loaded because eventually I want to move some of the search logic from `<pw-shell>` here; for now this just fixes a more pressing issue.

Staged at https://2017-10-12-search-dot-polymer-project.appspot.com


katejeffreys:
Works beautifully, thanks so much! LGTM

keanulee:
Deployed

PR #2361, opened 2017-10-19
Search element

keanulee:
Building off of #2358, this change moves most of the search functionality to the lazy-loaded `<pw-search-results>` element. This reduces the complexity of `<pw-shell>`.

Staged at https://search-dot-polymer-project.appspot.com/search/?q=css


arthurevans:
:heart: it. We've come full-circle back to a search-results element!

PR #2362, opened 2017-10-20
update tests.md/safari safaridriver needs argument

eskan:
safaridriver needs one argument at least : **--enable** allows sessions to run without authenticate next time


ghost (commented):
`--enable` is not an option for my version (10.1.2)

This thread suggests it becomes an option from 11 onwards: https://stackoverflow.com/questions/44006628/how-to-enable-allow-remote-automation-in-safari-progrmmatically

`/usr/bin/safaridriver -p 0` maybe?

PR #2370, opened 2017-10-30
Add info on polymer serve --compile

### LINKED ISSUE #2369: polymer serve (compile)
I've been using an overcomplicated workflow specifically when wanting to serve for ie11:

- `polymer build`
- `polymer serve build/es5-bundled`

Then I happened to read some documentation that is present on the polymer cli repo's README...

>By default, the server will automatically use Babel to transpile any ES6 code down to ES5 for browsers that don't have native support for important ES6 features like classes. This behavior can be explicitly turned on/off for all browsers via the --compile option.

Should this be added to the `polymer serve` documentation? Pehaps along with an example in the "server options" section that demonstrates the --compile flag?

Thanks

katejeffreys:
Fixes #2369

### DIFF
--- a/app/2.0/docs/tools/polymer-cli-commands.md
+++ b/app/2.0/docs/tools/polymer-cli-commands.md
@@ -230,19 +230,48 @@ App project demo:
 
 This section shows examples of using various `polymer serve` options.
 
+#### --port
+
 Serve from port 3000:
 
     polymer serve --port 3000
+
+#### --hostname 
 
 If you have configured a custom hostname on your machine, Polymer CLI can serve it with the
 `--hostname` argument (for example, app project demo is available at `http://test:8080`):
 
     polymer serve --hostname test
 
+#### --open
+
 Open up a page other than the default `index.html` in a specific browser
 (Apple Safari, in this case):
 
     polymer serve --open app.html --browser Safari
+
+#### --compile
+
+By default, the server will automatically use Babel to transpile any ES6 code down to ES5 for
+browsers that don't have native support for important ES6 features like classes. This behavior
+can be explicitly turned on or off for all browsers via the `--compile` option.
+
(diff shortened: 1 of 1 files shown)


app/2.0/docs/tools/polymer-cli-commands.md
```diff
@@ -230,20 +230,49 @@ App project demo:
...
+
+Always compile ES6 to ES5:
+
+    polymer serve --compile always
+    
+Never compile ES6 to ES5:
+
+    polymer serve --compile never
+
+Automatically compile to ES5 for browsers that don't fully support ES6:
```
arthurevans: Should probably add (default).

katejeffreys: `never` is the default, right?

app/2.0/docs/tools/polymer-cli-commands.md
```diff
@@ -230,20 +230,49 @@ App project demo:
...
 (Apple Safari, in this case):
 
     polymer serve --open app.html --browser Safari
 
+#### --compile
+
+By default, the server will automatically use Babel to transpile any ES6 code down to ES5 for
+browsers that don't have native support for important ES6 features like classes. This behavior
+can be explicitly turned on or off for all browsers via the `--compile` option.
+
```
arthurevans: Might add here or after the --never argument:

Automatic compilation can cause problems when running in device emulation mode on Chrome. If the browser sends another browser's user-agent string, the server may switch between compiled and uncompiled responses, leaving the browser with an inconsistent set of resources in its cache. This issue may also occur in other browsers with device emulation capabilities. Use `--compile always` or `--compile never` to avoid this problem.

app/2.0/docs/tools/polymer-cli-commands.md
arthurevans: No, `auto` is the default.

PR #2373, opened 2017-11-03
Pre-3.0 cleanup.

arthurevans:


arthurevans:
Modernize some stuff: mention setting property defaults in the constructor; update DOM templating examples to handle superclasses with string templates. Remove the example that showed a vanilla custom element because it confused some. Correct the example that included `afterNextRender` to show the required import.

app/2.0/docs/devguide/dom-template.md
```diff
@@ -94,20 +94,30 @@ important rules:
...
     static get template() {
       if (!memoizedTemplate) {
-        // create a clone of superclass template (`true` = "deep" clone)
-        memoizedTemplate = MySuperClass.template.cloneNode(true);
+        // get the superclass template, which could be 
+        // either a string or a template element
+        const superTemplate = MySuperClass.template;
+        if (typeof superTemplate == 'string') {
+          // create a new template element
+          memoizedTemplate = document.createElement('template);
```
abdonrd: Missed close `'` :)

app/2.0/docs/devguide/custom-elements.md
```diff
@@ -159,25 +129,28 @@ For a complete list of limitations, see [Requirements for custom element constru
...
+``` 
 
 The `Polymer.Element` class initializes your element's template and data system during the `ready`
-callback, so if you override ready, you must call `super.ready()` at some point.
+callback, so if you override `ready`, you must call `super.ready()` at some point.
+
+Polymer does several things at `ready` time:
+
+-   Sets default property values.
+-   Allows observers and computed properties to run (as soon as their dependencies are defined).
```
kevinpschaaf: In 2.0+ it's technically "as soon as _any_ of their dependencies are defined"

app/2.0/docs/devguide/dom-template.md
```diff
@@ -94,20 +94,30 @@ important rules:
...
-  class MyExtension extends MySuperClass {
+  class SimpleExtension extends MySuperClass {
     static get template() {
       if (!memoizedTemplate) {
-        // create a clone of superclass template (`true` = "deep" clone)
-        memoizedTemplate = MySuperClass.template.cloneNode(true);
+        // get the superclass template, which could be 
+        // either a string or a template element
+        const superTemplate = MySuperClass.template;
+        if (typeof superTemplate == 'string') {
```
kevinpschaaf: @sorvell @justinfagnani Have you discussed and/or hit this issue yet in 2.0 → 3.0 migration?  Post-`modularizer`, any subclassers that extended the super template under the assumption that it was an `HTMLTemplateElement` would be broken unless they used this type of super-defensive code (which I'm sure most would not have).  Most of our elements wouldn't have hit this yet since they're all still hybrid, so that won't be a good litmus test.  Seems like we kinda screwed up allowing a subclassable field to have an ambiguous type. 🤦‍♂️

Maybe you've already discussed this and have a plan? They'll be 3.0 elements so maybe breaking subclassers is just ok?

Is it worth considering having `modularizer` do a safer thing by generating code that creates a template element rather than just a string:

```js
static get template() {
  return makeTemplate(`<-- (template html) -->`);
}
```

Or we could go the other way and make a breaking change to 3.0 by saying template strings have to be defined via e.g. `static get templateString()`, and `static get template()` is always a `HTMLTemplateElement` (and `modularizer` would generate `templateString`).

app/2.0/docs/devguide/dom-template.md
```diff
@@ -94,20 +94,30 @@ important rules:
 
 -   Don't modify the superclass template in place; make a copy before modifying.
 -   If you're doing anything expensive, like copying or modifying an existing template,
-    you should memoize the modified template so you don't have to regenerate it when the
+    you should memoize the modified template so you don't have to regenerate it each time the
     getter is called.
+-   Consider that the superclass template getter could return either a string _or a `<template>` element.
```
katejeffreys: needs terminating underscore

app/2.0/docs/devguide/dom-template.md
justinfagnani: @kevinpschaaf @sorvell can we chat about this in person?

app/2.0/docs/devguide/custom-elements.md
arthurevans: OK, I think I was a little confused. I thought it was waiting until default values were set, but after experimenting a little more, it appears that defaults are now set in `constructor`, so we're just guaranteeing that the observers won't run until after that's happened, right?

Do we set properties (i.e., from initial bindings or attributes) before we let the observers fire?

arthurevans:
Addressed feedback. Backed out all of the changes to the DOM templating chapter until we have a resolution on template handling from @justinfagnani @kevinpschaaf @sorvell.

arthurevans:
Addressed some of @kevinpschaaf's feedback in https://github.com/Polymer/docs/pull/2373/commits/af57caf3ff6524e216788c29087ba8c1928ed0e2. 

@katejeffreys can I get an LGTM for the updated PR and I'll go ahead and merge ...

katejeffreys:
LGTM.

PR #2380, opened 2017-11-08
Update quick-tour.md

spyhole:
delete useless item


arthurevans:
The comma is actually supposed to be there (because it's adding a new item to the array literal). @katejeffreys this looks quite confusing. Do you think it might help to insert the new item *before* the last item, so it could be:

`{first: 'Shawna', last: 'Williams'},` 

Instead? Might cause less confusion than the trailing comma.

(Also, we should probably change out these names at some point.)

katejeffreys:
closing - replaced by #2400

PR #2385, opened 2017-11-16
Use `polymer install`, not `bower install`

mroderick:
Installing `polymer-cli` doesn't give you a `bower` in your path, but it does give you `polymer`


mroderick:
I signed it!

katejeffreys:
@arthurevans was there an original rationale for using `bower install` over `polymer install`?

PR #2389, opened 2017-11-17
Add missing backticks

justincy:
`<example-card>` was being interpreted literally as DOM which didn't render anything and made that sentence read poorly.


katejeffreys:
Thanks!

PR #2390, opened 2017-11-18
Add plunker for unresolved style example

katejeffreys:


app/2.0/docs/devguide/style-shadow-dom.md
```diff
@@ -515,6 +515,8 @@ class myElement extends Polymer.Element(){
 }
 ```
 
+[See an example on Plunker](http://plnkr.co/edit/vWS9PIBTTbKHxRnJyT7T?p=preview)
+
```
arthurevans: Note that there's an error in this plunker: The <style> tag is between <head> and <body>, but it should be inside the head.

katejeffreys: Thanks! Updated it.

PR #2401, opened 2017-11-28
Update the Node versions we support.

aomarks:
Because it's already out of date.


justinfagnani:
This looks good as it removes outdated information, but I wonder if there's a place where we should keep the list of current versions.

katejeffreys:
@aomarks thanks for being on top of this :) I will try to be proactive about the release of the next version.

aomarks:
Should I update this with the specific versions?

PR #2408, opened 2017-12-06
Update first element tutorial for p3

katejeffreys:
Staged here: https://polymer-3-first-element-dot-polymer-project.appspot.com/3.0/start/first-element/intro


katejeffreys:
See also https://github.com/PolymerLabs/polymer-3-first-element

PR #2410, opened 2017-12-06
Update code samples and links to new plunkers

katejeffreys:
https://quick-tour-code-samples-dot-polymer-project.appspot.com/3.0/start/quick-tour
@keanulee @arthurevans 
Hoping for some help in figuring out why the code samples don't render correctly?


app/3.0/start/quick-tour.md
```diff
@@ -38,15 +50,24 @@ an element name with a class, so you can add properties and methods to your cust
...
-  </demo-tab>
-  <demo-tab heading="index.html">
-<pre><code>{{{include_file('2.0/start/samples/custom-element/index.html')}}}</code></pre>
-  </demo-tab>
+<demo-tabs selected="0" name="qt-1-register" src="http://plnkr.co/edit/r1bwLm?p=preview">
+  <paper-tab slot="tabs">custom-element.js</paper-tab>
+  <div>
+
+```html
+<!-- include_file 3.0/start/samples/custom-element/custom-element.js -->
```
keanulee: Try opening the code block with "\```js" or "\```javascript" instead

katejeffreys: Does not fix the rendering issue.. but the js is nicely highlighted now, thanks :) 

Re-uploading. Would you mind taking another look?

app/3.0/start/samples/custom-element/index.html
```diff
@@ -1,10 +1,11 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
-    <script src="https://polygit.org/components/webcomponentsjs/webcomponents-loader.js"></script>
-    <link rel="import" href="custom-element.html">
+    <script src="https://polygit.org/webcomponentsjs/components/webcomponentsjs/webcomponents-loader.js"></script>
```
keanulee: wc-loader URL should not change.

JS files (like custom-element.js) might not be copied to dist/, which may be why the demos aren't working (I see a bunch of 404s)

katejeffreys: oh yeah that makes sense (both things). Thank you! Fixing the wc-loader links now.

@arthurevans @keanulee what would be the correct process to get the samples into the distribution properly? I tried pasting one into the appropriate location in the dist/ folder but the sample still didn't render, idk

keanulee: Add to `gulp.task('copy', ...)`, something like:

```js
  const jsSamples = gulp.src([
      'app/**/*.js',
      '!app/{bower_components,elements}/**',
      '!app/3.0/samples/homepage/**', // excluded since homepage samples are put through polymer build - see demo1 below
      '!app/js/**/*.js'
    ], {base: 'app/'})
    .pipe(gulp.dest('dist'));

  return merge(app, docs, /* ... */, jsSamples);
```

keanulee: Hmm, seems like there's more config to do. I'll take a look a push a commit when I figure it out

katejeffreys: <3 <3 <3

keanulee: Give that a try. The editable-name sample is broken because unpkg won't redirect deps to pre-release versions (but you probably already knew that).

katejeffreys: omg, it's doing the thing

Thank you!!!!

katejeffreys:
I have done another pass of the quick tour for p3 - could use a sanity check @arthurevans. Also wrote a new sample to illustrate 2-way property binding because unpkg.com doesn't like our import paths and I couldn't get the `<iron-input>` example to work.

Staged here:

https://quick-tour-code-samples-dot-polymer-project.appspot.com/3.0/start/quick-tour#bind-to-a-property

Issue #2407, opened 2017-12-05
Document the new html template tag


justinfagnani:
See https://github.com/Polymer/polymer/issues/4937 and https://github.com/Polymer/polymer/pull/4971#pullrequestreview-81150361

justinfagnani:
I'm anticipating possible confusion when documenting the ability to have expressions in Polymer templates - people may wonder if they can do data-binding there, or how it relates to expressions in something like lit-html. One suggestion for explaining this is to emphasize the `static`-ness of the property/getter and clarify that the expressions are evaluated just once when the class is defined.

PR #2412, opened 2017-12-07
Html helper

katejeffreys:
Staged here: 

https://html-helper-dot-polymer-project.appspot.com/2.0/docs/devguide/dom-template

### DIFF
--- a/app/2.0/docs/devguide/dom-template.md
+++ b/app/2.0/docs/devguide/dom-template.md
@@ -4,9 +4,11 @@ title: DOM templating
 
 <!-- toc -->
 
-Many elements use a subtree of DOM elements to implement their features. DOM templating provides an easy way to create a DOM subtree for your element.
+Many elements use a subtree of DOM elements to implement their features. DOM templating provides 
+an easy way to create a DOM subtree for your element.
 
-By default, adding a DOM template to an element causes Polymer to create a shadow root for the element and clone the template into the shadow tree.
+By default, adding a DOM template to an element causes Polymer to create a shadow root for the 
+element and clone the template into the shadow tree.
 
 The DOM template is also processed to enable data binding and declarative event handlers.
 
@@ -14,11 +16,11 @@ The DOM template is also processed to enable data binding and declarative event 
 
 Polymer provides three basic ways to specify a DOM template:
 
--   [Specify a template using the `<dom-module>` element](#dommodule). This allows you to specify the
-    template entirely in markup, which is most efficient for an element defined in an HTML
+-   [Specify a template using the `<dom-module>` element](#dommodule). This allows you to specify
+    the template entirely in markup, which is most efficient for an element defined in an HTML
     import.
 -   [Provide a template by defining a template property on the constructor](#templateobject).
--   [Inherit and optionally modify a template from an existing base class](#inherited-templates).
+-   [Inherit and optionally modify a template from an existing element](#inherit).
 
 Polymer provides a default `template` getter that retrieves a template from the element's
 `<dom-module>`. You can override this getter to provide a string template or a generated template
@@ -69,7 +71,16 @@ As an alternative to specifying the element's template in markup, you can define
 `template` property. 
 
 One way to do this is to specify a string template by creating a static template getter that 
-returns a string. This getter is called _once_, when the first instance of the element is upgraded.
+returns a template literal. 
+
+Template literals are denoted by being enclosed in back-ticks (<code>` `</code>). Polymer 2.4+ uses
+the functionality of embedded expressions in template literals to provide convenient ways to 
+[extend inherited templates](#inherited-templates). [Read more about template literals on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals).
+
+To create a static template getter, override Polymer's default `template` getter. This getter is
+called _once_, when the first instance of the element is upgraded.
+
+Specify a string template in a static template getter: { .caption }
 
 ```js
 class MyElement extends Polymer.Element {
@@ -92,7 +103,9 @@ However, the tag name still needs to be passed as the first argument to
 {.alert .alert-info}
 
 Polymer 2.4+ provides the `Polymer.html` function that can be applied to a template literal to
-produce an `HTMLTemplateElement`. For example:
+produce an `HTMLTemplateElement`. 
+
+Use the Polymer.html function to convert a template literal to an HTMLTemplateElement: { .caption }
 
 ```js
 class MyElement extends Polymer.Element {
@@ -110,7 +123,9 @@ customElements.define('my-element', MyElement);
 
 Some text editors support HTML code highlighting in template literals tagged with a function
 called `html`. To enable HTML code highlighting in such text editors, declare a constant called
-`html` to hold the `Polymer.html` function. Use the `html` constant in the template getter:
+`html` to hold the `Polymer.html` function. Use the `html` constant in the template getter.
+
+Declare a constant called html to hold the Polymer.html function: { .caption }
 
 ```js
 const html = Polymer.html;
@@ -126,23 +141,24 @@ class MyElement extends Polymer.Element {
 customElements.define('my-element', MyElement);
 ```
 
-### Inherited templates {#inherited-templates}
+### Inherited templates {#inherit}
 
-You can inherit an existing template by extending a base class with a template. This
-functionality provides for a variety of scenarios: 
+An element that extends another Polymer element can inherit its template. If the element doesn't
+provide its own DOM template (using either a `<dom-module>` or a static template object), Polymer
+uses the same template as the superclass, if any.
 
-* To [inherit a base class template with **no modifications**](#nomods), do not supply a
-  template definition in the child class declaration.
-* To [**override** a template definition](#overridebase), supply your own template definition 
-  in the child class.
-* To [**extend** a base class template](#extendbase), supply a template in the child class that includes
-  a reference to the base class template.
-* To [**insert** content from a child class into placeholders supplied by a base class](#insertcontent),
-  create placeholders in the base class and supply the content with functions.
+* [Polymer 2.x: Inherit a base class template **without modifying it**](#nomods)
+* [Polymer 2.x: **Override** a base class template in a child class](#overridebase)
+* [Polymer 2.x: **Modify** a copy of a superclass template](#modify)
+* [Polymer 2.4+: **Extend** a base class template in a child class](#extendbase)
+* [Polymer 2.4+: Provide **template extension points** in a base class for content from a child class](#insertcontent)
 
-#### Inherit a base class template with no modifications {#nomods}
+#### Polymer 2.x: Inherit a base class template without modifying it {#nomods}
 
-Base class definition { .caption }
+To inherit a base class template without modifying it, do not supply a template definition in the
+child class declaration.
+
+Base class definition: { .caption }
 
 ```html
 <dom-module id="base-class">
@@ -159,7 +175,7 @@ Base class definition { .caption }
 </dom-module>
 ```
 
-Child class definition { .caption }
+Child class definition: { .caption }
 ```html
 <dom-module id="child-class">
 
@@ -175,15 +191,16 @@ Child class definition { .caption }
 
 [See a working example in Plunker](http://plnkr.co/edit/vS99al?p=preview).
 
-#### Override a template definition {#overridebase}
+#### Polymer 2.x: Override a base class template in a child class {#overridebase}
 
-To override a base class's template definition inside a `<dom-module>`, without extending it, create a
-`<dom-module>` for your child class, and write a template for it:
+To override a base class's template definition, supply your own template for your child class.
+
+Base class definition: { .caption }
 
 ```html
 <dom-module id="base-class">
 
-  <template>This content has been inherited from BaseClass.</template>
+  <template>This is BaseClass's template.</template>
 
   <script>
     class BaseClass extends Polymer.Element {
@@ -195,7 +212,7 @@ To override a base class's template definition inside a `<dom-module>`, without 
 </dom-module>
 ```
 
-Override the base class template with a child class dom-module { .caption }
+Child class definition: { .caption }
 
 ```js
 <dom-module id="child-class">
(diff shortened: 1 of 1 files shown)


katejeffreys:
Fixes #2407

justinfagnani:
LGTM!

My only small nit is not with this change, but how we talk about properties like `template` and `properties`.  We always call them the "static X getter", but really, a getter is just one way to implement the property.

You could also just set the property on the constructor:

```js
class MyElement {}
MyElement.template = html`...`;
```

Or if you using a compiler, and in the near future in native JS, declare a static property:

```js
class MyElement {
  static template = html`...`;
}
```

katejeffreys:
This is a great point & something I was processing this morning after
talking to Arthur about templates in a static context.

I created an issue in the docs to consider more general ways of referring
to `template` and `properties`. #2414

Is it okay to merge now, and address this later? It goes deeper than this
issue, as you said, and could be addressed throughout the site.

justinfagnani:
Oh yeah, this looks good to me as-is, I just wanted to mention that nit since it's been on my mind for a while.

katejeffreys:
Sweet, thanks Justin. I'll get Arthur to give it a look, then merge if he's cool with it.

app/2.0/docs/devguide/dom-template.md
```diff
@@ -14,18 +14,18 @@ The DOM template is also processed to enable data binding and declarative event
...
--   `<dom-module>` element. This allows you to specify the template entirely in markup,
-    which is most efficient for an element defined in an HTML import.
--   Provide a string template. This works well for elements defined entirely in JavaScript
-    (for example, in an ES6 module).
--   Retrieve or generate your own template element. This allows you to define how the template is
-    retrieved or constructed, and can be used to modify a superclass template.
+-   [Specify a template using the `<dom-module>` element](#dommodule). This allows you to specify the
+    template entirely in markup, which is most efficient for an element defined in an HTML
+    import.
+-   [Assign a string template to an HTMLTemplateElement object, and return the
```
arthurevans: I like the new links.

I think the description of this one focuses too much on the implementation. The gist here is that you can define a `template` getter that returns the template. 

Except that @justinfagnani raised a concern about the wording of that. Um, so,...

"Add a template property on the constructor that returns the template?"

"Provide a template by defining a template property on the constructor"?

app/2.0/docs/devguide/dom-template.md
```diff
@@ -60,106 +60,184 @@ except for testing. For caveats about defining elements in the main document,
...
-       <div>I've got a string template!</div>`
+       <div>I've got a string template!</div>`;
   }
 }
 customElements.define('my-element', MyElement);
 ```
 
-When using a string template, the element doesn't need to provide an `is` getter (however, the tag
-name still needs to  be  passed as the first argument to `customElements.define`).
+Alternatively, declare a variable called `html` to hold the `Polymer.html` object, and
```
arthurevans: I feel like this should be listed goal-first. It's a lot of text to read through to figure out that you'd want to do this to enable syntax highlighting.

app/2.0/docs/devguide/dom-template.md
```diff
@@ -60,106 +60,184 @@ except for testing. For caveats about defining elements in the main document,
 see [main document definitions](registering-elements#main-document-definitions).
 {.alert .alert-info}
 
-### Specify a string template
+### Return an HTMLTemplateElement {#templateobject}
 
-As an alternative to specifying the element's template in markup, you can specify a string template
-by creating a static `template` getter that returns a string.
+As an alternative to specifying the element's template in markup, you can create a static
+`template` getter that returns an HTMLTemplateElement.
```
arthurevans: HTMLTemplateElement should be in code format.

The magic `Polymer.html` should be introduced, like,  "Polymer provides an `html` function that can be applied to a template literal to produce an `HTMLTemplateElement`." Or something @justinfagnani ? 

Unfortunately, ECMAScript kind of failed when it comes to terminology here. `foo` is a template literal. html`foo` is a _tagged template literal_. What exactly the `html` function is in this syntax is not specified, AFAICT, in the ECMAScript spec. It's... the function that you tag the literal with. Or something.

Sometimes I hate engineers. Just sayin'.

Anyway, point is, I'm basically just specifying a string, but presto change-o, I return a full-blown template element.

arthurevans:
Few comments in today. I have to run, will finish up tomorrow. I think we should iterate a little more here...

katejeffreys:
@arthurevans Great points, thank you. Incorporated all. Let's iterate until it's ready~

app/2.0/docs/devguide/dom-template.md
```diff
@@ -60,106 +59,183 @@ except for testing. For caveats about defining elements in the main document,
...
+* To [**extend** a base class template](#extendbase), supply a template in the child class that includes
+  a reference to the base class template.
+* To [**insert** content from a child class into placeholders supplied by a base class](#insertcontent),
+  create placeholders in the base class and supply the content with functions.
 
--   Don't modify the superclass template in place; make a copy before modifying.
--   If you're doing anything expensive, like copying or modifying an existing template,
-    you should memoize the modified template so you don't have to regenerate it when the
-    getter is called.
+#### Inherit a base class template with no modifications {#nomods}
```
arthurevans: These examples should use dom-module, since it's the default mechanism for Polymer 2.x.

katejeffreys:
http://plnkr.co/edit/f5utiX?p=preview does not work

katejeffreys:
Neither does http://plnkr.co/edit/YQ5t5I?p=preview

app/2.0/docs/devguide/dom-template.md
katejeffreys: Converted back to dom-module examples - a couple of issues with the following

http://plnkr.co/edit/f5utiX?p=preview
http://plnkr.co/edit/YQ5t5I?p=preview

@arthurevans @justinfagnani

app/2.0/docs/devguide/dom-template.md
```diff
@@ -60,106 +59,183 @@ except for testing. For caveats about defining elements in the main document,
 see [main document definitions](registering-elements#main-document-definitions).
 {.alert .alert-info}
 
-### Specify a string template
+### Provide a template by defining a template property on the constructor {#templateobject}
 
-As an alternative to specifying the element's template in markup, you can specify a string template
-by creating a static `template` getter that returns a string.
+As an alternative to specifying the element's template in markup, you can define a static `template`
+property. The `template` property must be an `HTMLTemplateElement`.
```
arthurevans: must be => should be

This section is correct for 3.x, but for 2.x we support returning a string, as well.

Maybe in place of the second sentence, add something to the next para, like,

In Polymer 2.x, the template can be supplied as an `HTMLTemplateElement` instance or a string. Support for returning a string template is being discontinued in Polymer 3.x. Starting in Polymer 2.4 <<<verify actual release when it lands>>>, Polymer provides ... This is the recommended way of defining a string template in Polymer 2.4 and later.

app/2.0/docs/devguide/dom-template.md
```diff
@@ -60,106 +59,183 @@ except for testing. For caveats about defining elements in the main document,
...
-       <div>I've got a string template!</div>`
+       <div>I've got a string template!</div>`;
   }
 }
 customElements.define('my-element', MyElement);
 ```
 
-When using a string template, the element doesn't need to provide an `is` getter (however, the tag
-name still needs to  be  passed as the first argument to `customElements.define`).
+To enable HTML code highlighting in text editors with lit-html code highlighting functionality, you
```
arthurevans: "with lit-html code highlighting functionality" => I gather this isn't strictly true; some just have support for this built-in according to @justinfagnani ... Also, somewhat confusing to mention this in this context, because the `html` already _looks_ kind of like lit-html, and we don't want people to confuse the two.

"some text editors" is kind of lame. 

the full explanation is probably a bit wordy, would probably be something like,
"Some text editors support HTML code highlighting in template literals tagged with a function named `html`." Again, @justinfagnani to confirm perhaps?

app/2.0/docs/devguide/dom-template.md
```diff
@@ -60,106 +59,183 @@ except for testing. For caveats about defining elements in the main document,
...
+can declare a variable called `html` to hold the `Polymer.html` object, and return it from a `template` getter:
```
arthurevans: Two details: `Polymer.html` is a function, and `html` here is a constant. 

"And return it from the template getter..." sounds like you're returning the function. Use it in the template getter, perhaps?

app/2.0/docs/devguide/dom-template.md
```diff
@@ -60,106 +59,183 @@ except for testing. For caveats about defining elements in the main document,
...
+  static get template() {
+    return html`
+      This content has been inherited from BaseClass.
+      `;
+  }
+}
+```
+
+Child class definition { .caption }
+```js
```
arthurevans: Arguably this is simpler than using the existing template manipulation code, but it involves some magic that should probably be explained. 

1) The `${expression}` inside a template literal interpolates the value of an expression. (This is basic ES6, and maybe it doesn't need to be spelled out, but if so, we should at least link to the MDN description of template literals at the first reference (in the string template section), because we're using a whole bunch of their features here and not everyone is familiar with them yet.)
2)  Because the getter is static, `super.template` accesses the template property on the superclass constructor.
3) The `Polymer.html` function checks each expression value. If the value is a string, the string is interpolated directly. If the expression is an `HTMLTemplateElement`, the `innerHTML` of the template is interpolated. This makes it very simple to combine existing templates.

arthurevans: And this is a lot of words, and maybe should be pushed up to some concepts section somewhere, but I think it's important because we're using a lot of ES6-fu here that seems completely transparent to the @justinfagnani s of the world...

app/2.0/docs/devguide/dom-template.md
```diff
@@ -60,106 +59,183 @@ except for testing. For caveats about defining elements in the main document,
...
+    ${super.template}
+    <p>Hello again from ChildClass.</p>
+    `;
+  }
+}
 ```
 
+[See a working example in Plunker](http://plnkr.co/edit/f5utiX?p=preview).
+
+#### Insert content from a child class into placeholders supplied by a base class {#insertcontent}
```
arthurevans: I'm not sure this is the correct title for this use case... It seems like the primary use case here is the _superclass_ use case: that is, "provide overrideable sub-templates." Or something.  Replaceable sub-templates? Allow easy template extension with modular templates?

@justinfagnani @kevinpschaaf @graynorton Anyone want to name this pattern? :D 

Just seems to me that the important part is, a superclass providing a focused way to extend its template with content. They could be placeholders, or not—for example, they could be a valid part of the superclass template that a subclass may or may not replace.

arthurevans: Also: We may want to leave in one or both of the old-school template extension examples, since they're still valid and ```html`${foo}` ``` will only work in 2.4 and later.

app/2.0/docs/devguide/dom-template.md
arthurevans: I meant that specifically for the "inherit a base class with no modifications."

The `${expr}` format is specific to template literals (those things in ``` `backticks` ```) and won't work in `<template>` elements... despite the names, they're rather different creatures.

The extension examples you added with the `html` function and template literals are much clearer than the previous ones that require manipulating template elements in JS. And for that reason I think we might want to keep them in. However, they'll only work in 2.4+, so we might want to leave in at least one of the previous examples.

I had some more specific comments on this stuff inline.

app/2.0/docs/devguide/dom-template.md
kevinpschaaf: * "Template extension points"
* "Overridable sub-templates"

I'd probably go with the first...?

katejeffreys:
@arthurevans @justinfagnani Revised the page based on feedback. Please have another look. https://www.polymer-project.org/2.0/docs/devguide/dom-template is updated with the revised version.

justinfagnani:
@katejeffreys the one question I have is whether or not it's confusing to mention plain string templates at all. The "Specify a string template in a static template getter" example will be deprecated in the next release. Maybe remove that example and add a note that says something like "Note: It used to be possible to specify a plain string template as a static property on your element class, but this feature was deprecated in 2.? and will be removed in Polymer 3.0"

app/2.0/docs/devguide/dom-template.md
```diff
@@ -76,19 +91,148 @@ class MyElement extends Polymer.Element {
...
+customElements.define('my-element', MyElement);
+```
+
+### Inherited templates {#inherit}
 
 An element that extends another Polymer element can inherit its template. If the element doesn't
-provide its own DOM template (using either a `<dom-module>` or a string template), Polymer uses the
-same template as the superclass, if any.
+provide its own DOM template (using either a `<dom-module>` or a static template object), Polymer
+uses the same template as the superclass, if any.
```
arthurevans: I think this list of items needs some introduction, like: 

The `Polymer.html` helper introduced in Polymer 2.4 makes template extension a little simpler, so this section shows examples with `Polymer.html` (usable on 2.4 and later) and without (usable on all 2.x versions).

Or... something.

arthurevans:
Ping @katejeffreys on the updates here. As soon as we have those updates we can merge this for the 2.4 release and cut a 2.x branch.

katejeffreys:
Thanks for the reminder! Will do.

app/2.0/docs/devguide/dom-template.md
```diff
@@ -4,28 +4,29 @@ title: DOM templating
...
 
 Polymer provides three basic ways to specify a DOM template:
 
--   `<dom-module>` element. This allows you to specify the template entirely in markup,
-    which is most efficient for an element defined in an HTML import.
--   Provide a string template. This works well for elements defined entirely in JavaScript
-    (for example, in an ES6 module).
--   Retrieve or generate your own template element. This allows you to define how the template is
-    retrieved or constructed, and can be used to modify a superclass template.
+-   [Specify a template using the `<dom-module>` element](#dommodule). This allows you to specify
```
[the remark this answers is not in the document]
katejeffreys: - [ ] update headings and links to make parallel & simpler

app/2.0/docs/devguide/dom-template.md
```diff
@@ -53,19 +54,33 @@ Example: { .caption }
...
 **Note:** Elements should generally be defined outside of the main document,
 except for testing. For caveats about defining elements in the main document,
 see [main document definitions](registering-elements#main-document-definitions).
 {.alert .alert-info}
 
-### Specify a string template
+### Provide a template by defining a template property on the constructor {#templateobject}
+
+As an alternative to specifying the element's template in markup, you can define a static 
+`template` property.
```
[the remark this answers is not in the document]
katejeffreys: - [ ] static template property => template property on the constructor.

app/2.0/docs/devguide/dom-template.md
```diff
@@ -53,19 +54,33 @@ Example: { .caption }
...
 {.alert .alert-info}
 
-### Specify a string template
+### Provide a template by defining a template property on the constructor {#templateobject}
+
+As an alternative to specifying the element's template in markup, you can define a static 
+`template` property. 
+
+One way to do this is to specify a string template by creating a static template getter that 
+returns a template literal.
```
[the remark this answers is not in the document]
katejeffreys: - [ ] clarify intro to this section
- [ ] make sure terminology is correct
- [ ] subsection: 2.4+ 
- [ ] subsection: 2.x all

justinfagnani:
@katejeffreys did you see https://github.com/Polymer/polymer/pull/5023

We're not going to allow arbitrary values for interpolations into templates. Only other templates.

katejeffreys:
Thx for heads-up @justinfagnani. Editing now.

app/2.0/docs/devguide/dom-template.md
katejeffreys: - [ ] intro to list

katejeffreys:
@arthurevans @justinfagnani This is ready for another look.

app/2.0/docs/devguide/dom-template.md
```diff
@@ -65,42 +61,32 @@ except for testing. For caveats about defining elements in the main document,
...
+    // Polymer 2.4+: Return HTMLTemplateElement 
   }
 }
-
-customElements.define('my-element', MyElement);
 ```
 
-[See a working example in Plunker](http://plnkr.co/edit/AzkMDJ?p=preview).
+Polymer 2.4+ provides a helper function (`Polymer.html`) to generate an `HTMLTemplateElement`
+instance from a template literal.
```
justinfagnani: I'd do "JavaScript template literal" to be extra clear, since we're also talking about HTML templates too

app/2.0/docs/devguide/dom-template.md
```diff
@@ -141,12 +127,41 @@ class MyElement extends Polymer.Element {
...
+#### Return a string template (all 2.x versions) {#stringtemplate}
+Template literals are denoted by being enclosed in back-ticks (<code>` `</code>). 
+
+Specify a string template in a static template getter: { .caption }
+
+```js
+class MyElement extends Polymer.Element {
+
+  static get template() {
+    return `<style>:host { color: blue; }</style>
```
justinfagnani: Doesn't this need to be:

```js
  return html`<style>:host { color: blue; }</style>
```

katejeffreys: @justinfagnani per @arthurevans' feedback, I left that example in for users of pre-2.4 versions. 

I may have misunderstood which forms are available in which versions, though. My understanding was that in 2.4+, the string template return value is deprecated but still usable. For that reason the heading indicates that the example applies to "all 2.x versions" - would you mind clarifying on that point?

@arthurevans WDYT?

katejeffreys:
This is almost ready to merge, I think. I need a couple points of clarification though:

* Which versions the string template is still a possible return value of `static get template(){...}` (i.e. 2.x or only < 2.4) 
* Whether to include the string template example in the docs (my preference is to keep it)

@arthurevans @justinfagnani

kevinpschaaf:
@katejeffreys 

> * Is a string template still a possible return value of `static get template(){...}` in 2.4+?

It is deprecated (a warning will print), but still allowed.  The intention is to actually remove support before the final 3.0 release is tagged.

> * Should we keep the string template example in the docs for users of pre-2.4 versions?

I'd say no, let's switch to using `Polymer.html` (maybe with a note that this is available in 2.4)?

arthurevans:
@kevinpschaaf @katejeffreys For the string template, I think we should avoid giving people the impression that they need to change this now if we're going to provide a we're going to provide an automated transition to 3.0. Plus, some people won't upgrade to 2.4 right away. So we should definitely provide the alternative.

How about we show the Polymer.html version first, then note: `Polymer.html` was added in release 2.4. For earlier versions, you can return a plain string:

```
    return `<div>A plain string template</div>`;
```

Returning a string is deprecated in favor of the `html` helper in release 2.4. Support for the string template will be removed in 3.0.

... Or words to that effect.

katejeffreys:
@arthurevans @kevinpschaaf @justinfagnani Updated and staged here: https://html-helper-dot-polymer-project.appspot.com/2.0/docs/devguide/dom-template#templateobject

PR #2413, opened 2017-12-07
Update style-shadow-dom.md

fincha:
Added an important note for <custom-style>


arthurevans:
I think you may be seeing an issue fixed in Polymer 2.3.1: https://github.com/Polymer/polymer/pull/4979

If that's not the case, please open an issue on the Polymer repo for the problem you're seeing. Hybrid mode should be working fine in the latest version of Polymer, and if it doesn't, it's a serious bug.

arthurevans:
Closing this docs PR. Please file an issue a library issue if you're still seeing this problem.

PR #2420, opened 2017-12-21
Add information concerning .length observable path

Flamenco:


katejeffreys:
@Flamenco looks like a useful addition IMO, but maybe could be in a different section in the doc. I'll take a look. @arthurevans fyi

katejeffreys:
Replacing this PR with #2636

katejeffreys:
@Flamenco pls see #2636 which has edits to your PR, and also modifies 3.0 pages. Apologies for the delay.

PR #2433, opened 2018-01-10
Add autobasepath to Polymer.json docs

### LINKED ISSUE #2425: Document new autoBasePath feature.
https://github.com/Polymer/polymer-project-config/pull/48 (already released) adds a new top-level `polymer.json` option called `autoBasePath`. When `true`, it will automatically set `basePath: true` on all of the project's builds. The idea is that setting that option on each build individually is kind of awkward, since you should usually want it set on all of them. We should update https://www.polymer-project.org/2.0/docs/tools/polymer-json to mention this new option.

katejeffreys:
Fixes #2425 

& also the markdown renders, which helps

### DIFF
--- a/app/2.0/docs/tools/polymer-json.md
+++ b/app/2.0/docs/tools/polymer-json.md
@@ -208,6 +208,9 @@ for every run:
 
     Note that `basePath` must be set to `true` if using [prpl-server](https://github.com/Polymer/prpl-server-node).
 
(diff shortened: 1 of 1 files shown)


katejeffreys:
@aomarks Pls use this PR instead - sorry for the inconvenience

aomarks:
Np, lgtn

PR #2440, opened 2018-01-17
Just a typo fix

h-nasu:


katejeffreys:
Replacing this PR with #2458 because CLAs. Thanks for spotting the typo @h-nasu :)

PR #2445, opened 2018-01-27
Add polymer 2.4 release notes

katejeffreys:


kevinpschaaf:
Actually... do we usually include a list of all the bugfixes in the release notes?  I believe there were a few that were at least on people's radar.

katejeffreys:
Yes we do, good point. These are the ones I can see that I think should be
included:

https://github.com/Polymer/polymer/pull/5062 fixes
https://github.com/Polymer/polymer/issues/5060
https://github.com/Polymer/polymer/pull/4991 fixes
https://github.com/Polymer/polymer/issues/5047
https://github.com/Polymer/polymer/pull/5048 fixes
https://github.com/Polymer/polymer/issues/4983

any others?

PR #2453, opened 2018-02-05
Tools docs: Fix doc comments in class mixin section

MajorBreakfast:
These comments seem weird to me. I think they need to be fixed.


MajorBreakfast:
👍

PR #2470, opened 2018-02-13
Fix code indentation

repinel:


repinel:
CLA should be good now.

PR #2476, opened 2018-02-20
update shadow dom docs for p3 - first pass

katejeffreys:


app/3.0/docs/devguide/shadow-dom.md
```diff
@@ -321,10 +332,10 @@ For more details, see [Working with slots in JS](https://developers.google.com/w
 The `Polymer.FlattenedNodesObserver` class provides utilities to track an element's _flattened tree_.
 That is, a list of the node's child nodes, with any `<slot>` elements replaced by their distributed
 nodes. `FlattenedNodesObserver` is an optional utility that can be loaded from
-`lib/utils/flattened-nodes-observer.html`.
+`lib/utils/flattened-nodes-observer.js`.
 
-```html
-<link rel="import" href="/bower_components/polymer/lib/utils/flattened-nodes-observer.html">
+```js
+import * from '@polymer/polymer/utils/flattened-nodes-observer.js';
```
katejeffreys: check this

katejeffreys:
probably needs some iteration but ready for a look @arthurevans https://p3-shadow-dom-concepts-dot-polymer-project.appspot.com/

app/3.0/docs/devguide/style-shadow-dom.md
```diff
@@ -278,110 +302,121 @@ You can create **slots** in an element's template that are populated at runtime.
...
+
+```js
+...
+static get template() {
+  return html`
     <style>
       h1 ::slotted(h1) {
         font-family: sans-serif;
-        color: green;
+        color:green;
```
arthurevans: for consistency, there should be a space between the property and the value (`color: green;`) here and below (`color: blue;`).

app/3.0/docs/devguide/style-shadow-dom.md
```diff
@@ -429,260 +471,273 @@ And you can select by slot name:
...
+</body>
 ```
 
+[See it on Plunker](http://plnkr.co/edit/hR3I4w?p=preview)
+
 ## Share styles between elements
 
 ### Use style modules {#style-modules}
 
 The preferred way to share styles is with *style modules*. You can package up styles in a style module, and share them between elements.
```
arthurevans: Going forward, we're probably going to recommend using template mechanism rather than style modules. We should add that as a separate update after this one, though, I think.

app/3.0/docs/devguide/style-shadow-dom.md
```diff
@@ -429,260 +471,273 @@ And you can select by slot name:
...
-<dom-module id="new-element">
-  <template>
-    <style include="my-style-module">
-      <!-- Any additional styles go here -->
-    </style>
-    <!-- The rest of your element template goes here -->
-  </template>
-</dom-module>
+```js
+  static template get() {
```
arthurevans: `static get template()`

app/3.0/docs/devguide/style-shadow-dom.md
```diff
@@ -429,260 +471,273 @@ And you can select by slot name:
...
-<x-foo></x-foo>
+<custom-style>
+  <style>
+    /* Document-level styles go here */
+  </style>
+</custom-style>
 ```
 
-## Use `custom-style` in document-level styles {#custom-style}
+`custom-style` is not included in the `@polymer/polymer/polymer-element.js` package. Import `custom-style` from `@polymer/polymer/lib/elements/custom-style.js`:
```
arthurevans: package => module

app/3.0/docs/devguide/style-shadow-dom.md
```diff
@@ -429,260 +471,273 @@ And you can select by slot name:
...
+      /* Document-level styles go here */
+    </style>
+  </custom-style>
+</head>
+<body>
+  <custom-element></custom-element>
+</body>
+```
 
 *Note: You should only use `custom-style` to define styles for the main document. To define styles for an element's local DOM, just use a `<style>` block.*
```
arthurevans: Fix note formatting: Only bold run-in head, add `{.alert .alert-info}` ... Extra credit, replace Note with a more informative run-in so people know whether they should read it. Like:

**Don't use custom-style inside an element's template.**

also: local DOM => shadow DOM

app/3.0/docs/devguide/style-shadow-dom.md
```diff
@@ -429,260 +471,273 @@ And you can select by slot name:
...
 
 ### Examples
 
-In the first code sample, the style for the `p` element “leaks” into Paragraph B in browsers that haven’t implemented the Shadow DOM v1 specs. In the second code sample, the developer has used `custom-style` to wrap the style block, preventing this leak.
+In the following code sample, the document-level style in index.html "leaks" into the shadow DOM of `<custom-element>` in browsers that haven’t implemented the Shadow DOM v1 specs.
```
arthurevans: Note for the future: at some point (not right now) we should clarify what we mean by "leaking." It could be confusing, because inheritance is also a form of "leaking," right?

Specifically, what we mean when we say styles can't leak is that _a selector in one scope (i.e., in the main document) can't match an element in a different scope (like in a shadow root)... Except in a couple of very specific  instances, namely :host and ::slotted() ...

Anyway, think on it... We should convey that at some point.

katejeffreys: agree

app/3.0/docs/devguide/style-shadow-dom.md
katejeffreys: ok, created #2513 to track

katejeffreys:
@arthurevans Thanks! Incorporated feedback

katejeffreys:
@arthurevans would you mind taking another look - there were a couple of missing parts https://github.com/Polymer/docs/pull/2476/commits/d9d4ff6473c85393def6d755d7745b661f3d9d63

PR #2479, opened 2018-02-26
Add paths-to-names blog post.

arthurevans:


app/blog/2018-02-26-3.0-preview-paths-and-names.md
```diff
@@ -0,0 +1,159 @@
...
+*   You can already preview code using bare module specifiers using `polymer serve`: 
+    
+    ```bash
+    polymer serve --module-resolution=node
+    ```
+
+Our support for bare specifiers is independent of our other build steps, like bundling. This step simply resolves specifiers and rewrites them as paths—`polyserve` will still serve modules, and `polymer-build` will output modules for browsers that support them. Bundling isn't required. 
+
+
+### Renaming Polymer.Element
```
kevinpschaaf: This section seems out of place since it's not related to the path->name transition.  How about we nix the section and instead just talk about this in the actual release blog post and/or notes?

PR #2485, opened 2018-03-01
Polycast link update

e111077:
I'm the polycast now!

Tangentially I also released a polycast that goes through installing node + polymer CLI + vs code extensions (Will have to make another one for 3.0 when that comes out)


arthurevans:
LGTM2. But should we maybe add @e111077's name in here too?

arthurevans:
Learn Polymer with Elliott Marquez and Rob Dodson?

e111077:
Rob is bowing out of Polycasts, but he has so far done 63/65 episodes. Perhaps we should also put a reminder to phase it out? Is it okay if we merge this one and I can do that in another PR?

e111077:
see #2488

PR #2486, opened 2018-03-01
3.0-ify devguide/model-data

katejeffreys:
https://data-binding-dot-polymer-project.appspot.com

### DIFF
--- a/app/3.0/docs/devguide/model-data.md
+++ b/app/3.0/docs/devguide/model-data.md
@@ -154,8 +154,8 @@ Example { .caption }
 [See it on Plunker](https://plnkr.co/edit/3LTcb2?p=info)
 
 ```js
-import { Element as PolymerElement, html } from '@polymer/polymer/polymer-element.js';
-import '@polymer/polymer@3.0.0-pre.10/lib/elements/dom-repeat.js';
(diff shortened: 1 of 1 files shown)


app/3.0/docs/devguide/model-data.md
```diff
@@ -151,33 +151,26 @@ Every Polymer element has the following array mutation methods available:
 
 Example { .caption }
 
-```html
-<link rel="import" href="components/polymer/polymer-element.html">
-<link rel="import" href="components/polymer/src/elements/dom-repeat.html">
+[See it on Plunker](https://plnkr.co/edit/3LTcb2?p=info)
```
arthurevans: Plunker doesn't quite match the sample below (maybe that doesn't matter? I found it a little confusing, though).

Code style:

Order for functions should be:
- static functions (get template, etc)
- constructor (if defined)
- other functions (lifecycle callbacks, etc.)

For the plunker, specifically:

More idiomatic to set the default values in the constructor, rather than `ready`.

If you already have a `reset` method, you can just call it from the constructor, instead of setting the value twice.

```
constructor() { 
  super();
  this._reset();
}
```

Like yonder: https://plnkr.co/edit/BeOdBr?p=preview

PR #2494, opened 2018-03-12
3.0-ify devguide/data-binding

katejeffreys:
https://data-binding-dot-polymer-project.appspot.com/

### DIFF
--- a/app/3.0/docs/devguide/data-binding.md
+++ b/app/3.0/docs/devguide/data-binding.md
@@ -12,7 +12,7 @@ or _target element_). The host element data can be a property or sub-property re
 You create data bindings by adding annotations to an element's local DOM template.
 
 ```js
-static get template(){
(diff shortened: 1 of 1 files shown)


app/3.0/docs/devguide/data-binding.md
```diff
@@ -460,36 +490,30 @@ backslash (`\`).
...
-    class XCustom extends Polymer.Element {
-
-      static get is() {return 'x-custom'}
-
-      doThisOnce: function() {
-        return Math.random();
-      }
-
-    }
+[See it in Plunker](https://plnkr.co/edit/iWFpsh?p=preview)
```
arthurevans: Plunker here still has the name being set in index.html, which doesn't cause any problems, but is kind of confusing since this element doesn't do anything with the name.

PR #2506, opened 2018-03-26
fix typos in setter names

bashmish:
I spent an hour debugging this being not able to find the setter. Let's fix this typo.


bashmish:
Maybe docs for Polymer 1 and 3 needs to be fixed as well.

PR #2530, opened 2018-04-19
3.0-ify tools/reusable-elements.md

### LINKED ISSUE #2522: 3.0-ify app/3.0/docs/tools/reusable-elements.md

katejeffreys:
Fixes #2522 

Note: Removed instructions about gh-pages demo until I can get it working with paths etc


arthurevans:
Just checked--this is not in the left nav for 2.x. So the 2.x file should be deleted, and we can dispatch with this one.

PR #2533, opened 2018-04-24
temporarily remove stuff about publishing

katejeffreys:
Will be able to put this page back in when webcomponents.org can publish npm modules https://github.com/webcomponents/webcomponents.org/issues/1145


app/3.0/docs/tools/publish.md
```diff
@@ -4,33 +4,6 @@ title: Publish your elements
...
-## Host your elements privately with indie-catalog
-
-If you have elements that can't be made public, you can host them locally with [indie-catalog](https://github.com/PolymerLabs/indie-catalog).
-
-
-
-
+Publish your elements on [WebComponents.org](https://webcomponents.org) to make them available on the Internet.
 
+**Instructions TBD.** This page will be updated when WebComponents.org is able to host elements published on npm.
```
arthurevans: Looks like it's supported at least minimally now. So you can just point people at:

https://www.webcomponents.org/publish

For now. Maybe add a note that support for npm-packaged elements in wc.org is new and incomplete. Demos and API docs don't work at this point.

arthurevans:
Ignore comments since we're hiding this page. Merging.

PR #2534, opened 2018-04-25
3.0-ify polymer-cli-commands.md

katejeffreys:


app/3.0/docs/tools/polymer-cli-commands.md
```diff
@@ -87,6 +110,18 @@ A filename. This is the main entrypoint into your application for all routes. Of
 your `index.html` file. This file should import the app shell file specified in the
 [`shell`](#shell) option. It should be minimal since it's loaded and cached for each route.
 
+### --fragment {#fragment}
+
+This flag supports dynamic dependencies. It is an array of any HTML filenames that are not
+statically linked from the app shell (that is, imports loaded on demand by `importHref`).
```
bicknellr: I think `importHref` has been replaced with dynamic import (`import(url).then(...)`). Also, @usergenic mentioned to me the other day that build can detect dynamic dependencies automatically iff the url passed to `import()` is a static string.

bicknellr: Also, this section still talks about these imports as if they were HTML imports; maybe I'm missing something about what this field does?

katejeffreys: >this section still talks about these imports as if they were HTML imports; maybe I'm missing something about what this field does?

nah i just missed this bit :D thanks

katejeffreys: @usergenic can you clarify how the `--fragment` flag and `fragments` array in Polymer.json currently work?

katejeffreys:
Requesting review from @usergenic

katejeffreys:
- [ ] usergenic review
- [ ] update fragments thing based on info from usergenic
- [ ] remove refs to HTML imports

app/3.0/docs/tools/polymer-cli-commands.md
justinfagnani: They're the same as before, but now JS files to be loaded by dynamic `import()` (or any other means really, like a `<script>` tag.

app/3.0/docs/tools/polymer-cli-commands.md
```diff
@@ -163,21 +196,23 @@ See also:
 
 ## polymer install {#install}
 
-Installs Bower dependencies. Running `polymer install` is equivalent to running `bower install`.
+Installs dependencies.
 
-The `--variants` flag allows you to install dependency variants. See the documentation on [managing dependencies for hybrid elements](/{{{polymer_version_dir}}}/docs/devguide/hybrid-elements#dependency-variants) for more information.
+If the --npm flag is given, running `polymer install` is equivalent to running `npm install`:
```
justinfagnani: Uh... we need to double check this. No one's worked on `polymer install` for npm, we don't have variants for npm, and I'm inclined to just deprecate the command.

app/3.0/docs/tools/polymer-cli-commands.md
```diff
@@ -232,11 +267,47 @@ App project demo:
...
+Sets npm mode. Dependencies are installed from npm, the component directory is set to `node_modules` and the package name is read from `package.json`.
+
+```
+polymer serve --npm --module-resolution="node"
+```
+
+### --module-resolution
 
-Serve from port 3000:
+Specifies how to resolve module specifiers in import and export statements when rewriting them to be web-compatible.
```
justinfagnani: This could use a link to something that describes that web-compatible imports are path-only, though I don't have a great reference in mind right now.

app/3.0/docs/tools/polymer-cli-commands.md
```diff
@@ -232,11 +267,47 @@ App project demo:
...
 
-    polymer serve --port 3000
+Valid values are `"none"` and `"node"`. Defaults to `"none"`. 
+
+* `"none"` disables module specifier rewriting. 
+
+* `"node"` uses Node.js resolution to find modules.
```
justinfagnani: We should mention that because Polymer projects: elements, LitElement, etc., use bare specifiers, that `--module-resolution=node` is generally required, and that it's much easier to just put these in polymer.json.

app/3.0/docs/tools/polymer-cli-commands.md
```diff
@@ -232,11 +267,47 @@ App project demo:
...
+
+### --compile
+
+Compiler options. Valid values are `"auto"`, `"always"` and `"never"`. Defaults to `"auto"`.
+
+`"auto"` compiles JavaScript to ES5 for browsers that don't fully support ES6.
+
+### --component-dir, -c
+
+The component directory to use. When `--npm` is true, defaults to `"node_modules"`.
```
justinfagnani: Might as well mention that without `--npm` is defaults to `bowser_components`

app/3.0/docs/tools/polymer-cli-commands.md
```diff
@@ -232,11 +267,47 @@ App project demo:
...
+
+`"auto"` compiles JavaScript to ES5 for browsers that don't fully support ES6.
+
+### --component-dir, -c
+
+The component directory to use. When `--npm` is true, defaults to `"node_modules"`. 
+
+### --port, -p
+
+To serve from port 3000:
```
justinfagnani: By default an open port is chosen.

app/3.0/docs/tools/polymer-cli-commands.md
keanulee: Since `import('./path.js')` is automatically detected now, we don't use the `fragments` config any more in 3.0 apps.

app/3.0/docs/tools/polymer-cli-commands.md
katejeffreys: maybe http://2ality.com/2017/05/es-module-specifiers.html#how-do-module-specifiers-change-with-esm, wdyt?

app/3.0/docs/tools/polymer-cli-commands.md
katejeffreys: https://github.com/Polymer/tools/blob/master/packages/cli/README.md#polymer-install---variants

https://github.com/Polymer/tools/pull/303

@justinfagnani Looks like it has been done. I will update the docs to include the info about `polymer install` unless there's some reason I shouldn't.

cc @bicknellr

app/3.0/docs/tools/polymer-cli-commands.md
```diff
@@ -277,9 +233,11 @@ Sets npm mode. Dependencies are installed from npm, the component directory is s
 polymer serve --npm --module-resolution="node"
 ```
 
+Because the dependencies for Polymer 3.0 projects are defined in `package.json`, the `--npm` flag is usually required. You will probably find it easier to [add this option to your polymer.json configuration](polymer-json).
```
arthurevans: I don't know how we want to phrase this, but it appears that this should not be required for most cases anymore.

(It's only required if they're using npm, and using old-style element project URLs that access peer dependencies using '../' instead of by name, like '@polymer/polymer/...')

app/3.0/docs/tools/polymer-cli-commands.md
```diff
@@ -277,9 +233,11 @@ Sets npm mode. Dependencies are installed from npm, the component directory is s
...
+
 ### --module-resolution
 
-Specifies how to resolve module specifiers in import and export statements when rewriting them to be web-compatible.
+Specifies how to resolve module specifiers in import and export statements when rewriting them to URLs.
 
 Valid values are `"none"` and `"node"`. Defaults to `"none"`.
```
arthurevans: Now defaults to `node`. This isn't a breaking change for 1.x or 2.x projects, so most people should not need to use this flag on any version.

app/3.0/docs/tools/polymer-cli-commands.md
bicknellr: Yes, `polymer install --npm` just runs `npm install` for you and all the other flags *should* do nothing.

app/3.0/docs/tools/polymer-cli-commands.md
```diff
@@ -299,22 +259,26 @@ Compiler options. Valid values are `"auto"`, `"always"` and `"never"`. Defaults
 
 ### --component-dir, -c
 
-The component directory to use. When `--npm` is true, defaults to `"node_modules"`. 
+The component directory to use. When `--npm` is true, defaults to `"node_modules"`. Without the `--npm` flag, defaults to `"bower_components"`.
```
arthurevans: This is useful for 1.x and 2.x element projects, which use ../ URLs to access peer dependencies. 3.x projects should use names to access peer dependencies, so they don't require the same path rewriting as 1.x and 2.x element projects.

app/3.0/docs/tools/polymer-cli-commands.md
katejeffreys: I am not clear on what this flag does or when to use it :/ here is a guess: 

`--npm` is required when: 

* you are using npm to install dependencies, and
* you are importing dependencies by path (e.g. `'../node_modules/@polymer/...'`) instead of by module specifiers

If you are using npm and importing by module specifiers ('@polymer/polymer/...'), you don't need this flag.

katejeffreys:
@arthurevans @justinfagnani needs another review for recent changes

Issue #2379, opened 2017-11-07
API doc generation for Polymer 3.0


justinfagnani:
We likely need a new module page at the least, and if we're ambitious we can look at using TypeDoc with ou TypeScript declarations and migrate away from iron-doc-viewer.

PR #2535, opened 2018-04-25
3.0 api docs

rictic:
Fixes https://github.com/Polymer/docs/issues/2379

Needs https://github.com/Polymer/tools/pull/220 in order to produce this output, though we could merge this before that lands if we needed to.


scripts/generate_api_docs_3.js
```diff
@@ -0,0 +1,501 @@
+/**
```
justinfagnani: license

scripts/generate_api_docs_3.js
```diff
@@ -0,0 +1,501 @@
...
+/** @typedef {format.Element|format.Class|format.Function|format.ElementMixin} SimpleFeature */
+/** @typedef {Map<string, {allFeatures: SimpleFeature[], analysis: format.Analysis}>} FeatureIndex */
+
+/** @param {format.Analysis} analysis */
+function indexFeaturesByFile(analysis) {
+  /** @type {FeatureIndex} */
+  const map = new Map();
+  /** @param {SimpleFeature} feature */
+  function index(feature, kind) {
+    const filename = getFilename(feature) || '???';
```
justinfagnani: warn?

scripts/generate_api_docs_3.js
```diff
@@ -0,0 +1,501 @@
...
+<section anchor-id="${subsection}">
+  <h2>${subsection}</h2>
+
+  ${filenamesIn.map(fn => getFilenameDescription(fn, featureIndex)).join('\n\n')}
+</section>
+`
+}
+
+const hardcodedDescriptions = new Map([
+  ['lib/elements/array-selector.js', `
```
justinfagnani: What's the way forward here? Can you open issues in Analyzer and Polymer to add these to the source?

scripts/generate_api_docs_3.js
```diff
@@ -0,0 +1,501 @@
...
+    throw new Error(`Got no features for filename ${filename}`);
+  }
+  const features = result.allFeatures;
+  const shortName = filename.split('/').slice(2).join('/');
+  let description;
+  if (features.length === 1) {
+    const [feature] = features;
+    description = feature.summary;
+  }
+  description = description || hardcodedDescriptions.get(filename) || `TODO(write me): ${filename}`;
```
justinfagnani: TODO or just warn?

Maybe these debug defaults should only be written with a flag on.

rictic:
All comments addressed. https://github.com/Polymer/tools/issues/241 is the path forward for file overviews, and instead of producing ??? or TODO in the final output, it throws instead.

arthurevans:
A few thoughts:

Current file overview is a little confusing. Especially weird to have the tiny import line be the first thing on the page, followed by a giant headline for the element/class. This might look better once the `@fileoverview` tag is implemented.

Current format doesn't seem to document symbols that are re-exported. For example, the `polymer-element Module` page doesn't show the `html` function.

My first thought was that it was weird to have the element/class embedded directly in the module page , but I guess the vast majority of our modules only export one thing. The challenge here is that if a single module exported two elements, or an element and some functions, it would be virtually impossible to figure out that the other items were there unless you knew to look for them.

I think this issue could be addressed with a quick-reference at the beginning of the file, so you have something like:

File overview description

import {} from '@polymer/polymer/polymer-element.js' 

Exports: PolymerElement, html, fooBar

Where each item in the Exports list is a link to the corresponding item.

But this might look weird if almost all of our files have only a single export. 

A TOC or some other navigation aid would serve the same purpose of orienting people in the file (in fact, it'd be even better for large elements/mixins with lots of props and methods). Maybe we can get someone with a design-y eye to look at it.

As far as the URLs go, the current URL fragments assume that each module is only exporting a single class/element. For example, if you had a single module that exported simple-dialog and fancy-dialog, each with an `open` method, the fragments for both would look identical:

my-dialogs#method-open

Not sure if I am borrowing trouble here. But I'd hate to rewrite all of the links to the API docs and then change the URLs, so let's discuss this assumption.

arthurevans:
Opened https://github.com/Polymer/docs/issues/2541 to track suggestions above.

PR #2536, opened 2018-04-25
3.0-ify create an app tutorial

### LINKED ISSUE #2515: 3.0-ify app/3.0/docs/tools/create-app-polymer-cli.md

katejeffreys:
fixes #2515


app/3.0/start/toolbox/set-up.md
```diff
@@ -35,9 +35,10 @@ To install Polymer CLI:
 
         polymer init
 
-    Press the down arrow until `polymer-2-starter-kit` is highlighted and press the enter / return
+    Press the down arrow until `polymer-3-starter-kit` is highlighted and press the enter / return
     key to select.
 
+Polymer CLI downloads and installs dependencies, and creates an app from the `polymer-3-start-kit template`.
 
 ## Serve your project
```
bicknellr: I think we might want to change this section to show `polymer serve` without `--open` and mention that the user should go to the url shown after `applications:`. `polymer serve --open` opens the component-style URL which doesn't work anymore.

app/3.0/start/toolbox/set-up.md
```diff
@@ -35,9 +35,10 @@ To install Polymer CLI:
 
         polymer init
 
-    Press the down arrow until `polymer-2-starter-kit` is highlighted and press the enter / return
+    Press the down arrow until `polymer-3-starter-kit` is highlighted and press the enter / return
     key to select.
 
+Polymer CLI downloads and installs dependencies, and creates an app from the `polymer-3-start-kit template`.
```
bicknellr: `` `polymer-3-start-kit template` `` -> `` `polymer-3-starter-kit` template``

app/3.0/start/toolbox/create-a-page.md
```diff
@@ -97,107 +95,157 @@ you need to add it to your app's HTML.
...
-    ```
-    "fragments": [
-      "src/my-view1.html",
-      "src/my-view2.html",
-      "src/my-view3.html",
-      "src/my-new-view.html",
-      "src/my-view404.html"
-    ]
-    ```
+Your new page is now ready! Serve your app with `polymer serve --open`.
```
bicknellr: (I left another comment on the setup page about this.)

app/3.0/start/toolbox/add-elements.md
```diff
@@ -15,25 +15,25 @@ to some off-the-shelf components, for example from [webcomponents.org][webcompon
...
+the npm package name for the component.
 
 In this step, you'll add Polymer's `<paper-checkbox>` element to your app, which is listed on
-[webcomponents.org][paper-checkbox].  You can use Bower to install it.
+[webcomponents.org][paper-checkbox].  You can use Yarn to install it.
 
 Run this command from your project root directory:
 
-    bower install --save PolymerElements/paper-checkbox
+    yarn add --flat @polymer/paper-checkbox
```
bicknellr: Are we recommending yarn? I was never really sure where we landed on yarn vs. npm but I've been using `npm` lately. Also, we may need to explicitly point out that this doesn't work at the moment unless you install `@polymer/paper-checkbox@next` (or some other tag pointing at a 3.0 prerelease version).

bicknellr: Also, I just discovered that `polymer serve` seems to cache failed module resolution results. So if the user forgets to install or installs the wrong version and it fails to resolve the first time the specifier is encountered, then they will continue to see that error until they restart the server - even if they get it correctly installed later.

This was a problem for me here since I tried installing `@polymer/paper-checkbox` (without `@next`) which has a year-old release available (v0.0.4) that looks like Polymer 2 code: it has `paper-checkbox.html` but not `paper-checkbox.js`, causing the specifier lookup to fail.

app/3.0/start/toolbox/deploy.md
```diff
@@ -13,39 +13,89 @@ Type `polymer build` to build your Polymer application for production.
...
-* An unbundled, minified build with a service worker. ES6 code is served as-is. This build is for browsers that support HTTP/2 push.
+* `esm-bundled`: A bundled, minified build with a service worker. ES6 code is served as-is. This build is for browsers that can handle ES6 modules and dynamic imports.
 
-In this step, you'll deploy the bundled, compiled build (`es5-bundled`) for maximum compatibility. Serving the other builds requires a more complex serving setup.
+* `es6-bundled`: A bundled, minified build with a service worker. ES6 modules are compiled to AMD modules.
 
-Builds are configured in the `builds` object in `polymer.json`, a configuration file in the root project folder:
+* `es5-bundled`: A bundled, minified build with a service worker. ES6 code is compiled to ES5 for older browsers.
+
+In this step, you'll deploy `es6-bundled`. Builds are configured in the `builds` object in `polymer.json`, a configuration file in the root project folder:
```
keanulee: (Following points not finalized yet, so no action needed now, but FYI)

- `esm-bundled` - JS code is served without compilation; ES modules are used
- `es6-bundled` - JS code is compiled to ES2015; AMD modules are used
- `es5-bundled` - JS code is compiled to ES5; AMD modules are used

`es5-bundled` output is broken as of cli@pre.14, but we expect it to work soon. Perhaps the deploy instructions should continue to use this build.

app/3.0/start/toolbox/deploy.md
```diff
@@ -13,39 +13,89 @@ Type `polymer build` to build your Polymer application for production.
...
+  },
+  "css": {
+    "minify": true
+  },
+  "html": {
+    "minify": true
+  },
+  "bundle": {
+    "inlineScripts": false
+  },
```
keanulee: @bicknellr I recently changed our repos to use `"bundle": true` for all builds instead of the options object (unnecessary). This and PSK should do the same.

app/3.0/start/toolbox/add-elements.md
bicknellr: (filed as https://github.com/Polymer/tools/issues/238)

app/3.0/start/toolbox/deploy.md
```diff
@@ -88,31 +138,31 @@ and create a new project.
...
+      upload: build/es6-bundled/service-worker.js
       secure: always
 
     - url: /.*
-      static_files: build/es5-bundled/index.html
-      upload: build/es5-bundled/index.html
+      static_files: build/es6-bundled/index.html
+      upload: build/es6-bundled/index.html
       secure: always
     ```
```
keanulee: Suggest you add `skip_files` to app.yaml to avoid uploading unused files (we are doing this to our pwa-starter-kit):

```
skip_files:
- build/es5-bundled/
- build/esm-bundled/
- images/
- node_modules/
- src/
- test/
```

(if necessary, replace `build/es5-bundled/` with whatever build _isn't_ used above)

app/3.0/start/toolbox/add-elements.md
bicknellr: @keanulee [says that we're recommending `npm` now](https://github.com/Polymer/docs/pull/2539#discussion_r184543074).

app/3.0/start/toolbox/deploy.md
bicknellr: Yes, this is definitely critical! Just did this and found that it uploaded 26000 objects thanks to `node_modules/`, which is greater than the 10000 max allowed to deploy!

bicknellr: I wonder if we should tell users to put this file in the `build/es5-bundled/` folder itself and remove the folder prefixes here, then they wouldn't need to have `skip_files` and it wouldn't upload any extra files at all.

bicknellr: Actually, I suppose that would upload `app.yaml` itself, which would probably be bad.

keanulee: `polymer build` cleans out the build/ directory. You could add app.yaml to `extraDependencies` in polymer.json and remember to deploy from build/es5-bundled or run `gcloud deploy build/es5-bundled/app.yaml`

keanulee: You have to upload app.yaml anyways, and it wouldn't be served since there's no handlers that point to it.

katejeffreys: I am ok with leaving app.yaml where it is as long as it works..

app/3.0/start/toolbox/add-elements.md
katejeffreys: (Moved discussion to an email thread)

app/3.0/start/first-element/step-5.md
```diff
@@ -1,124 +1,152 @@
...
+
+Add the following code inside the `<head>` tag of `demo/index.html`:
+
+demo/index.html {.caption}
 
 ```html
 <custom-style>
   <style>
-    /* Define a document-wide default—will not override a :host rule in  */
+    /* Define a document-wide default */
```
arthurevans: These two comments should agree in tense, so either Defines/Overrides or Define / Override.

katejeffreys:
i b0rked this branch while trying to fix it, re-doing.

bicknellr:
`git reflog` is your friend!

PR #2537, opened 2018-04-25
Stackblitz demos for 3.x elements

e111077:
The commit messages might be helpful to follow for review here.

## What I've done:
I added inline demos for `3.0`. It uses [stackblitz](https://stackblitz.com). See gif at the bottom of this PR for example

* [I installed the stackblitz sdk](https://github.com/Polymer/docs/commit/8401dde487258cf624cd216415a36c2eab041e26#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R26)
* ~added the [`generate-stackblitz` npm command which copies it from node modules to a servable folder in the postinstall](https://github.com/Polymer/docs/commit/8401dde487258cf624cd216415a36c2eab041e26#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R11)~
* created a [`stack-blitz` element](https://github.com/Polymer/docs/pull/2537/files#diff-271cb82c8d2ff36f61a164cfc21f938fR1)
  * API is basically `<stack-blitz projectPath="/path/to/project"></stack-blitz>`
  * There may be some vestigial features on the element
* [I updated the `demo-tabs` element to implement stack-blitz](https://github.com/Polymer/docs/pull/2537/files#diff-2ea6bb932a9b49c921a5b0f57caecb71R97)
  * it can detect if it is a [plunkr-style demo or a stackblitz-style demo](https://github.com/Polymer/docs/pull/2537/files#diff-2ea6bb932a9b49c921a5b0f57caecb71R160)
  * It still displays static content to reduce unnecessary data usage
  * [`stack-blitz` is lazy loaded](https://github.com/Polymer/docs/pull/2537/files#diff-2ea6bb932a9b49c921a5b0f57caecb71R198)
  * It functions exactly the same as before for our current `1.0` and `2.0 examples
* I made the [gulpfile copy all the demo files to the dist folder](https://github.com/Polymer/docs/commit/c192eef00068555d0c0a4975279eb49b6eb91530#diff-b9e12334e9eafd8341a6107dd98510c9R300)
* I updated the `3.0` docs to use this new stackblitz demo tabs
* I updated the `3.0` samples to use the [manifest](https://github.com/Polymer/docs/commit/3da5fd30e21438161be147a9451497338c5891c3#diff-ce084cf0fa01b0f1d22b12f0d3a0ea55)

## Usage:
1. Create a sample folder
    1. Add your files
    2. Must include an `index.html` and an `index.js`
    3. Files must be included in a `manifest.json` at the root of that folder
        1. API of manifest json is `{ files: string[], dependencies: { [npmDepName: string]: string }}`
2. [Create a demo tabs element](https://github.com/Polymer/docs/compare/stack-blitz?expand=1#diff-f774835fb93135572b2e7074595423c3R52)
    1. Usage is the same as before
    2. Instead of adding a `src` which was used for plunkr links, use `projectPath` which is the path to the sample folder you just created (e.g. `/3.0/samples/custom-element`)
    3. optional: define an `editorOpenFile` to declare what file will be shown upon editor embed

## Performance
3.0 quick-tour Chrome with service worker and cache on

### first time page load
| test | new | old | change | explanation |
| ---- | --- | --- | ------ | ----------- |
| page load | 662kb | 1mb | **down** |  removed iframe demo |
| opening single editor | 3.7mb | 1mb  | **up** | cannot be service-worked only cached |
| opening all editors | 3.8mb | 1mb  | **up** | editor static files cached |

### subsequent page load
| test | new | old | change | explanation |
| ---- | --- | --- | ------ | ----------- |
| page load | 147kb | 160kb | **down** |  removed iframe demo |
| opening single editor | 156kb | 160kb  | **down** | editor cached |
| opening all editors | 227kb | 160kb  | **up** | editor static files cached |

### Lighthouse
Lighthouse score went up to high 60s from low 60s as we removed the iframe demos and lazy load the editor.

## Outstanding issues
**Critical**
* ~Importing an element that uses `polymer-legacy.js` causes compile error stackblitz/core#441~
* import.meta support stackblitz/core#378 ([to be fixed in end of May](https://github.com/Microsoft/TypeScript/pull/23327))

**Nice to have**
* cannot use non-blocking webcomponents-loader.js stackblitz/core#363
* ~write `node_modules` instead of `turbo_modules` stackblitz/core#333~
* editing a webcomponent's file causes an element registry error if editor preview is not reloaded stackblitz/core#416

## Demo
Try it out! [Link](https://stack-blitz-dot-polymer-project.appspot.com/3.0/start/quick-tour)
clicktoload


app/elements/demo-tabs.html
```diff
@@ -97,22 +120,88 @@
...
+      this._projectPathDefined = !!this.projectPath;
+
+      if (!this._srcDefined && !this._projectPathDefined) {
+        this._hideEditorButton = true;
+      }
     },
 
-    // Call out to a global method defined by app.js
-    _plunkerLaunched: function() {
+    _hideEditorButton: function(srcDefined, prjoectPathDefined) {
```
keanulee: `_hideEditorButton` is a method AND a property?

app/elements/demo-tabs.html
```diff
@@ -97,22 +120,88 @@
...
+        value: 'static',
+      },
+
+      _isPlunkr: {
+        type: Boolean,
+      },
+
+      _isStackBlitz: {
+        type: Boolean,
+      },
```
keanulee: `_isPlunkr` and `_isStackBlitz` unused

app/elements/demo-tabs.html
```diff
@@ -97,22 +120,88 @@
...
+
     attached: function() {
       this._selected = 0;
     },
 
-    _hidePlunkrButton: function(src) {
-      return (!src || src === '');
+    _srcProjectPathChanged: function(src, projectPath) {
+      this._srcDefined = !!this.src;
+      this._projectPathDefined = !!this.projectPath;
```
keanulee: Feel like `_srcDefined` and `_projectPathDefined` aren't needed since they're just proxies for `src/projectPath`

app/elements/stack-blitz.html
```diff
@@ -0,0 +1,186 @@
+<link rel="import" href="./stack-blitz-sdk.html">
```
keanulee: stack-blitz-sdk.html doesn't need to exist (not used elsewhere so doesn't need to be de-duped) - just have `<script src="../js/sdk.umd.js"></script>` here

app/elements/demo-tabs.html
```diff
@@ -97,22 +120,88 @@
...
+      }.bind(this);
+
+      const onLoad = function() {
+        this.$.editor.embed()
+            .then(function(){
+          this._displaySelected = 'editor';
+        }.bind(this)).catch(onError);
+      }.bind(this);
+
+      this.importHref('/elements/stack-blitz.html', onLoad, onError, true);
```
keanulee: 💯 👍

keanulee (changes requested):
Do you have a sample deployment? Maybe @arthurevans or @katejeffreys should try the gulp/deploy process?

app/elements/demo-tabs.html
e111077: oops

app/elements/demo-tabs.html
e111077: ah I forgot to rename `srcDefined` to this etc.

app/elements/demo-tabs.html
e111077: In the process of removing them, but I'm going to need one that is of type boolean as the databinding will not work if I try

`<div hidden$=[[!stringThatIsUndefined]]>`

EricSimons:
Hey everyone- StackBlitz co-creator here! Super excited to see this PR :)

Just wanted to give y'all an update on the ETA for the various items outlined above from our end. With ngconf done and our [Teleport feature](https://www.youtube.com/watch?v=P1-HAN1g4_4) out the door, we're _finally_ able to focus on the next version of our bundler/dev server which solves all the 'critical' & 'nice to have' items. Specifically, we're going to enable serving the static JS files themselves without any transpilation done beforehand.

We're realistically shooting to have this live on prod in ~4 weeks- would love to hear everyone's thoughts/if there's func in polymer CLI that'd be ideal to have on StackBlitz as well 🥂

e111077:
@EricSimons TBH it's miraculous that bare module specifiers work out of the box for ES6 modules with no additional tooling.

There is nothing really polymer-cli-specific that I can think of atm, but there are a few rough patches that we've been filing issues for. Perhaps we can sync up later sometime in VC to discuss this further?

CC @justinfagnani

EricSimons:
Sounds good! My email is eric at esft.com, so feel free to drop me a note and we'll get something on the calendar 👍

katejeffreys:
\o/ i will try to review this Thursday 26th

app/elements/demo-tabs.html
e111077: changes implemented

package.json
```diff
@@ -7,7 +7,8 @@
     "node": ">=5.0.0"
   },
   "scripts": {
-    "postinstall": "bower install; gulp",
+    "postinstall": "bower install; npm run generate-stackblitz; gulp",
+    "generate-stackblitz": "cp ./node_modules/@stackblitz/sdk/bundles/sdk.umd.js ./app/js/sdk.umd.js; cp ./node_modules/@stackblitz/sdk/bundles/sdk.umd.js.map ./app/js/sdk.umd.js.map;",
```
keanulee: Can we use a hosted version of the SDK instead? Not a huge fan of this one-off, manual copying from node_modules/.

```html
<script src="https://unpkg.com/@stackblitz/sdk/bundles/sdk.umd.js"></script>
```

.gitignore
```diff
@@ -3,9 +3,11 @@
...
 *.pyc
 app/css/*
+app/js/sdk*
 dist
 build
 
 # Dev directories
 /app/bower_components
 /node_modules
+/app/node_modules
```
keanulee: Don't think this is created.

keanulee (changes requested):

package.json
e111077: Sure, I'll just need to add this to the service worker

e111077:
@keanulee Here it is staged: https://stack-blitz-dot-polymer-project.appspot.com/3.0/start/quick-tour

package.json
e111077: spoke offline, decided it's best not to add SW

keanulee (changes requested):

gulpfile.js
```diff
@@ -251,6 +251,7 @@ gulp.task('jshint', 'Lint JS', function() {
   return gulp.src([
       'gruntfile.js',
       'app/js/**/*.js',
+      '!app/js/**/sdk*',
```
keanulee: argh, one more thing to remove

e111077: I'm just testing you... you pass

keanulee (approved):
Code changes LGTM. Will defer to @arthurevans and/or @katejeffreys to test on their machines as well.

ghost (approved):
Tested, works beautifully on my machine. Thank you so much for this!!

PR #2560, opened 2018-05-01
add info on stackblitz

katejeffreys:


katejeffreys:
needs lgtm @arthurevans

app/3.0/docs/stack-blitz.md
```diff
@@ -0,0 +1,39 @@
...
+index.js {.caption}
+
+```
+import `my-app.js`;
+```
+
+index.html {.caption}
+```
+<!-- Include WebComponents polyfills -->
+<script src="./node_modules/@webcomponents/webcomponentsjs/webcomponents-lite.js"></script>
```
arthurevans: Going to be webcomponents-bundle.js (or webcomponents-loader.js? I think we'll be able to use the loader here). /cc: @e111077 who's been working with stackblitz, @kevinpschaaf who should know stuff just because.

e111077: I think loader is more-correct?

app/3.0/docs/stack-blitz.md
```diff
@@ -0,0 +1,39 @@
+---
+title: "Polymer 3.0 on StackBlitz"
+---
+
+<!-- toc -->
+
+[StackBlitz](https://stackblitz.com) is an online editor for web projects. You can use StackBlitz to prototype and preview Polymer elements.
+
```
arthurevans: Maybe we should add a quick summary of what StackBlitz does that other online editors don't?

In particular, it transpiles ES6 to ES5 so the samples can run on any browser we support; and that it lets you install dependencies using npm and supports node-style module resolution, so you can preview code without a build step.

You can use Polymer 3 in other online editors (like Plunker or jsbin), but the code may not be transpiled for older browsers, and you may need to use non-standard paths to import dependencies. 

I might not have said that very well. But, um. I think that's why we chose it instead of using other things.

app/3.0/docs/stack-blitz.md
katejeffreys: I thought it was bundle?

PR #2570, opened 2018-05-04
3.0-ify case studies of news and shop apps

katejeffreys:


app/3.0/toolbox/news-case-study.md
```diff
@@ -228,78 +225,14 @@ visits `/list/top_stories`:
   </div>
 </div>
 
-## Resource URLs
```
arthurevans: Why are these sections cut?

katejeffreys: Resource URLs - this seemed server-specific to me. Server-specific config stuff (including app.yaml) was removed in this commit https://github.com/Polymer/news/commit/1a97bb0690eda9bd953eae1bcfdb13f922431b8c

Since the repo is agnostic on how the resource URLs should be handled, I took out the section. Maybe I misunderstood though... wdyt?

The section on AMP I removed because AFAIK there is no AMP branch for 3.0.

app/3.0/toolbox/case-study.md
```diff
@@ -107,38 +104,43 @@ When you change pages, the application loads the definition for the active view.
...
 In the logic above, the home view is built into the app shell, but the other
-views are demand-loaded fragments.
+views are [dynamic imports](https://developers.google.com/web/updates/2017/11/dynamic-import).
 
 Shop also uses [`dom-if`](/{{{polymer_version_dir}}}/docs/api/elements/Polymer.DomIf) templates to lazily create views:
 
 ```html
 <!-- Lazy-create the tabs for larger screen sizes. -->
-<div id="tabContainer" sticky$="[[_shouldShowTabs]]" hidden$="[[!_shouldShowTabs]]">
+<div id="tabContainer" sticky\$="[[_shouldShowTabs]]" hidden\$="[[!_shouldShowTabs]]">
```
notwaldorf: whats up with these `\$`? are they escaping?

katejeffreys: I have no idea what happened there, thanks for picking that up!

katejeffreys: Here it is in the source:

https://github.com/Polymer/shop/blob/c0017828188ad576f0dde5dbe17d1dd277b12b26/src/shop-app.js#L259

katejeffreys: I took out the line.. I think the sample is more focused on the dom-if anyway

PR #2572, opened 2018-05-04
3.0 update - polytool is not a thing, remove page

katejeffreys:


arthurevans:
This was just an old name for the `polymer-cli` package. But not sure we were using this page, anyway.

PR #2574, opened 2018-05-05
3.0-ify toolbox/localize

katejeffreys:


e111077:
app localize is passing if we move the polyfill out of the element as it cannot be modulized. We decided that we will do this manually after modulization. see:

PolymerElements/app-localize-behavior#122

It should be passing when we do the 3.0 element release

app/3.0/toolbox/localize.md
```diff
@@ -4,113 +4,123 @@ title: Localization
...
-          resources: {
-            value() {
-              return {
-                'en': { 'hello': 'My name is {name}.' },
-                'fr': { 'hello': 'Je m\'apelle {name}.' }
-              }
-            }
-          }
-        };
+app-localize.js {.caption}
```
arthurevans: Can we call this something that's less easily confused with app-localize-behavior itself? 
Like localized-app or my-app?

katejeffreys: yep good idea

app/3.0/toolbox/localize.md
```diff
@@ -4,113 +4,123 @@ title: Localization
...
+
+```js
+// Imports go here
+class AppLocalize extends mixinBehaviors([AppLocalizeBehavior], PolymerElement) {
+  static get properties(){
+    return{
```
arthurevans: space after `return`

Or alternately, it would be more idiomatic and more readable to set these values in the constructor and leave off the `propeties` object entirely:

```js
constructor() { 
  super();
  // set default language and localization strings
  this.language = 'en';
  this.resources = {
    'en': { 'hello': 'My name is {name}.' },
    'fr': { 'hello': 'Je m\'apelle {name}.' }
  };
}

File: app/3.0/toolbox/localize.md [hunk 3]
```diff
@@ -4,113 +4,123 @@ title: Localization
...
-        // load localized messages
-        attached: function() {
-          this.loadResources(this.resolveUrl('locales.json'));
-        },
-      });
-   </script>
-</dom-module>
+<head>
+  <script src="./node_modules/@webcomponents/webcomponentsjs/webcomponents-bundle.js"></script>
+  <script type="module" src="app-localize.js"></script>
```
arthurevans: again, if we could change the name to something like localized=app it would be clearer what you'd importing.

PR #2582, opened 2018-05-07
Edits for I/O 2018, 3.0 release

graynorton:


app/index.html
```diff
@@ -327,27 +314,27 @@ <h2 class="underline">Fully interoperable</h2>
           <p>Polymer-powered elements are based on standard Web Components APIs, so they work seamlessly with the browser's built-in elements and other Web Components.</p>
         </div>
       </div>
-    </section>
+    </section> -->
 
     <section class="page-index-examples light">
       <div class="subsection">
-        <h2>Use elements from the catalog</h2>
+        <h2>Use elements</h2>
```
arthurevans: I think we could just cut the examples entirely for this release to focus the page on the main messages:

- 3.0 is out.
- We have new stuff coming.

These examples were here to sell Polymer, which we don't really need to do for 3.0. And they're somewhat confusing with the new hp focus. Is the "create an element" example using Polymer or LitElement? Shouldn't the PWA example be using the PWA kit?

arthurevans: If we cut the examples, the hp is short and sweet and more like a multi-product landing page.

graynorton: Yeah, I thought of cutting them entirely but didn't do it because I was sad about having no code or example stuff on the home page at all. Still, may be the best solution for the moment.

arthurevans:
👍 I may add a couple of the samples into about 3.0, instead. It'll be much lighter than the about 2.0 was, since the changes are not as widespread.

PR #2583, opened 2018-05-07
Update browser compat docs.

arthurevans:
Take 1, updating all this stuff.

Staged here: https://3-0-polyfills-dot-polymer-project.appspot.com/3.0/docs/browsers

Most of the volume here is in my stab at adding info on modules.

Already really good stuff in Dan's README for the polyfills, unsure of whether to pull some of that in here, or to just point to it.

Also, it's still really hard to tell what browser needs what.
I wonder if it'd be better to just have a table like:

Browser & version range   | ES6/ES5 | ES Modules/AMD | Polyfills Req'd | Module name transform

Where we'd have multiple lines for a browser if the values were different.

(And yes, all browsers (currently) require the name=>path transform, but at least it makes it clear that some transform is always required.)

LMK what you think.

### DIFF
--- a/app/3.0/docs/es6.md
+++ b/app/3.0/docs/es6.md
@@ -67,7 +67,7 @@ in the next section.
 
 ### Module specifiers
 
-The browser only accepts on kind of module specifier in an `import` statement: a URL, which must
+The browser accepts only one kind of module specifier in an `import` statement: a URL, which must
 be either fully-qualified, or a path starting with `/`, `./` or `../`. This works fine for importing 
 application-specific elements and modules:
 
@@ -92,7 +92,7 @@ Many third-party build tools, like WebPack and Rollup also support named modules
 
 ### Dynamic imports
 
-Dynamic imports is a language feature that enables lazy-loading of resources, replacing the functionality 
+Dynamic import is a language feature that enables lazy-loading of resources, replacing the functionality 
 that was previously used for this in Polymer 2.x (`Polymer.importHref`). 
 
 The import operator acts like a function, and returns a `Promise`:
@@ -108,7 +108,7 @@ import('my-view1.js').then((MyView1Module) => {
 The latest versions of Chrome and Safari support dynamic imports. For other browsers, the Polymer build tools can transform
 dynamic imports into AMD modules. The Polymer development server can do this transformation on the fly.
 
-More on dyanmic import, see:
+For more on dynamic import, see:
 
 *   [Dynamic import()](https://developers.google.com/web/updates/2017/11/dynamic-import) on developers.google.com.
 
--- a/app/3.0/docs/es6.md
+++ b/app/3.0/docs/es6.md
@@ -1,46 +1,145 @@
 ---
-title: ES6 and compilation to ES5
+title: ES6 and modules
 ---
 
 <!--toc -->
 
-<div>
-{% include 'outdated.html' %}
-</div>
+Polymer 3.x uses JavaScript language features from the ECMAScript 2015 standard 
+(commonly known as ES6). Polymer 3.x code is distributed as a set of ES6 modules.
 
-Polymer 2.x uses EcmaScript 2015 (commonly known as ES6). The following browsers support all of the
-ES6 features required by Polymer:
+Not all browsers support ES6 modules and other ES6 language features, but the Polymer
+tools support transforming code into supported forms:
 
--   Chrome or Chromium version 49 or later.
--   Opera 36 or later.
--   Safari or Mobile Safari 10 or later.
--   Edge 15.15063 or later.
--   Firefox 51 or later.
+*   ES6 modules can be transformed into AMD modules, which are loaded using a small 
+    JavaScript library, instead of depending on native module support. 
+*   Other ES6 language features can be compiled to use a browser-compatible set of
+    features (usually ES5, which is supported by all modern browsers). 
 
-For other browsers, you should compile your application to ES5.
+To deploy your code, you can either transform it for the lowest common denomniator (ES5,
+transformed to AMD modules), or serve multiple versions of the code to different browsers.
 
 Information on multiple builds and differential serving is available in the
 [documentation on building for production](/{{{polymer_version_dir}}}/toolbox/build-for-production).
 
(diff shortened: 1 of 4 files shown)


app/3.0/docs/es6.md
```diff
@@ -1,46 +1,145 @@
...
+    [export](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export) pages on MDN 
+    provide a quick reference on module syntax.
+*   [Modules chapter](http://exploringjs.com/es6/ch_modules.html) from Axel Rauschmayer's Exploring ES6.
+
+As an addition to the standard ES6 module syntax, Polymer supports importing modules by name, as described
+in the next section.
+
+### Module specifiers
+
+The browser only accepts on kind of module specifier in an `import` statement: a URL, which must
```
katejeffreys: on -> one

app/3.0/docs/es6.md
```diff
@@ -1,46 +1,145 @@
...
+Where `@polymer/polymer` is the _name_ of the npm package. (This style of specifier is sometimes called a 
+"bare module specifier".)
+
+These module specifiers need to be transformed to paths before they're served to the browser. The Polymer CLI can transform 
+them at build time, and the Polymer development server can transform them at runtime, so you can test code without a build step.
+Many third-party build tools, like WebPack and Rollup also support named modules.
+
+### Dynamic imports
+
+Dynamic imports is a language feature that enables lazy-loading of resources, replacing the functionality
```
katejeffreys: "Dynamic imports is a language feature"

is -> are? & if so, enables -> enable.

If not, maybe "Dynamic `import()` is a ..."

app/3.0/docs/es6.md
```diff
@@ -1,46 +1,145 @@
...
+  console.log("MyView1 loaded");
+}).catch((reason) => {
+  console.log("MyView1 failed to load", reason);
+});
+```
+
+The latest versions of Chrome and Safari support dynamic imports. For other browsers, the Polymer build tools can transform
+dynamic imports into AMD modules. The Polymer development server can do this transformation on the fly.
+
+More on dyanmic import, see:
```
katejeffreys: More -> For more

"dyanmic" -> "dynamic"

"import" -> "imports" or "`import()`"

app/3.0/docs/es6.md
```diff
@@ -1,46 +1,145 @@
...
-browser.
+browser. 
 
-**Device emulation may cause errors**. When using DevTools device emulation in Chrome,
-`polymer serve` will serve compiled code when emulating iOS devices, and uncompiled code when
-emulating Android devices. When switching between devices, the browser may end up with both
-compiled and uncompiled code in its cache, resulting in errors. To avoid this problem, run
-`polymer serve` with the `--compile never` option when testing with device emulation.
-{.alert .alert info}
+While the Polymer library doesn't depend on any features from newer JavaScript standards,
```
katejeffreys: This sentence sounds weird :)

app/3.0/docs/polyfills.md
```diff
@@ -32,7 +28,7 @@ There are a couple of other related polyfill files that you may need:
     on browsers that support native custom elements. This is useful in static serving environments
     where you need to serve a single app version to all browsers. The adapter is discussed in more
     detail in [ES6](es6) and in [Build for production](/{{{polymer_version_dir}}}/toolbox/build-for-production).
-*   `apply-shim.html`. A polyfill for CSS mixins. Unlike the other polyfills, it should be included
+*   `apply-shim.j`. A polyfill for CSS mixins. Unlike the other polyfills, it should be imported
```
katejeffreys: .j -> .js

app/3.0/docs/polyfills.md
```diff
@@ -130,10 +126,10 @@ Query parameter:
 
 ## Progress of native browser support
 
-As of April 2017, there has been broad cross-browser agreement around the v1 versions of the [Custom
+As of May 2018 the v1 versions of the [Custom
 Elements](https://w3c.github.io/webcomponents/spec/custom/) and [Shadow
-DOM](https://w3c.github.io/webcomponents/spec/shadow/) APIs, with support in Chrome, Opera, and
-Safari Tech Preview, and implementations underway in other browsers.
+DOM](https://w3c.github.io/webcomponents/spec/shadow/) APIs have been shipped in Chrome, Opera, and
+Safari, and implementations underway in other browsers.
```
katejeffreys: are underway

app/3.0/docs/polyfills.md
```diff
@@ -4,18 +4,14 @@ title: Polyfills
...
 
-<div>
-{% include 'outdated.html' %}
-</div>
-
-Polymer 2.x has been developed alongside and tested with a new suite of v1-spec compatible polyfills
-for custom elements and shadow DOM. You can test Polymer 2.x by using the latest 1.0 version of
-`webcomponentsjs`, which is included as a bower dependency to Polymer 2.x. (`webcomponentsjs`
+Polymer 3.x has been developed alongside and tested with a new suite of v1-spec compatible polyfills
+for custom elements and shadow DOM. You can test Polymer 3.x using the latest 2.0 version of
```
katejeffreys: should we mention explicitly that the HTML Imports polyfill has been removed from v2.0 & there is no webcomponents-lite.js?

katejeffreys:
i like the idea of a table showing exactly what polyfills & build transforms each browser requires, provided it wouldn't require too much maintenance for it to stay accurate.

app/3.0/docs/polyfills.md
katejeffreys: & maybe even that you need to do this in polymer.json if using loader?

```
"extraDependencies": [
    "node_modules/@webcomponents/webcomponentsjs/bundles/**"
  ]
```

web-padawan:
BTW the relevant quote from Polymer/tools#191

>
Modules support now requires dynamic import and import.meta.

Specifically, this means:

Safari modules support is now from 11.1 instead of 10.1/10.3
Chrome modules support is now from 64 instead of 61.
Vivaldi supports modules (tested).

PR #2584, opened 2018-05-07
update glossary for 3.0

katejeffreys:


justinfagnani:
Our tools still support all the old workflows and Bower... Is removing the terms the right thing?

app/3.0/docs/glossary.md
```diff
@@ -23,50 +20,28 @@ The [Polymer CLI](#polymer-cli) is designed to work with apps that use an app sh
...
-### bower.json
-`bower.json` is a manifest used by Bower for configuring packages for use as dependencies. See [the 
-`bower.json` specification](https://github.com/bower/spec/blob/master/json.md) for more information.
+Bundling concatenates an app's scripts and resources to reduce the number of network requests it takes to load dependencies. 
 
-## C
-### crisper
-crisper extracts inline scripts from an HTML file. crisper is useful for separating JavaScript from 
-HTML in order to deploy in an environment that uses CSP (Content Security Policy).
+By contrast, if both server and browser support HTTP/2 (a newer protocol that permits concurrent resource requests), it can be optimal to leave resources unbundled. Unbundled builds permit lazy loading of dynamic dependencies. By default, Polymer apps are unbundled.
```
justinfagnani: It's never really optimal. Acceptable?

app/3.0/docs/glossary.md
```diff
@@ -23,50 +20,28 @@ The [Polymer CLI](#polymer-cli) is designed to work with apps that use an app sh
...
 * A data dependency in the context of the Polymer data system: a piece of data that an observer, 
 computed property or computed binding depends on.
 
-### dependency variants
-
-Dependency variants are named sets of Bower [package dependencies](#dependency) at specific 
-versions which are different (usually more specific) than the default dependencies. Variants are 
-specified in [`bower.json`](#bower-json).
+### dynamic import()
+A [dynamic JavaScript `import()` statement](#https://developers.google.com/web/updates/2017/11/dynamic-import) loads the bindings exported by another module. In contrast to static [import](#import) statements, dynamic imports load their targets at runtime.
```
justinfagnani: "Loads the binding" is a little specific. Loads the module is fine. Both static and dynamic import load at runtime, but dynamic imports load based on code execution, and can take a computed string.

app/3.0/docs/glossary.md
```diff
@@ -78,24 +53,34 @@ The HTML actually loaded by the browser to start the application. Usually `index
...
 
+### ES (ES5, ES6/ES2015)
+
+ECMAScript. A specification created to standardize JavaScript. 
+
+ES5 and ES6 (ES2015) are editions of ECMAScript; ES6 adds features to ES5, such as [imports](#import).
+
+The Web Components standards and the Polymer library are based on ES6 JavaScript features. Polymer CLI can compile JavaScript code that uses newer features to ES5 so that older browsers can still run your app.
+
+### ES modules
```
justinfagnani: AKA "Standard JavaScript Modules"

app/3.0/docs/glossary.md
arthurevans: SG

app/3.0/docs/glossary.md
arthurevans: Elsewhere in the docs we've used ES6 modules, rather than just ES modules, I think.

arthurevans:
I agree with Justin that we should leave in old terms (like bower) and just change "Polymer uses" to  "Polymer 1.x & 2.x use ..." ... Maybe with a pointer to the replacement.

arthurevans:
@katejeffreys Ping on this one?

app/3.0/docs/glossary.md
arthurevans: Nm. I think ES modules is good. Will work on standardizing that.

app/3.0/docs/glossary.md
```diff
@@ -4,14 +4,11 @@ title: "Polymer Glossary"
...
-{% include 'outdated.html' %}
-</div>
-
 ## A
 
-### Analyzer
-See [Polymer Analyzer](#polymer-analyzer).
+### AMD modules
+
+AMD (Asynchronous Module Definition) modules use the `define()` function to write modular JavaScript code that can be loaded asynchronously. Polymer CLI can convert [ES modules](#es-modules) to AMD modules so that older browsers can load them.
```
arthurevans: I think the most relevant information here is that AMD modules are loaded using a (small) library, so they can be used with browsers that don't support native ES modules.

app/3.0/docs/glossary.md
arthurevans: Maybe add (to Justin's point) that ES modules are a JavaScript language feature added in ES6, so they're available natively in newer browsers unlike AMD modules.

app/3.0/docs/glossary.md
```diff
@@ -78,24 +53,34 @@ The HTML actually loaded by the browser to start the application. Usually `index
...
 [`polymer-build` library](#polymer-build).
 
-## H
-### hydrolysis
-This tool has been replaced by [Polymer Analyzer](#polymer-analyzer). It is a static analysis 
-framework for Web Components. 
+## I
+### import
+
+The [JavaScript `import` statement](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) loads bindings exported by another module. Polymer 3.0 uses `import` statements to load the Polymer library, utilities and elements. `import` is an [ES6](#es6) feature.
```
arthurevans: Specifically: loads an ES module.

app/3.0/docs/glossary.md
katejeffreys: I made the change to "acceptable". 

@justinfagnani @arthurevans For my own understanding, can you explain more about why it's never optimal to have unbundled resources? I thought bundling was suboptimal with lazy loading, and therefore unbundled builds would be optimal.

katejeffreys:
@arthurevans @justinfagnani I will merge today unless either of you wants to take another look :)

PR #2599, opened 2018-05-09
Hotfixes from 2.x

keanulee:


keanulee:
All committers have CLAs

PR #2607, opened 2018-05-12
type miss on document.

takahiro-saeki:


takahiro-saeki:
@googlebot I signed it!

PR #2622, opened 2018-05-24
add an elements 3.0 faq post

notwaldorf:


app/blog/2018-05-25-polymer-elements-3-faq.md
```diff
@@ -0,0 +1,32 @@
...
+Uncaught TypeError: Failed to resolve module specifier. Relative references must start with either "/", "./", or "../".
+```
+
+This error occurs in Chrome when loading module specifiers by name rather than by URL.
+The easiest way to address this during development is to use the `polymer-cli` and run
+`polymer serve` to serve your app. For deployment and production, you can run
+`polymer build`, or a Webpack/Rollup solution that converts the named imports to paths.
+
+## The `package.json` does not have a `main` field
+This is due to multiple elements existing in the same repo, and multiple main files being listed in our `bower.json`. This is required
```
arthurevans: 2nd sentence is a little confusing. Suggest maybe:

This happens when multiple elements are defined in the same repo. Our modulizer tool, which converts the 2.0 elements to 3.0, requires each of these elements to be listed as main files in our `bower.json`. It uses this information to crawl the dependency tree and convert multiple elements. `package.json` won't accept multiple files in the `main` field. So some of our packages ...

notwaldorf:
@arthurevans ping?

PR #2623, opened 2018-05-25
update testing doc 3.0 with a11y paragraph

lemoustachiste:
Adding a paragraph about how to set up accessibility testing in a test suite. I did lose way too much of my time trying to get `a11ySuite` to work only to discover that this is not the right way any more.

### DIFF
--- a/app/3.0/docs/tools/tests.md
+++ b/app/3.0/docs/tools/tests.md
@@ -295,6 +295,57 @@ Check out Sinon's documentation for more in-depth examples.
 assertion style.
 { .alert .alert-info }
 
+### Testing for accessibility
+To test for accessibility, you will need to add the following dependencies:
+
+`npm i -D @webcomponents/webcomponentsjs axe-core web-component-tester pwa-helpers`
+
+or with yarn:
+
+`yarn add -D @webcomponents/webcomponentsjs axe-core web-component-tester pwa-helpers`
+
+Then you need to setup your test file
+
+```html
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>A11y test</title>
+
+    <script src="../node_modules/@webcomponents/webcomponentsjs/webcomponents-bundle.js"></script>
+    <script src="../node_modules/web-component-tester/browser.js"></script>
+
+    <script type="module" src="path/to/your-module.js"></script>
+</head>
+<body>
+    <test-fixture id="a11y">
+        <template>
+            <your-module></your-module>
+        </template>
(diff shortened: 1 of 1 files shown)


katejeffreys:
Thanks! I will take a look as soon as possible

katejeffreys:
Thanks again @lemoustachiste and sorry for the delay.

There's nothing in the [web-component-tester changelog](https://github.com/Polymer/tools/blob/master/packages/web-component-tester/CHANGELOG.md) about a11y-suite no longer working, so I'm curious as to whether something needs to be fixed rather than changing the docs to outline a new process.

What makes you say that a11y-suite is no longer the right way to do it - were you given this info by someone on the Polymer team or did you see a commit somewhere that removed support for it?

Can you point me in the right direction by linking a team member/commit/etc so I can check in with the team? :) Thank you in advance

Ping @aomarks

aomarks:
I'm not really familiar with axe/pwa-helpers, @notwaldorf does this look reasonable?

lemoustachiste:
@katejeffreys, no worries. 

I am a newcomer to Polymer and I initially tried to implement everything following the docs, only to hit walls (and I tried many ways to go about them). 

I had `a11ySuite` implemented as documented, but that was definitely not working (and I did spend a solid amount of time there). I posted on stackoverflow, as well as on the slack channel, and this is what I got: https://stackoverflow.com/questions/50517884/polymer-3-0-a11ysuite-testing-error.

Resorting to the approach described in the PR worked, and seeing that `pwa-starter-kit` (which seems to be a more solid reference than the docs) used that too, I thought this was the new way.

katejeffreys:
@lemoustachiste thanks for the clarification!

@usergenic, @rictic, the [most recent entry about a11y in the web-components-tester changelog](https://github.com/Polymer/tools/blob/master/packages/web-component-tester/CHANGELOG.md#643---2018-01-11) mentions that users should import a11ySuite from wct-browser-legacy. The [web-component-tester README](https://github.com/Polymer/tools/tree/master/packages/web-component-tester) still says it uses a11ySuite. 

From looking at the [pre-monorepo history of the web-component-tester changelog](https://github.com/Polymer/web-component-tester/commit/bb7e8a0d32f513ac554b8c315825a240c11a031d), it looks like Fred had something to do with this so I'm not sure who to ask. 

Can either of you confirm 

1. Should importing a11ySuite from wct-browser-legacy still work for users of web-component-tester?
2. Should the testing docs recommend and describe how to use [axe-core](https://github.com/dequelabs/axe-core) for accessibility testing?

PR #2631, opened 2018-06-07
Create CODEOWNERS

samuelli:


samuelli:
We can also be more specific, like code owners for particular directories.

arthurevans:
List looks like a good start. @notwaldorf are you OK with being on this list? Maybe @keanulee  should be, too?

If I'm getting the format correct, I think this should send content reviews straight to Kate & me, and send other reviews to Monica and Keanu (with me as backup). 

```
* @notwaldorf @keanulee @arthurevans
**/*.md @katejeffreys @arthurevans
**/*.yaml @katejeffreys @arthurevans
```
@katejeffreys LMK if you want to be on the global list as well. This should only get PRs that include files other than MD and YAML changes.

keanulee:
I only want to be assigned non-content changes. If that's not possible, then maybe I only get assigned changes to elements/ and gulpfile.js (or just leave me off CODEOWNERS and manually assign me as necessary).

samuelli:
The code owners implementation isn't great. The main thing is a) point of contact for us to look at, b) external community PRs get auto-assigned code owners. So we should keep the list pretty small. We don't have forced code owners checks, so you can always add someone else as needed.

samuelli:
I might change it so its just @arthurevans @katejeffreys for all changes?

katejeffreys:
I'm basically fine with anything; depends more on @notwaldorf and @keanulee opinions on being on the list (i.e. if it's going to be too much or not relevant traffic).

I'm happy to be assigned all changes and follow up with people internally, or to be assigned to content changes only, whatever works.

arthurevans:
Looks good. Thanks @samuelli!

PR #2632, opened 2018-06-12
Fix incorrect legacy behavior code sample in 3.0 docs

justinribeiro:
The legacy behavior code sample in the current 3.0 docs is incorrect.

1. The imports are not correct (they're actually backwards).
2. `Polymer.mixinBehaviors` is used when it should be only `mixinBehaviors`
3. It is unclear where `MyBehavior` and `MyBehavior2` come from.

This PR fixes the above issues by updating the code sample, with the addition of `MyBehavior` and `MyBehavior2` as imports to help devs who are new to `import` how a legacy behavior might be used.


katejeffreys:
Thank you!

PR #2636, opened 2018-06-14
Add info about array .length observers (Edit #2420)

katejeffreys:
Edits for #2420 by Flamenco


katejeffreys:
Will merge in a couple days unless there are issues :)

PR #2646, opened 2018-07-20
Add note about duplicate dependency installation

keanulee:
Ref https://github.com/PolymerElements/paper-button/issues/174


jajaperson:
Why can't I see these changes yet?

PR #2664, opened 2018-10-06
Added the missed steps

vidhya03:
If we continued as per the documentation , we end up with variable override in :host pesudo classes


arthurevans:
Thanks for the PR. In this case, the instructions are correct. If you follow the instructions, you should see the pressed icon toggle is blue. because that rule is overriding the one in `:host`.

The docs describe this:

"The demo-element selector matches the <demo-element> element and has a higher specificity than the :host selector inside demo-element. You can use this fact to specify your own values for custom properties that the author of a custom element has defined inside that custom element, without altering the element's code."

Hope that helps. Please feel free to follow up if it's still not working for you.

Issue #2508, opened 2018-03-31
Add note about the strict-binding-parser in data-binding section


TimvdLippe:
In the section https://www.polymer-project.org/2.0/docs/devguide/data-binding#anatomy-of-a-data-binding, the docs should be expanded with a note stating there is a strict-binding-parser (introduced in https://github.com/Polymer/polymer/pull/4768) that can handle more cases, but is less performant.

For an example note, see https://github.com/Polymer/polymer/pull/4768/commits/89ab7385d24dc44cd920979023e00c99e4268657

rodrigo-garcia-leon:
Hi, I'd like to work on this one

TimvdLippe:
Feel free to send a PR and tag me as reviewer!

PR #2665, opened 2018-10-15
WIP: Add strict binding parser note

rodrigo-garcia-leon:
Hi @TimvdLippe,

This is what I got so far from issue #2508.
I'm missing the info for Polymer 3, I wanted to check with you before adding that.
Also, I'm not sure if Polymer 1 should be in scope for this.

Anyways, looking forward to your comments

### DIFF
--- a/app/2.0/docs/devguide/data-binding.md
+++ b/app/2.0/docs/devguide/data-binding.md
@@ -697,6 +697,44 @@ to listen for `property-changed` events.  The following constructions are equiva
 <my-element value="{{hostValue}}">
 ```
 
+## Strict binding parser
+
+The default implementation of the binding parser uses a regular expression, due to its better
+performance. However, since it uses a white-list of allowed characters, some characters won't be
+accepted as valid within a binding expression.
+
+There's a strict binding parser implementation, extracted in a separate mixin, which uses a
+state-machine instead of a regular expression. This implementation is able to handle all possible
+cases, although it's slightly less performant.
+
+Example: { .caption }
+
+```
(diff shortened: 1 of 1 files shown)


TimvdLippe:
Thanks! This all looks fine to me, but I will leave the rest of the review to our doc writers.

app/2.0/docs/devguide/data-binding.md
```diff
@@ -697,6 +697,44 @@ to listen for `property-changed` events.  The following constructions are equiva
 <my-element value="{{hostValue}}">
 ```
 
+## Strict binding parser
+
+The default implementation of the binding parser uses a regular expression, due to its better
```
arthurevans: Thanks for this! I'd like to suggest alternate text to focus more on the use cases (when you need it, what to do) and less on the implementation. Here's my first take:

===
The default binding parser emphasizes performance over completeness. Some bindings can't be parsed correctly by the default parser—for example, bindings with JavaScript identifiers that include non-ASCII characters.

For a more exact binding parser, use the `StrictBindingParser` mixin. This mixin provides an alternate binding parser that handles all possible cases, although it's slightly less performant than the default parser.
===

I don't love the phrase, "handles all possible cases," but I can't think of anything better at the moment, either.

rodrigo-garcia-leon: Thanks for the feedback. Regarding the last sentence, what about something like this:
This mixin provides an alternate binding parser that supports Unicode characters, although it's slightly less performant than the default parser.

